PR #4, opened 2019-02-20
guard against no repo tags

TomShacham:


cristianrgreco:
I've implemented this slightly differently: https://github.com/testcontainers/testcontainers-node/commit/f186d763bddba0af6e380267d719e5471bd3e3f7

It's fixed in version 1.1.10 👍

PR #6, opened 2019-05-07
fix: delegate startup timeout setting to wait strategy

Meemaw:
I have problem with an image inside a Gitlab CI that requires more than 10 seconds (default `startupTimeout`) to start.

Even though `GenericContainer` exposes a method `withStartupTimeout`, it is not delegated to the `WaitStrategy`(HostPortWaitStrategy) and thus ignored.

  This MR adds a simple fix for that.


cristianrgreco:
Great, thank you @Meemaw!

PR #8, opened 2019-05-15
Add a command parameter to DockerClient.create()

bonomat:
Hi, 

In this little PR I've added the ability to provide a (optional) command when creating a new container. 
The reason is that sometimes you want to start the service inside a container with specific parameters. 

Changes in a nutshell: 
* `DockerCLient.create(repoTag: RepoTag, env: Env, boundPorts: BoundPorts, cmd: Command[] = []): Promise<Container>;` --> added new `cmd` parameter

Extending this function with the new parameter is not a breaking change as `cmd` is optional and defaults to an empty array. 

However, the default parameter might become a problem if more parameters are added: 
the underlying library (Dockerode) allows to pass in various options in `createContainer()`, i.e.:
https://github.com/DefinitelyTyped/DefinitelyTyped/blob/1957e4cb6a5ed54ca310f015e5a61411bb713da7/types/dockerode/index.d.ts#L836

In the future, if you would like to allow more (optional) parameters to be added, I would propose going for a different approach as described in here: https://www.csgpro.com/blog/2016/08/typescript-default-parameters-and-destructuring-as-a-substitute-for-named-parameters

Let me know if you have any suggestions for improvements.
Cheers

### DIFF
--- a/src/docker-client.ts
+++ b/src/docker-client.ts
@@ -21,7 +21,7 @@ type DockerodeExposedPorts = { [port in PortString]: {} };
 
 export interface DockerClient {
   pull(repoTag: RepoTag): Promise<void>;
-  create(repoTag: RepoTag, env: Env, boundPorts: BoundPorts): Promise<Container>;
+  create(repoTag: RepoTag, env: Env, boundPorts: BoundPorts, cmd?: Command[]): Promise<Container>;
   start(container: Container): Promise<void>;
   exec(container: Container, command: Command[]): Promise<ExecResult>;
   fetchRepoTags(): Promise<RepoTag[]>;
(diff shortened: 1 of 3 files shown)


src/docker-client.ts
```diff
@@ -21,7 +21,7 @@ type DockerodeExposedPorts = { [port in PortString]: {} };
 
 export interface DockerClient {
   pull(repoTag: RepoTag): Promise<void>;
-  create(repoTag: RepoTag, env: Env, boundPorts: BoundPorts): Promise<Container>;
+  create(repoTag: RepoTag, env: Env, boundPorts: BoundPorts, cmd?: Command[]): Promise<Container>;
```
cristianrgreco: As this will always be an empty array, could we make it a required field?

src/docker-client.ts
```diff
@@ -37,13 +37,14 @@ export class DockerodeClient implements DockerClient {
     await streamToArray(stream);
   }
 
-  public async create(repoTag: RepoTag, env: Env, boundPorts: BoundPorts): Promise<Container> {
+  public async create(repoTag: RepoTag, env: Env, boundPorts: BoundPorts, cmd: Command[] = []): Promise<Container> {
```
cristianrgreco: Same here, could we remove the default initialisation?

cristianrgreco:
Hi @bonomat 

Thanks for raising this PR!
I've just left a couple of nitpickity comments.

> In the future, if you would like to allow more (optional) parameters to be added, I would propose going for a different approach as described in here: https://www.csgpro.com/blog/2016/08/typescript-default-parameters-and-destructuring-as-a-substitute-for-named-parameters

Totally agree, now that the level of customisation is growing it'll becoming something we need to do - perhaps when we add the next field we can change it 🙂 

Cheers

src/docker-client.ts
bonomat: Sure thing, I can remove it. 
The reason why I proposed this was to introduce no breaking changes to the API.

Issue #9, opened 2019-05-15
Allow to wait for log messages


thomaseizinger:
Hello from the Rust universe! :)

Thanks for creating a `node` port of testcontainers!

**Context:** Some containers take quite a while to start up and open ports before they are actually ready (bitcoin for example).

**Possible solution:** One way that works well for us in [testcontainers-rs](https://github.com/testcontainers/testcontainers-rs) is to wait for a particular log message to be emitted:

https://github.com/testcontainers/testcontainers-rs/blob/f2f94f247fbdb87605c8e7dc18a67fa409250578/images/coblox_bitcoincore/src/image.rs#L167-L172

It would be nice if `testcontainers-node` would support that aswell :)
Happy to send PR if you outline how you would like the library to be extended.

cristianrgreco:
Hi @thomaseizinger, could you take a look at the PR and let me know your thoughts?

cristianrgreco:
Hi @thomaseizinger, this has been released in version `1.1.18` which is available now 🙂

PR #10, opened 2019-05-15
Implement LogWaitStrategy

cristianrgreco:
Addresses #9. Example usage:

```javascript
const container = await new GenericContainer("redis")
    .withExposedPorts(6379)
    .withWaitStrategy(Wait.forLogMessage("Ready to accept connections"))
    .start();
```

### DIFF
--- a/src/wait-strategy.ts
+++ b/src/wait-strategy.ts
@@ -85,13 +85,13 @@ export class LogWaitStrategy extends AbstractWaitStrategy {
     return new Promise(async (resolve, reject) => {
       const stream = await container.logs();
       stream
-        .on("data", chunk => {
-          if (chunk.toString().includes(this.message)) {
+        .on("data", line => {
+          if (line.toString().includes(this.message)) {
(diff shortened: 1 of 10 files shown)


src/wait-strategy.ts
```diff
@@ -64,3 +73,31 @@ export class HostPortWaitStrategy extends AbstractWaitStrategy {
...
+export class LogWaitStrategy extends AbstractWaitStrategy {
+  constructor(private readonly message: Log) {
+    super();
+  }
+
+  public async waitUntilReady(container: Container): Promise<void> {
+    return new Promise(async (resolve, reject) => {
+      const stream = await container.logs();
+      stream
+        .on("data", chunk => {
```
thomaseizinger: Is this chunk a single line?
If this is not line-buffered, it might be possible to miss certain messages and wait forever.

cristianrgreco: It is now line buffered 👍

PR #14, opened 2019-05-31
Feature/container creation api broader

vtajzich:
Enhanced API

- get ID from generic container
- set container name
- set network for attaching container
- set hostname

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
   "name": "testcontainers",
   "author": "Cristian Greco",
-  "version": "1.1.19",
+  "version": "1.2.0",
   "main": "dist/index",
   "types": "dist/index",
   "keywords": [
@@ -47,6 +47,7 @@
     "@types/node-fetch": "^2.1.4",
     "@types/stream-to-array": "^2.3.0",
     "@types/tar-fs": "^1.16.1",
+    "uuid": "^3.3.2",
     "husky": "^1.3.1",
     "jest": "^23.6.0",
     "node-fetch": "^2.3.0",
--- a/src/docker-client.ts
+++ b/src/docker-client.ts
@@ -24,23 +24,16 @@ type DockerodeExposedPorts = { [port in PortString]: {} };
 
 export interface DockerClient {
   pull(repoTag: RepoTag): Promise<void>;
-
   create(repoTag: RepoTag, env: Env, boundPorts: BoundPorts, cmd: Command[], name?: string, network?: string, hostname?: string): Promise<Container>;
-
   start(container: Container): Promise<void>;
-
   exec(container: Container, command: Command[]): Promise<ExecResult>;
-
   buildImage(repoTag: RepoTag, context: BuildContext): Promise<void>;
-
   fetchRepoTags(): Promise<RepoTag[]>;
-
   getHost(): Host;
 }
 
 export class DockerodeClient implements DockerClient {
-  constructor(private readonly host: Host, private readonly dockerode: Dockerode) {
-  }
+  constructor(private readonly host: Host, private readonly dockerode: Dockerode) {}
 
   public async pull(repoTag: RepoTag): Promise<void> {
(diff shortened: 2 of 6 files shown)


package.json
```diff
@@ -47,6 +47,7 @@
     "@types/node-fetch": "^2.1.4",
     "@types/stream-to-array": "^2.3.0",
     "@types/tar-fs": "^1.16.1",
+    "uuid": "^3.3.2",
```
cristianrgreco: There's already a `RandomUuid` class in the project which we can use.

src/generic-container.test.ts
```diff
@@ -19,6 +22,39 @@ describe("GenericContainer", () => {
     await expect(fetch(url)).rejects.toThrowError();
   });
 
+  it("should be assigned to network and reachable by name", async () => {
+
+    const dockerode = new Dockerode();
```
cristianrgreco: We've extracted a `DockerClient` interface to avoid dealing with `Dockerode` directly, as it's sometimes too low level. It also uses callbacks whereas `testcontainers` aims to use Promises and async/await.

Please implement the required functionality in the `DockerodeDockerClient` class.

src/generic-container.test.ts
```diff
@@ -1,10 +1,13 @@
...
 import fetch from "node-fetch";
 import path from "path";
+// @ts-ignore
+import uuid from 'uuid';
 import { GenericContainer } from "./generic-container";
 import { Wait } from "./wait";
 
 describe("GenericContainer", () => {
-  jest.setTimeout(45000);
+  jest.setTimeout(120000);
```
cristianrgreco: Why has this been increased so much?

src/generic-container.test.ts
```diff
@@ -1,10 +1,13 @@
+import Dockerode = require("dockerode");
 import fetch from "node-fetch";
 import path from "path";
+// @ts-ignore
```
cristianrgreco: We must not ignore the compiler.

src/test-container.ts
```diff
@@ -10,9 +10,13 @@ export interface TestContainer {
   withExposedPorts(...ports: Port[]): TestContainer;
   withWaitStrategy(waitStrategy: WaitStrategy): TestContainer;
   withStartupTimeout(startupTimeout: Duration): TestContainer;
+  withName(name: string): TestContainer;
+  withNetwork(network: string): TestContainer;
+  withHostname(hostname: string): TestContainer;
 }
 
 export interface StartedTestContainer {
+  getId(): string;
```
cristianrgreco: What is the use case for needing the ID of the container?

src/generic-container.ts
```diff
@@ -36,6 +36,9 @@ export class GenericContainer implements TestContainer {
   private cmd: Command[] = [];
   private waitStrategy?: WaitStrategy;
   private startupTimeout: Duration = new Duration(60_000, TemporalUnit.MILLISECONDS);
+  private name?: string;
```
cristianrgreco: We're introducing types to make them easier to validate/refactor in the future. 
Eg `name: ContainerName`
Could we do the same here?

src/docker-client.ts
```diff
@@ -24,7 +24,7 @@ type DockerodeExposedPorts = { [port in PortString]: {} };
 
 export interface DockerClient {
   pull(repoTag: RepoTag): Promise<void>;
-  create(repoTag: RepoTag, env: Env, boundPorts: BoundPorts, cmd: Command[]): Promise<Container>;
+  create(repoTag: RepoTag, env: Env, boundPorts: BoundPorts, cmd: Command[], name?: string, network?: string, hostname?: string): Promise<Container>;
```
cristianrgreco: I think it's time to create an object for all these parameters 🙂

cristianrgreco:
Thanks @vtajzich for the PR! 

I've left a few comments which should help bring the code inline with existing interfaces.

Issue #15, opened 2019-07-10
getContainerIpAddress() does not work as expected in Gitlab CI docker runner


KaiStapel:
I have a similar issue like #7 

I am starting two docker containers in my tests (postgres and rabbitMQ). Then I construct the connection strings using `container.getContainerIpAddress()` and `container.getMappedPort()`. Everything works locally. However, in Gitlab CI I get the following error.
```
     Error: connect ECONNREFUSED 127.0.0.1:44937
      at TCPConnectWrap.afterConnect [as oncomplete] (net.js:1056:14)
```

When checking the value of `getContainerIpAddress()` it always returns `localhost` even when run in the Gitlab CI docker runner. While the Java counter part of testcontainers returns a real IP address like `172.17.0.1` in Gitlab CI.

Can you adjust your code to behave like your Java counter part?

cristianrgreco:
Hi @KaiStapel 

Have you followed the docs here: https://www.testcontainers.org/supported_docker_environment/continuous_integration/gitlab_ci/ ?

Testcontainers will look for a `DOCKER_HOST` env variable, so as long as you are setting that it should work.

KaiStapel:
I am not using dind service. I am directly using the docker daemon of the docker runner itself which is on default socket `unix:///var/run/docker.sock`. This works fine with Java, but fails with NodeJS as the IP is not resolved correctly (or at all).

KaiStapel:
Looks like the Java code is doing some strange stuff if run inside a container: https://github.com/testcontainers/testcontainers-java/blob/master/core/src/main/java/org/testcontainers/dockerclient/DockerClientConfigUtils.java#L48

cristianrgreco:
@KaiStapel Could you please post the value of the `DOCKER_HOST` env variable in Gitlab CI?

KaiStapel:
In my case it is undefined. That means the default socket `unix:///var/run/docker.sock` is being used.

cristianrgreco:
@KaiStapel if you're using the socket, wouldn't `localhost` be the correct hostname?

KaiStapel:
No. I guess my setup is special because I am re-using the docker deamon of the gitlab CI docker runner itself. That means the docker runner first creates a container for the build. Lets call it `build`. Then, inside of `build`, the outer deamon is used to create a new container like `postgres`. Both containers are "local" to the runner. But from within the `build` container, in order to access the `postgres` container, one needs the IP address of the runner.

[...]

PR #16, opened 2019-07-11
Set hostname as default gateway when running in container

cristianrgreco:
#15

### DIFF
--- a/src/docker-client-factory.ts
+++ b/src/docker-client-factory.ts
@@ -1,22 +1,29 @@
 import Dockerode from "dockerode";
+import { Writable } from "stream";
 import url from "url";
 import { DockerClient, DockerodeClient } from "./docker-client";
+import { FileHelper, RealFileHelper } from "./file-helper";
 import log from "./logger";
+import ProcessEnv = NodeJS.ProcessEnv;
 
 export type Host = string;
 
 export interface DockerClientFactory {
-  getClient(): DockerClient;
-  getHost(): Host;
+  getClient(): Promise<DockerClient>;
+  getHost(): Promise<Host>;
 }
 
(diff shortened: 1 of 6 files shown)


src/docker-client-factory.ts
```diff
@@ -1,22 +1,29 @@
...
-  private readonly host: Host;
-  private readonly client: DockerClient;
+  private readonly host: Promise<Host>;
+  private readonly client: Promise<DockerClient>;
 
-  constructor() {
-    if (process.env.DOCKER_HOST) {
-      const { host, client } = this.fromDockerHost(process.env.DOCKER_HOST);
+  constructor(env: ProcessEnv = process.env, private readonly fileHelper: FileHelper = new RealFileHelper()) {
+    if (env.DOCKER_HOST) {
```
KaiStapel: What if someone sets `DOCKER_HOST=unix:///var/run/docker.sock`?

dweidenfeld:
What about requesting the default gateways address from the current host, since you're already running inside Docker, this seems to work on Gitlab CI / CircleCI, Localhost (Mac, Linux).

Just see https://github.com/testcontainers/testcontainers-node/pull/22
If this is not interesting for you, I can also delete this part from the PR.

src/docker-client-factory.ts
dweidenfeld: Is there a use-case for it?

KaiStapel: If someone configures their GitlabCI docker runner with `DOCKER_HOST=unix:///var/run/docker.sock` then it should still work, shouldn't it (despite this configuration being useless because it is the default)?

Imo, the distinction should not be done based on the presence of a environment variable, but rather based on its content (which includes the default if it is not present).

PR #18, opened 2019-07-22
Add exec support to started containers

mikesir87:
When testing base images in my CI system, there are times in which I'd like to exec into a container to validate command output. This PR adds that functionality.


cristianrgreco:
Awesome, thanks @mikesir87 !

cristianrgreco:
This is available now in version `1.2.0`

PR #22, opened 2019-08-06
feat(buildargs): specify build args on build time

dweidenfeld:
This feature includes the possiblity to add build args to the Docker
build process.
It also gets the IP from the default gateway, if you want to let it run
as DIND (Docker in Docker) or also known as wormhole pattern.


docker-with-buildargs/index.js
```diff
@@ -0,0 +1,18 @@
...
+  res.status(200).send("hello-world");
+});
+
+app.get("/env", (req, res) => {
+  res.status(200).json(process.env);
+});
+
+app.get("/cmd", (req, res) => {
+  res.status(200).json(process.argv);
+});
```
cristianrgreco: At the moment the test is only verifying that the image builds successfully. It would be nice to create an endpoint which returns build-args, that way we can curl this endpoint and verify the values in the test.

src/docker-client-factory.ts
```diff
@@ -1,4 +1,7 @@
+// @ts-ignore
+import defaultGateway from "default-gateway";
```
cristianrgreco: Let's add the typings for this module.

src/docker-client-factory.ts
```diff
@@ -1,4 +1,7 @@
...
 import Dockerode from "dockerode";
+import * as fs from "fs";
```
cristianrgreco: Equivalent to `import fs from "fs";`

src/docker-client.ts
```diff
@@ -27,7 +27,7 @@ export interface DockerClient {
   create(repoTag: RepoTag, env: Env, boundPorts: BoundPorts, cmd: Command[]): Promise<Container>;
   start(container: Container): Promise<void>;
   exec(container: Container, command: Command[]): Promise<ExecResult>;
-  buildImage(repoTag: RepoTag, context: BuildContext): Promise<void>;
+  buildImage(repoTag: RepoTag, context: BuildContext, buildParameters: {}): Promise<void>;
```
cristianrgreco: `buildParameters: {}` is telling TypeScript that the type of `buildParameters` is an empty object, not an object. You probably wanted `buildParameters: object`, but it'd actually be better to define a custom type for this. Something like: `type BuildParameters = {[key: string]: string}`.

src/generic-container.test.ts
```diff
@@ -74,6 +74,30 @@ describe("GenericContainer", () => {
...
+
+    await startedContainer.stop();
+    await expect(fetch(url)).rejects.toThrowError();
+  });
+
+  it("should not build and start from a Dockerfile with missing build arguments", async () => {
+    const context = path.resolve(__dirname, "..", "docker-with-buildargs");
+    const container = await GenericContainer.fromDockerfile(context);
+
+    expect(container.start()).rejects.toThrowError();
```
cristianrgreco: Can we add what the error will be? It should be meaningful. For example: 
`expect(container.start()).rejects.toThrowError("Failed to build Docker image, missing build-args: arg1, arg2");`.

src/generic-container.test.ts
```diff
@@ -74,6 +74,30 @@ describe("GenericContainer", () => {
     await expect(fetch(url)).rejects.toThrowError();
   });
 
+  it("should build and start from a Dockerfile with build arguments", async () => {
+    const context = path.resolve(__dirname, "..", "docker-with-buildargs");
+    const container = await GenericContainer.fromDockerfile(context, undefined, undefined, {
```
cristianrgreco: Not a fan of these `undefined`. If these arguments have to be there it'd be better to have `fromDockerfile` accept an object of parameters, so that we can skip the optional ones.

cristianrgreco:
Thank you for the PR @dweidenfeld! I've left a few suggestions. The build seems to be failing around creating the Docker container.

docker-with-buildargs/index.js
dweidenfeld: I don't think this is an issue, since I am currently setting the image tag with the build arg. So the image will never build if the build arg is not passed properly ;)
We should not test more functionality, than we need to test here.

src/docker-client-factory.ts
dweidenfeld: done

src/docker-client-factory.ts
dweidenfeld: done

src/docker-client.ts
dweidenfeld: I'll do a rework for the build arguments in a few minutes ;)

src/generic-container.test.ts
dweidenfeld: done

src/generic-container.test.ts
dweidenfeld: Like I said, I'll do a rework for the arguments ;)

dweidenfeld:
@cristianrgreco I've completely reworked the build parameter passing. This should be a nicer syntax now, especially for optional parameters.

The build should also work now.

cristianrgreco:
Hey @dweidenfeld, looks like the build is broken. Is this PR still active?

dweidenfeld:
Sorry man, I haven't had the time to look at it the last days. I hope to fix it this week ;)

dweidenfeld:
Okay, this was an easy one. It is fixed now.. Thanks for the reminder ;)

src/generic-container.test.ts
```diff
@@ -74,6 +75,25 @@ describe("GenericContainer", () => {
     await expect(fetch(url)).rejects.toThrowError();
   });
 
+  it("should build and start from a Dockerfile with build arguments", async () => {
+    const context = path.resolve(__dirname, "..", "docker-with-buildargs");
+    const container = await GenericContainer.fromDockerfile(withContext(context), withBuildArg("VERSION", "10-alpine"));
```
cristianrgreco: Instead of exposing different ways to compose options, I think it'd be great if we could turn this into a builder. For example:

```
GenericContainer.fromDockerfile(context)
  ..withBuildArg(...)
  ..withX(...)
```

That way we don't have to expose different options and clients can use the type system to create their Dockerfile

src/generic-container.ts
```diff
@@ -14,17 +15,28 @@ import { RandomUuid, Uuid } from "./uuid";
...
+    const image = opts.uuid.nextUuid();
+    const tag = opts.uuid.nextUuid();
     const repoTag = new RepoTag(image, tag);
-    const dockerClient = dockerClientFactory.getClient();
-    await dockerClient.buildImage(repoTag, context);
+    const dockerClient = opts.dockerClientFactory.getClient();
+    await dockerClient.buildImage(repoTag, opts);
     const container = new GenericContainer(image, tag);
+
+    // tslint:disable-next-line
```
cristianrgreco: Is it possible to keep tslint enabled?

cristianrgreco:
Once we know the final design of the API, it would be good to add an example in the README

src/generic-container.test.ts
dweidenfeld: You can also use the type system with this config option. I understand you, but this way it was a smaller API change, than switching to a completely different pattern ... args vs builder.

But we can switch it to a builder pattern if you like it more ;)
I'll try to do this within the next 1-2 weeks.

dweidenfeld: Okay. I moved everything to a builder pattern and added an example to the readme ;)

src/generic-container.ts
dweidenfeld: removed :)

README.md
```diff
@@ -55,7 +55,7 @@ const { GenericContainer } = require("testcontainers");
 (async () => {
   const buildContext = path.resolve(__dirname, "my-dir");
   
-  const container = await GenericContainer.fromDockerfile(buildContext);
+  const container = await GenericContainer.fromDockerfile(buildContext).build();
```
cristianrgreco: Will need a major release for this as it's a breaking change

src/index.ts
```diff
@@ -1,3 +1,3 @@
-export { GenericContainer } from "./generic-container";
+export { GenericContainer, GenericContainerBuilder } from "./generic-container";
```
cristianrgreco: What's the use case for exposing this?

src/typings.d.ts
```diff
@@ -0,0 +1,9 @@
+declare module "default-gateway" {
```
cristianrgreco: There is already a type definition for this: `npm install @types/default-gateway --save-dev`

src/generic-container.ts
```diff
@@ -19,20 +29,52 @@ import {
...
-    dockerClientFactory: DockerClientFactory = new DockerodeClientFactory()
-  ): Promise<GenericContainer> {
-    const image = uuid.nextUuid();
-    const tag = uuid.nextUuid();
+export class GenericContainerBuilder {
+  private factory: DockerClientFactory;
+  private uuid: Uuid;
+  private buildArgs: BuildArgs;
+
+  constructor(private context: BuildContext) {
```
cristianrgreco: The constructor style used elsewhere in the project is more succinct and marks the fields as readonly:

```typescript
constructor(private readonly context: BuildContext, ...) {
}
```

src/generic-container.test.ts
```diff
@@ -104,4 +104,59 @@ describe("GenericContainer", () => {
...
+    await expect(fetch(url)).rejects.toThrowError();
+  });
+
+  it("should not build and start from a Dockerfile with missing build arguments", async () => {
+    const context = path.resolve(__dirname, "..", "docker-with-buildargs");
+
+    await expect(GenericContainer.fromDockerfile(context).build()).rejects.toThrowError("Failed to build image");
+  });
+
+  it("should work for mysql", async () => {
```
cristianrgreco: These tests have been removed in master, not sure if there was an issue with the merge

README.md
dweidenfeld: Unfortunately yes. This was the cause I previously chose a different style, but I don't see another way :P

README.md
```diff
@@ -65,6 +65,79 @@ const { GenericContainer } = require("testcontainers");
 })();
 ```
 
+Passing arguments to your own Docker image:
```
cristianrgreco: I'm not sure why these are showing as additions?

src/index.ts
dweidenfeld: When you want to temporary store the builder somewhere and you want to have type-safety you need to export the builder type.

README.md
cristianrgreco: I also don't see any other way 👍 though I do think the builder pattern is the best way to go

src/typings.d.ts
dweidenfeld: Nice I haven't seen this. I'll update the PR

src/generic-container.test.ts
dweidenfeld: Why have they been removed? They're actually working great :)

README.md
dweidenfeld: Not sure either...

src/generic-container.test.ts
cristianrgreco: They were added when we had reported issues with mysql and couchdb to reproduce. Eventually we found the issue was with port padding in the host port check and a test was added specifically for that. These tests no longer add value and are expensive as the test will likely have to pull these images. We're trying to only use a single image in the build for performance reasons

dweidenfeld:
Okay, you've already merged another PR that will completely break this PR. I think I'm going to maintain an own version of testcontainers right now, since there is too much stuff going on here that is not really controllable.

Otherwise I have to remove a feature from master that has been merged previously. If you're still interested, just ask and I can do the change ;)

cristianrgreco:
Not sure what's broken, but if that's true then it was either this PR or the other which would break 🙂 My main concern at the moment is that there's a merge that doesn't look quite right. The original issue is still open so no worries if you don't have time to work on this PR, if so then please close any related PRs

dweidenfeld:
I was able to almost resolve the issues. Just tell me what tests you want to remove, but I think if we wait any longer with this PR it will be more complicated to merge it anyway.

src/generic-container.test.ts
dweidenfeld: Okay I removed the tests from master ;)

dweidenfeld:
All issues should be resolved now :smile: 
I misunderstood one change you had in master. So no break here, yet ;)

src/generic-container.ts
```diff
@@ -29,20 +30,52 @@ import {
...
-    dockerClientFactory: DockerClientFactory = new DockerodeClientFactory()
-  ): Promise<GenericContainer> {
-    const image = uuid.nextUuid();
-    const tag = uuid.nextUuid();
+export class GenericContainerBuilder {
+  private factory: DockerClientFactory;
+  private uuid: Uuid;
+  private buildArgs: BuildArgs;
+
+  constructor(private readonly context: BuildContext) {
```
cristianrgreco: Change constructor to:
```typescript
constructor(
  private readonly context: BuildContext,
  private readonly factory: DockerClientFactory = new DockerodeClientFactory(),
  private readonly uuid: Uuid = new RandomUuid()
) {}
```
And remove the fields

cristianrgreco:
I will wait to merge https://github.com/testcontainers/testcontainers-node/pull/35 so that you don't have to resolve anymore conflicts 🙂

src/generic-container.ts
dweidenfeld: But since we already have the API break because of the `.build()` I think it is the best way to only have required parameters in the constructor and pass the others the same way as the new parameters?!

cristianrgreco: Providing defaults makes them optional.

dweidenfeld: okay done

PR #24, opened 2019-08-19
feat(options): added named containers

dweidenfeld:
This commit enables you to name your containers and abort building a 
fresh image if it is already existing, so that you can speed up your
builds.

This PR is based upon https://github.com/testcontainers/testcontainers-node/pull/22. So please check if https://github.com/testcontainers/testcontainers-node/pull/22 is merged beforehand ;)


dweidenfeld:
@cristianrgreco I adjusted this PR to also use the builder pattern from #22

cristianrgreco:
Hi @dweidenfeld 

If I'm not mistaken this functionality has been addressed by https://github.com/testcontainers/testcontainers-node/pull/25

dweidenfeld:
@cristianrgreco 
No, #25 defines the container name for the runtime. This PR will make sure, that you'll build a container with an exact image name & tag with `fromDockerfile()`. You can also make sure, that you do not build the container twice if it already exists on the system etc.
Both options are very useful and I am very happy that #25 arrived, but this PR will improve build/testing performance by far.

My use case for this is:
I have a microservice architecture and I am testint myservice against mock versions of other services that will be requested during the tests. Those mocks are generated with `fromDockerfile()`, but it rebuilds them everytime I start the tests. With this PR I have the ability to name the image and check if the image already exists so there is no need to initiate a rebuild or rely on Docker caching ;)

dweidenfeld:
I adjusted the PR to make it work with the current master :)

src/docker-client.ts
```diff
@@ -103,9 +108,21 @@ export class DockerodeClient implements DockerClient {
...
+    repoTag: RepoTag,
+    context: BuildContext,
+    buildArgs: BuildArgs,
+    abortBuildOnExistingImage: boolean
+  ): Promise<void> {
     log.info(`Building image '${repoTag.toString()}' with context '${context}'`);
 
+    if (abortBuildOnExistingImage) {
+      const image: Dockerode.Image = this.dockerode.getImage(repoTag.toString());
+      if (!!image.id) {
```
cristianrgreco: Let's do an explicit check instead of coercing the value to a boolean, eg `image.id !== undefined`

src/generic-container.test.ts
```diff
@@ -149,4 +149,15 @@ describe("GenericContainer", () => {
       await startedContainer.stop();
     });
   });
+
+  it("should build and start named container from a Dockerfile", async () => {
```
cristianrgreco: There's a `describe` block for `fromDockerfile` functionality

src/generic-container.ts
```diff
@@ -35,6 +35,9 @@ import { HostPortWaitStrategy, WaitStrategy } from "./wait-strategy";
 
 export class GenericContainerBuilder {
   private buildArgs: BuildArgs = {};
+  private imageName: string | null = null;
```
cristianrgreco: Can we not use optional syntax here? `imageName?`

README.md
```diff
@@ -62,6 +62,28 @@ const startedContainer = await container
...
+```javascript
+const path = require("path");
+const { GenericContainer } = require("testcontainers");
+
+const buildContext = path.resolve(__dirname, "dir-containing-dockerfile");
+
+const container = await GenericContainer.fromDockerfile(buildContext)
+  .withImageName("my-image")
+  .withImageTag("my-tag")
+  .skipBuildOnExistingImage()
```
cristianrgreco: I think we're missing something here with the design because `skipBuildOnExistingImage` only has an effect if you provide one of `withImageName` or `withImageTag`. 

I wonder if we could better reflect that. For example:

```typescript
const container = await GenericContainer.fromDockerfile(buildContext)
  .withImageBuilder()
    .withName("my-image")
    .withTag("my-tag")
    .skipBuildOnExistingImage()
    .build()
  .withBuildArg("arg-key", "arg-value")
  .build()
```

I also wonder if it's really a use-case that anyone would want the build to fail if there already exists that container name and tag, maybe it's not required?

src/generic-container.ts
dweidenfeld: Sure. I just forgot about it. I'll now stucked in meetings etc. but hopefully can take care of it later that day ;)

README.md
dweidenfeld: Very good idea!

src/docker-client.ts
dweidenfeld: fixed ;)

src/generic-container.test.ts
dweidenfeld: Must be a merge conflict, just moved it to the right place :)

src/generic-container.ts
dweidenfeld: done

README.md
dweidenfeld: Done! I also forwarded the `Uuid` to the nested builder because it is only used there. I still made it passable by the parent constructor, to not break existing logic ;)

cristianrgreco:
@dweidenfeld Looks great! 

Only thing I'm not sure of is `.skipBuildOnExistingImage`, I can't imagine a use case where a person is building an image that may possibly already exist and wants the build to fail if it does.

IMO we should drop it for now and add it if there's a request for it.

dweidenfeld:
I am actually using `.skipBuildOnExistingImage`. The use case is:

I am testing a microservice in combination with other services. In order to not have to start all services during integration testing I switch to contract-first development. So imagine I am currently working on service A and it requires service B. I'll generate a contract for service B and service A automatically generates a mock server based on the contract of service B for service A. So when I start the tests I am testing against a mocked version of service B, so I don't have to actually run service B.
The problem is now when I am writing those tests I may want to execute the tests multiple times and without the option it is rebuilding the mock-server for service B before every test run. The only thing that speeds up the process is the Docker layer caching. 
Unfortunately the detection of what has been cached already takes a little time and I can avoid this by using `.skipBuildOnExistingImage`.

So basically it is a performance optimization for developing the tests.

cristianrgreco:
I think I understand, my suggestion is to enable that behaviour by default, so if the image already exists, it won't rebuild it, and won't result in a failure

dweidenfeld:
Okay I changed the option to be a `.forceRebuild()` and by default it uses the "cached" version.
The option never fails it only checks if it has to rebuild or not.

src/image-name-builder.ts
```diff
@@ -0,0 +1,41 @@
+import { GenericContainerBuilder } from "./generic-container-builder";
+import { Uuid } from "./uuid";
+
+export class ImageNameBuilder {
```
cristianrgreco: `ImageBuilder`?

dweidenfeld: :+1: done

dweidenfeld:
sorry for the delay, but I fixed the build error ;)

jayarjo:
I kind of thought it is already a default behaviour. Why is this not merged? I mean one would expect `testcontainers` to re-use as much as possible by default.

jayarjo:
@cristianrgreco building fresh image each time is terrible hole in the whole procedure. At the very least temp image should be removed after container is stopped and removed.

Also, from what I see there's nice interface for this kind of thing here: https://www.testcontainers.org/features/creating_images/. Why coming up with something different @dweidenfeld?

@cristianrgreco is this a port of TestContainer Java, or something else? Weren't you advised to follow their interfaces? Just wonder. Cause it kind of looks like a port, but diverges quite dramatically in some places.

jayarjo:
Btw, I thought, why not use a checksum of Dockerfile contents instead of just random name? That way it will be both unique per Dockerfile and the same for the same Dockerfile. Kind of kills both rabbits.

PR #25, opened 2019-08-19
Add optional name parameter for container creation options

Yengas:
In certain test environments, it may be possible that the `container.stop` may not be called. This could cause zombie containers. So it is nice to somehow keep track of containers created by testcontainers-node. Easiest way could be giving the option of setting container names. 

This PR adds that feature, so when you are running a container, you can set a name. I've tried to stay close to your coding standards, as much as possible. Please do comment if there are any discrepancies.

### DIFF
--- a/src/docker-client.ts
+++ b/src/docker-client.ts
@@ -9,6 +9,7 @@ import { PortString } from "./port";
 import { RepoTag } from "./repo-tag";
 
 export type Command = string;
+export type ContainerName = string;
 export type ExitCode = number;
 
 export type EnvKey = string;
@@ -24,7 +25,7 @@ type DockerodeExposedPorts = { [port in PortString]: {} };
 
 export interface DockerClient {
   pull(repoTag: RepoTag): Promise<void>;
-  create(repoTag: RepoTag, env: Env, boundPorts: BoundPorts, cmd: Command[]): Promise<Container>;
+  create(repoTag: RepoTag, env: Env, boundPorts: BoundPorts, cmd: Command[], name?: ContainerName): Promise<Container>;
   start(container: Container): Promise<void>;
   exec(container: Container, command: Command[]): Promise<ExecResult>;
   buildImage(repoTag: RepoTag, context: BuildContext): Promise<void>;
@@ -41,9 +42,15 @@ export class DockerodeClient implements DockerClient {
     await streamToArray(stream);
   }
 
-  public async create(repoTag: RepoTag, env: Env, boundPorts: BoundPorts, cmd: Command[]): Promise<Container> {
+  public async create(
+    repoTag: RepoTag,
+    env: Env,
+    boundPorts: BoundPorts,
+    cmd: Command[],
+    name?: ContainerName
+  ): Promise<Container> {
     log.info(`Creating container for image: ${repoTag}`);
-
+    const containerNameOption = name !== undefined ? { name } : {};
     const dockerodeContainer = await this.dockerode.createContainer({
       Image: repoTag.toString(),
       Env: this.getEnv(env),
@@ -51,7 +58,8 @@ export class DockerodeClient implements DockerClient {
       Cmd: cmd,
       HostConfig: {
         PortBindings: this.getPortBindings(boundPorts)
-      }
+      },
+      ...containerNameOption
     });
 
     return new DockerodeContainer(dockerodeContainer);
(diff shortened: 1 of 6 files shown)


cristianrgreco:
Hi @Yengas, thanks for raising this PR!

I noticed that testcontainers-java stores a list of generated container names in a static list, and adds a shutdown hook to stop them all. Would that be a better solution? I propose it because in this case, the user wouldn't expect the container not to be managed automatically because they didn't give it a name.

What do you think?

cristianrgreco:
There also seems to be some cross over with #24

Yengas:
> Hi @Yengas, thanks for raising this PR!
> 
> I noticed that testcontainers-java stores a list of generated container names in a static list, and adds a shutdown hook to stop them all. Would that be a better solution? I propose it because in this case, the user wouldn't expect the container not to be managed automatically because they didn't give it a name.
> 
> What do you think?

Hey,

I can give a more concrete example, and a use-case. I am using Jest for starting up some containers, before i run my integration tests. However when i force the test processes to exit(process.exit(0) just to test), the main process which i run the *testcontainers* in, dies aswell. And as far as i see, i can not register to *exit*, *SIGTERM* etc. handles on the main thread(whilst using Jest see [jest#6029](https://github.com/facebook/jest/issues/6029)). So i can not stop the started containers.

The worst is when i use Docker-out-of-Docker in the Jenkins pipeline. If this situation were to happen, and the node process somehow exits without stopping the container, i would not know which containers were started by this library. I want to put a safety net, by starting the containers with a known prefix from the pipeline, so i can manually stop them after my integration tests exits.

Besides using Docker inside Docker, i could not think of a way to do this cleanly but give prefix to the containers.

Yengas:
> There also seems to be some cross over with #24

I thought this was unrelated to my PR because it seems to be related to *building* images, instead of *running* them. But i would be happy to fix any conflicts, in case there are.

Yengas:
@cristianrgreco on a side note, can you check [jest-testcontainers](https://github.com/Trendyol/jest-testcontainers)? We've just open sourced it. The need was running integration tests over neo4j. This seemed much more cleaner compared to using ansible/docker-compose.

Yengas:
fixed conflicts

src/docker-client.ts
```diff
@@ -54,6 +56,7 @@ export class DockerodeClient implements DockerClient {
   public async create(options: CreateOptions): Promise<Container> {
     log.info(`Creating container for image: ${options.repoTag}`);
 
+    const containerNameOption = options.name !== undefined ? { name: options.name } : {};
```
cristianrgreco: The `name` field in Dockerode's `ContainerCreateOptions` is also optional, so we don't need to transform it. 
This should work:

```typescript
const dockerodeContainer = await this.dockerode.createContainer({
      name: options.name,
      Image: options.repoTag.toString(),
      ...
```

src/generic-container.ts
```diff
@@ -64,21 +75,33 @@ export class GenericContainer implements TestContainer {
...
     const inspectResult = await container.inspect();
     const containerState = new ContainerState(inspectResult);
     await this.waitForContainer(container, containerState, boundPorts);
 
-    return new StartedGenericContainer(container, this.dockerClient.getHost(), boundPorts, this.dockerClient);
+    return new StartedGenericContainer(
+      container,
+      this.dockerClient.getHost(),
+      boundPorts,
+      inspectResult.name,
```
cristianrgreco: Why not use `this.name`? That way we don't have to also add it to the `InspectResult` https://github.com/testcontainers/testcontainers-node/pull/25/files#diff-691ac3f0f06fddc7f3079b1a0f835425R12

src/test-container.ts
```diff
@@ -33,6 +33,7 @@ export interface StartedTestContainer {
   stop(options?: OptionalStopOptions): Promise<StoppedTestContainer>;
   getContainerIpAddress(): Host;
   getMappedPort(port: Port): Port;
+  getName(): ContainerName;
```
cristianrgreco: Is there a use case for exposing this?

cristianrgreco:
Thanks for updating this PR @Yengas.
I've just left a couple of comments/questions

src/generic-container.ts
Yengas: Containers are given 2 word names by default. And also it seems they [may contain](https://github.com/moby/moby/issues/6705#issuecomment-47298276) some prefix that *testcontainers-node* can not know about. So better the get it with inspect, rather than making the assumption the name is what we have given to the dockerode.

src/test-container.ts
Yengas: Just for the sake of testing and for the fact that every field of *InspectResult* is exposed with a getter. Per [my previous comment](https://github.com/testcontainers/testcontainers-node/pull/25/files#r319774038), it may be necessary in cases where we would like to know what the docker engine named the container.

PR #34, opened 2019-08-29
Expose container stop and remove options in GenericContainer

cristianrgreco:
https://github.com/testcontainers/testcontainers-node/pull/31#discussion_r319225306

### DIFF
--- a/src/generic-container.ts
+++ b/src/generic-container.ts
@@ -139,7 +139,7 @@ class StartedGenericContainer implements StartedTestContainer {
   ) {}
 
   public async stop(options: OptionalStopOptions = {}): Promise<StoppedTestContainer> {
-    const resolvedOptions = { ...DEFAULT_STOP_OPTIONS, options };
+    const resolvedOptions = { ...DEFAULT_STOP_OPTIONS, ...options };
     await this.container.stop({ timeout: resolvedOptions.timeout });
     await this.container.remove({ removeVolumes: resolvedOptions.removeVolumes });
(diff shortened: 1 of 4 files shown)


src/index.ts
```diff
@@ -1,2 +1,3 @@
 export { GenericContainer } from "./generic-container";
+export { TestContainer, StartedTestContainer, StoppedTestContainer } from "./test-container";
```
markmsmith: If you also export DEFAULT_STOP_OPTIONS then users can just override the parts they care about, eg
```ts
await testContainer.stop({
  ...DEFAULT_STOP_OPTIONS,
  timeout: 0
});
```

markmsmith: Alternatively, you could do a merge between the passed options and the full set of options, which is probably nicer.

src/generic-container.ts
```diff
@@ -132,9 +138,10 @@ class StartedGenericContainer implements StartedTestContainer {
     private readonly dockerClient: DockerClient
   ) {}
 
-  public async stop(): Promise<StoppedTestContainer> {
-    await this.container.stop();
-    await this.container.remove();
+  public async stop(options: OptionalStopOptions = {}): Promise<StoppedTestContainer> {
+    const resolvedOptions = { ...DEFAULT_STOP_OPTIONS, options };
```
markmsmith: I think you need a spread on `options` as well.

cristianrgreco: D'oh, thanks

PR #46, opened 2020-01-22
Add ID getter to GenericContainer

castarco:
This minor change allows for better interoperability with `dockerode`, as this library mainly relies on the container ID to instantiate "Container" objects.

### DIFF
--- a/src/container.ts
+++ b/src/container.ts
@@ -4,7 +4,7 @@ import { Duration, TemporalUnit } from "node-duration";
 import { Command, ContainerName, ExitCode } from "./docker-client";
 import { Port } from "./port";
 
-type Id = string;
+export type Id = string;
 
 export type InspectResult = {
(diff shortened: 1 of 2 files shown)


cristianrgreco:
Testcontainers uses Dockerode under the hood, the way containers are orchestrated isn't a concern of testcontainers, what's the use casing for exposing this information?

castarco:
## Case 1
What you already saw in the PR #47. The specific test for that new feature is excessively complicated because not being able to access the container ID, I'm being forced to obtain it indirectly through its name.

I admit that Dockerode could also implement a method to obtain this information through the container name, but that's another story.

## Case 2
If the feature provided by the PR #47 is not accepted, the only clean way to retrieve containers and connect them to a specific network (using dockerode) is through their ID.

Regarding on why I do care about specific networks, that's a use case for complex microservices scenarios.

PR #47, opened 2020-01-22
Enable users to set NetworkMode for containers

castarco:
This PR enables users to configure containers to run in specific networks. This is quite useful in complex CI setups with microservices.

### DIFF
--- a/src/docker-client.ts
+++ b/src/docker-client.ts
@@ -10,6 +10,7 @@ import { RepoTag } from "./repo-tag";
 
 export type Command = string;
 export type ContainerName = string;
+export type NetworkMode = string;
 export type ExitCode = number;
 
 export type EnvKey = string;
@@ -44,6 +45,7 @@ type CreateOptions = {
   tmpFs: TmpFs;
   boundPorts: BoundPorts;
   name?: ContainerName;
+  networkMode?: NetworkMode;
 };
 
 export interface DockerClient {
@@ -75,6 +77,7 @@ export class DockerodeClient implements DockerClient {
       ExposedPorts: this.getExposedPorts(options.boundPorts),
       Cmd: options.cmd,
       HostConfig: {
+        NetworkMode: options.networkMode,
         PortBindings: this.getPortBindings(options.boundPorts),
         Binds: this.getBindMounts(options.bindMounts),
         Tmpfs: options.tmpFs
--- a/src/generic-container.test.ts
+++ b/src/generic-container.test.ts
@@ -36,25 +36,14 @@ describe("GenericContainer", () => {
 
   it("should set network mode", async () => {
     const container = await new GenericContainer("cristianrgreco/testcontainer", "1.1.11")
-      .withName("special-test-container")
       .withNetworkMode("host")
       .start();
 
     const dockerodeClient = new Dockerode();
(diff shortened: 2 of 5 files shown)


src/generic-container.test.ts
```diff
@@ -33,6 +34,31 @@ describe("GenericContainer", () => {
...
+      .withNetworkMode("host")
+      .start();
+
+    const dockerodeClient = new Dockerode()
+    const listedContainers = await dockerodeClient.listContainers()
+
+    let containerInfo: undefined | Dockerode.ContainerInfo
+    for (const listedContainer of listedContainers) {
+      if (listedContainer.Names.includes("/special-test-container")) {
+        containerInfo = listedContainer
```
castarco: The way we obtain `containerInfo` can be drastically simplified if something like what the PR #46 provides is implemented.

cristianrgreco: I'd like to keep the tests as 'black box' tests, where we test a behaviour. It is for this reason as well that we don't currently use Dockerode in our tests. This test only checks that a configuration parameter is passed to Dockerode.

castarco: It does not check that a configuration is passed to Dockerode, it tests that the configuration reached the docker daemon.

Dockerode obtains this information on its own making a call to the docker daemon, not recovering the objects that we created previously.

cristianrgreco: My point is that the existing tests test a given behaviour. If we swapped dockerode for something else the tests would still pass. Ideally we shouldn't have to dig into the internals to test that something works.

castarco: I'm not digging into the internals, I use a tool that, luckily for me, you also use for the internals. But I'm not accessing the internals of the classes in any way.

src/docker-client.ts
```diff
@@ -10,6 +10,7 @@ import { RepoTag } from "./repo-tag";
 
 export type Command = string;
 export type ContainerName = string;
+export type NetworkMode = string
```
cristianrgreco: `NetworkMode = 'bridge' | 'host' | 'overlay'` ?

castarco: That would be awesome, but it's also possible to create private networks in docker with other names, so the best we can do is to leave it open to almost anything.

castarco: Ok, it seems I was _partially_ right, but my guess was good enough:
Screenshot from 2020-01-22 19-13-38

cristianrgreco:
Could you please update the README to provide an example?

castarco:
Yes, of course :) .

cristianrgreco:
Your PR to expose the ID has also been merged, in case you'd rather use that in the test

castarco:
I force-pushed because I rebased against master.

src/test-container.ts
```diff
@@ -35,6 +36,7 @@ export interface StartedTestContainer {
   getContainerIpAddress(): Host;
   getMappedPort(port: Port): Port;
   getName(): ContainerName;
+  getId(): ContainerId;
```
castarco: It was my mistake not doing this in the PR #46 .

cristianrgreco:
Published, it's available in 2.3.0

PR #48, opened 2020-01-23
fix: add network mode to TestContainer

kincaidoneil:
(Side note: I ran into an issue where I needed to change the network mode, and started looking through the source code to see how I'd do it, but then found... it was just added an hour earlier! 😄 Great timing @castarco @cristianrgreco)


cristianrgreco:
Thanks for the fix @kincaidoneil, it's available in 2.3.1 🙂

Issue #51, opened 2020-02-12
TestContainer fluent API returning 'this' type


r3mi:
Hello,
thanks for this useful library.
I am trying to extend (subclass) `GenericContainer` but the current typing for the `withXxx` method gets in the way. To be able to extend the fluent API, these methods should return the actual type of Container class, not the base `TestContainer`.
In other words, simply return a [`this type`](https://www.typescriptlang.org/docs/handbook/advanced-types.html#polymorphic-this-types) from the methods e.g.
```ts
export interface TestContainer {
  withEnv(key: EnvKey, value: EnvValue): this;
  withCmd(cmd: Command[]): this;
  // ...
}
```

cristianrgreco:
Hi @r3mi 

Could you show me an example of where it's getting in the way so that I can better understand the problem? I'd guess that the base interface of `TestContainer` is fine as `GenericContainer` implements it anyways. So you'd end up with `CustomGenericContainer -> GenericContainer -> TestContainer`.

r3mi:
Hi @cristianrgreco 
sorry my issue was a bit short. Here are simplified examples:
```ts
class CustomGenericContainer extends GenericContainer {
  ...
  withSomething(): this
  ...
}
const container = await new CustomGenericContainer()
    .withStartupTimeout(new Duration(20, TemporalUnit.SECONDS))
    .withSomething() // <-- HERE can't compile because we are back on a TestContainer
    .start()
```
Another example:
```ts
class CustomGenericContainer extends GenericContainer {
  ...
  start(): Promise<CustomStartedTestContainer>
  ...
}
const container = await new CustomGenericContainer()
    .withStartupTimeout(new Duration(20, TemporalUnit.SECONDS))
    .start()
container.customMethod() // <-- HERE can't compile, container is not expected "CustomStartedTestContainer"
```
Both examples would be fine if `withStartupTimeout` was declared to return '`this` type' instead of `TestContainer`.

cristianrgreco:
@r3mi Please review https://github.com/testcontainers/testcontainers-node/pull/52

cristianrgreco:
@r3mi Available in 2.4.0. Could you have a try and let me know?

r3mi:
works fine, thank you !

PR #52, opened 2020-02-14
Allow extension of GenericContainer by returning `this` in builder

cristianrgreco:
Address #51

### DIFF
--- a/src/generic-container.ts
+++ b/src/generic-container.ts
@@ -122,47 +122,47 @@ export class GenericContainer implements TestContainer {
     );
   }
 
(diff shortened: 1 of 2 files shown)


r3mi:
thanks for the quick change. That's fine for my purpose. I feel it would be more complete and generalised if the same change was also applied to the `TestContainer` interface.

cristianrgreco:
Thanks @r3mi, I wasn't aware you could reference `this` from an interface. Please have another look.

r3mi:
all fine, thanks !

PR #57, opened 2020-02-27
60 seconds is not enough for the Gitlab CI. 120 seems to work fine

antoniocordasco:


cristianrgreco:
Hi @antoniocordasco 

Does the following not work for you?

```javascript
const container = await new GenericContainer("...")
  .withStartupTimeout(new Duration(2, TemporalUnit.MINUTES))
  .start();
```

I don't think we keep updating the default timeout each time a slower CI environment is found.

antoniocordasco:
it does! 
thanks for your help. You have no idea how much time I have spent debugging this :-)
I also sent a request for help the wrong github project by mistake https://github.com/nodejs/docker-node/issues/1220

But in the end I got it working and that's what matters.
Thanks for checking this PR.

jeantil:
@antoniocordasco if you could document somewhere how you got it working, I would be very interested. I encoutered the same issue where testcontainer tries to send http queries to a tls enabled dind on gitlab ci and haven't figured it out yet.

antoniocordasco:
hi @jeantil in the end, it was a very simple fix for me, to sort out the timeout issue. I simply did wha t @cristianrgreco suggested, which is to add .withStartupTimeout(new Duration(2, TemporalUnit.MINUTES)) so that it runs form 120 seconds instead of 60. Have you tried this? 
If your issue is not specifically with timeout, give us some more details and I'll see if I can help.

jeantil:
Well i am trying to use  docker in docker on gitlab ci, ideally with TLS enabled. but when I follow the recommended configurations steps for dind+tls in the gitlab documentation, i get 
>    Error: (HTTP code 400) bad parameter - Client sent an HTTP request to an HTTPS server
which is the error you reported in nodejs/docker-node#1220 where you state that it should have been opened here :)

antoniocordasco:
right I see. That sounds familiar. I have had a series of issues while setting this up and managed to fix hem one by one. Let me have a dig around tomorrow and get back to you.

antoniocordasco:
sorry for the delay.
This is what I have in my .gitlab-ci.yml file:

variables:
  DOCKER_DRIVER: overlay2

.quality:
  stage: Quality
  image:  $RUNNER_IMAGE
  before_script:
    - cp -r /app/node_modules . 

Unit test:
  extends: .quality
  script:
    - sleep 10
    - DEBUG=testcontainers yarn test
  variables:
    DOCKER_HOST: 'tcp://docker:2375'
    SESSION_SECRET: ...
  services:
    - docker:dind


I don't have the variables DOCKER_TLS_CERTDIR, DOCKER_CERT_PATH, DOCKER_TLS_VERIFY set anywhere.
Could you try this set up and see if it works?

jeantil:
That setup is indeed using http instead of https that's what I have at the
moment :) I was hoping you got https (thus tls) working.
Thanks for taking the time, I very much appreciate it

Le jeu. 21 mai 2020 à 11:16, antoniocordasco <notifications@github.com> a
écrit :

antoniocordasco:
oh I see.
No unfortunately I only got this working, but in my case it is enough.
You're welcome

PR #58, opened 2020-03-11
Option to provide additional host configuration

sweiler:
In the Java implementation, it is possible to modify the `docker run` command before it is being executed with `withCreateContainerCmdModifier`. So far there is no similar option in testcontainers-node.

This PR adds an option to add additional host parameters, exposing the underlying dockerode API. This is very similar to how the Java implementation with the above mentioned `withCreateContainerCmdModifier` exposes the underlying docker-java API. (See [here](https://github.com/testcontainers/testcontainers-java/blob/6622578bed5e66bfb979b9bfeeb0e22321e21266/core/src/main/java/org/testcontainers/containers/GenericContainer.java#L1351))

The specific reason why we would need this feature, is that we want to run a test using testcontainers on a docker host with splunk set as the log driver. But Splunk is a write-only log driver, which leads to errors when waiting for a log line in testcontainers. With this PR, we could set the log-driver for the test explicitly to json-file.

### DIFF
--- a/README.md
+++ b/README.md
@@ -181,6 +181,20 @@ const container = await new GenericContainer("alpine")
   .start();
 ```
 
+You can override the logging driver used by docker to be the default one (json-file).
+This might be necessary when the driver of your docker host does not support reading logs
+and you want to use the `Wait.forLogMessage` wait strategy.
+
+This is the same as [--log-driver json-file on docker run](https://docs.docker.com/config/containers/logging/configure/#configure-the-logging-driver-for-a-container).
+
+```javascript
+const { GenericContainer } = require("testcontainers");
+
+const container = await new GenericContainer("redis")
+  .withDefaultLogDriver()
+  .start();
+```
+
 Testcontainers will wait 10 seconds for a container to stop, to override:
 
 ```javascript
--- a/src/docker-client.ts
+++ b/src/docker-client.ts
@@ -1,4 +1,4 @@
-import Dockerode, { PortMap as DockerodePortBindings } from "dockerode";
+import Dockerode, { HostConfig, PortMap as DockerodePortBindings } from "dockerode";
 import { Duration, TemporalUnit } from "node-duration";
 import streamToArray from "stream-to-array";
 import tar from "tar-fs";
@@ -53,6 +53,15 @@ export type BindMount = {
 };
 type DockerodeBindMount = string;
 
+export type LogConfig = {
+  logDriver: string;
+  logOpts?: object;
+};
+type DockerodeLogConfig = {
+  Type: string;
+  Config: any;
+};
+
 type CreateOptions = {
   repoTag: RepoTag;
   env: Env;
@@ -63,6 +72,7 @@ type CreateOptions = {
   name?: ContainerName;
   networkMode?: NetworkMode;
   healthCheck?: HealthCheck;
+  useDefaultLogDriver: boolean;
 };
 
 export interface DockerClient {
@@ -99,7 +109,8 @@ export class DockerodeClient implements DockerClient {
         NetworkMode: options.networkMode,
         PortBindings: this.getPortBindings(options.boundPorts),
         Binds: this.getBindMounts(options.bindMounts),
-        Tmpfs: options.tmpFs
+        Tmpfs: options.tmpFs,
+        LogConfig: this.getLogConfig(options.useDefaultLogDriver)
       }
     });
 
@@ -203,5 +214,16 @@ export class DockerodeClient implements DockerClient {
 
   private getBindMounts(bindMounts: BindMount[]): DockerodeBindMount[] {
     return bindMounts.map(({ source, target, bindMode }) => `${source}:${target}:${bindMode}`);
+  }
+
+  private getLogConfig(useDefaultLogDriver: boolean): DockerodeLogConfig | undefined {
+    if (!useDefaultLogDriver) {
+      return undefined;
+    }
+
+    return {
+      Type: "json-file",
+      Config: {}
+    };
   }
 }
--- a/src/generic-container.test.ts
+++ b/src/generic-container.test.ts
@@ -160,6 +160,23 @@ describe("GenericContainer", () => {
     await container.stop();
   });
 
+  it("should set default log driver", async () => {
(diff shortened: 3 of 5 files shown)


cristianrgreco:
Hi @sweiler, thanks for the PR!

Would you mind also updating the README to show an example of how it can be used?

I'm not sure if it should be called `additionalHostConfig` instead of for example `hostConfig`, or even something more specific in this case like `withLogConfig`. One of the goals of testcontainers is to abstract away the details of Docker, and using `withHostConfig` would mean a user would need to look at the docs to know what (including what types) can be provided.

I think in this case providing a specific, typed API for building the container is better. What do you think?

sweiler:
Thanks for the quick response :)

I'm happy to update the README as well.
I decided to go with the very generic host config, because I didn't want to add an option only for my own case. Also I thought it would be more aligned with the way testcontainers-java does it, which is also the very generic 'all options available' way. But if you say you'd like a specific `withLogConfig` option more, I'm more than happy to change my PR accordingly.

cristianrgreco:
I think a typed `withLogConfig` is what we want, ideally users shouldn't have to read the Docker docs to know what values to provide

sweiler:
Hey @cristianrgreco, I just updated the PR to contain the withLogConfig. I also changed the naming to be more similar to the docker run command and the docker daemon config, instead of the Dockerode naming.
Also I added a paragraph to the README.

What do you think?

README.md
```diff
@@ -247,3 +247,25 @@ const container = await new GenericContainer("redis")
   .withWaitStrategy(Wait.forHealthCheck())
   .start();
 ```
+
+## Log Config
```
cristianrgreco: This should probably be further up with the other examples, it's currently in the wait strategies section

README.md
```diff
@@ -247,3 +247,25 @@ const container = await new GenericContainer("redis")
...
+when the default driver of your docker host does not support reading logs
+and you want to use the `Wait.forLogMessage` wait strategy.
+
+This is the same as [--log-driver on docker run](https://docs.docker.com/config/containers/logging/configure/#configure-the-logging-driver-for-a-container).
+
+```javascript
+const { GenericContainer, Wait } = require("testcontainers");
+
+const container = await new GenericContainer("redis")
+  .withExposedPorts(6379)
```
cristianrgreco: The `Wait` import and `withExposedPorts` could be removed to simplify the example

README.md
sweiler: You're right, I didn't quite get the structure I have to admit. Moved it further up to a point where I think it makes sense.

src/docker-client.ts
```diff
@@ -53,6 +53,15 @@ export type BindMount = {
 };
 type DockerodeBindMount = string;
 
+export type LogConfig = {
+  logDriver: string;
```
cristianrgreco: I was going through the Docker docs to see what types we can add to this. It seems there's a limited number of logDrivers, but each one has different options. For example for Splunk, we would need to provide splunk-token and splunk-url (https://docs.docker.com/config/containers/logging/splunk/). As a result, what do you think about adding log-driver specific implementations? For example:

`withLogDriverForSplunk(token: String, url: String)`

I'm not sure what your use-case is, but you could add your specific log-driver implementation, and specify the mandatory options?

src/docker-client.ts
```diff
@@ -53,6 +53,15 @@ export type BindMount = {
...
+  logOpts?: object;
+};
+type DockerodeLogConfig = {
+  Type: string;
+  Config: any;
```
cristianrgreco: I think this will always be an `object`

src/generic-container.test.ts
```diff
@@ -160,6 +160,30 @@ describe("GenericContainer", () => {
     await container.stop();
   });
 
+  it("should set log config", async () => {
+    const containerName = "special-test-container";
```
cristianrgreco: Can remove the name expectations, the assertion we care about is on https://github.com/testcontainers/testcontainers-node/pull/58/files#diff-9c365222db7442c7a59145d8c7b621e2R179

src/docker-client.ts
sweiler: Hm, I tend to disagree, as I think that people who need this feature also know quite well what log-driver they want and how the options they need for it are called, as they probably have this configuration already somewhere else.
On the other hand, I think it is quite possible that the use-case for most people is the same I had, which is that you configured a special log-driver on your docker host, but want to use the plain default 'json-file' logger with no special options for testcontainers. So what about having a withDefaultLogDriver() which has no more options and just resets the log driver to json-file with no options, which is the default in docker?

sweiler: @cristianrgreco any updates on this?

sweiler:
Any updates on this @cristianrgreco ?

cristianrgreco:
Sorry @sweiler, completely forgot about this. I think we're missing the `withLogConfig` method in the `TestContainer` interface. Once added I think we're good to go.

sweiler:
No worries, thanks for responding :) I added it now to the interface, so I would be happy to get this merged as it is now.

But I also think your point in one of the comments above about this requiring every user now to know the driver options is still valid. So If you woud prefer the variant `withDefaultLogDriver()` that I proposed over this one, I'm also happy to change it again.

cristianrgreco:
Thanks @sweiler. 

I think I would prefer `withDefaultLogDriver`. As it would allow more specific log driver implementations to be added in the future, like `withSplunkLogDriver()`, etc.

 Although the implementation as it is would allow someone to configure the log driver however they see fit, my concern is that testcontainers isn't helping by abstracting away the Docker internals, it's surfacing them - the user **has** to read the Docker docs to figure out what to provide here

sweiler:
Cool, I changed the PR now accordingly. Could you also check again if I documented it in the README in a way users understand it?

cristianrgreco:
Thanks @sweiler, LGTM. Are you happy for me to merge?

sweiler:
Cool, very happy if you'd merge it :) Thanks for the nice collaboration experience 😊

cristianrgreco:
Thank you @sweiler! The change is available in version 2.7.0.

Issue #62, opened 2020-04-13
[feature-request] Private Registry


AntonovRoman:
### Overview
It would be nice to have the ability to pull images from private registries the same way as `java` lib does.
https://github.com/testcontainers/testcontainers-java/blob/master/core/src/main/java/org/testcontainers/utility/RegistryAuthLocator.java

- Auth from local `~/.docker/config.json`
- Support for `credHelpers` and `credsStore`

cristianrgreco:
Hi @AntonovRoman 

What do you think about something like:

```javascript
new GenericContainer("private-image")
  .withAuthentication({
    username: '...',
    password: '...',
    email: '...',
    serverAddress: '...'
  })
```

AntonovRoman:
@cristianrgreco That's what i was thinking for a short term solution. I can create a PR for that.

cristianrgreco:
@AntonovRoman Available in 2.8.0

AntonovRoman:
@cristianrgreco thank you !

PR #63, opened 2020-04-13
Private Registry Authentication

AntonovRoman:
### Overview

- Add ability to pass auth object to be able to pull private images
- Ref https://github.com/testcontainers/testcontainers-node/issues/62

### DIFF
--- a/src/generic-container.ts
+++ b/src/generic-container.ts
@@ -1,9 +1,9 @@
-import { AuthConfig } from "dockerode";
 import { Duration, TemporalUnit } from "node-duration";
 import { BoundPorts } from "./bound-ports";
 import { Container, Id as ContainerId } from "./container";
 import { ContainerState } from "./container-state";
 import {
+  AuthConfig,
   BindMode,
   BindMount,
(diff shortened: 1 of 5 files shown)


src/generic-container.ts
```diff
@@ -87,6 +86,7 @@ export class GenericContainer implements TestContainer {
   private waitStrategy?: WaitStrategy;
   private startupTimeout: Duration = new Duration(60_000, TemporalUnit.MILLISECONDS);
   private useDefaultLogDriver: boolean = false;
+  private authConfig = {} as AuthConfig;
```
cristianrgreco, line 89: I think we should define our own type to avoid having to cast. 
Probably it'd be an object with similar fields, and the type definition here would be optional, like 
`private authConfig?: AuthConfig`

cristianrgreco:
Hi @AntonovRoman, thanks for the PR! I've just left one comment, other than that LGTM!

src/generic-container.ts
AntonovRoman: Corrected

PR #70, opened 2020-04-22
GenericContainer: add withExposedPort as a way to set pre-defined mapped port

jayarjo:
I think it makes a lot of sense to have this one. In fact we never use random ports on our test containers. It always something pre-defined in the ENV.

### DIFF
--- a/src/bound-ports.ts
+++ b/src/bound-ports.ts
@@ -13,8 +13,9 @@ export class BoundPorts {
     return binding;
   }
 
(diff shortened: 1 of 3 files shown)


cristianrgreco:
HI @jayarjo, fixed ports isn't something we recommend and we want to avoid introducing it. See this blog post for more info: https://bsideup.github.io/posts/testcontainers_fixed_ports/

Issue #66, opened 2020-04-18
support for user-defined networks?


jayarjo:
Any way/plans to support user-defined networks? From what I see `testcontainers-java` has [such functionality](https://www.testcontainers.org/features/networking#advanced-networking).

cristianrgreco:
Yes, it's in the README: `.withNetworkMode("network_name")`

cristianrgreco:
Sorry, I closed the issue by mistake.

jayarjo:
But you can't create the network itself with it, or can you? I thought it was only for well the mode (e.g. bridge, etc)

cristianrgreco:
That's right, currently you cannot create the network with testcontainers. Feel free to raise a PR, if not I'll look into this when I can.

katerina-semikina:
Hi everyone. Looking forward for this useful change. How the things are going?

jayarjo:
> Hi everyone. Looking forward for this useful change. How the things are going?

Вау, девчонка интересуется возможностью создания сетки между докер контейнерами! Клааасс! Я тебе обещаю, если ответишь на этот комментарий я добавлю детали о моем ПиаРе в ридми ) До сих пор никак руки не доходили просто...

katerina-semikina:
> > Hi everyone. Looking forward for this useful change. How the things are going?
> 
> Вау, девчонка интересуется возможностью создания сетки между докер контейнерами! Клааасс! Я тебе обещаю, если ответишь на этот комментарий я добавлю детали о моем ПиаРе в ридми ) До сих пор никак руки не доходили просто...

Это было бы здорово!

jayarjo:
> Это было бы здорово!

Сделано 🙌  
😉

cristianrgreco:
Hi @jayarjo, I've pushed a couple of minor changes, do you want to take a look? If OK I'll release it

jayarjo:
@cristianrgreco well... I'd say not ok, since I took those interfaces and method names from Testcontainers Java, shouldn't we follow them in that?

Also the idea behind `Network.newNetwork()` was that network creation is asynchronous,  so `new Network()` doesn't properly represent the process. Also I didn't want Network instance to be accessible/available before `Network.newNetwork()` resolves. That way you have natural defense in place against people trying to invoke methods on the instance before it's ready to be used. Now that you've changed that, you need to provide some alternatives, like state of the Network maybe or something.

In fact you are using similar pattern for Container instantiation logic, so why break it?

[...]

PR #76, opened 2020-05-09
adding Network

jayarjo:


jayarjo:
Addresses https://github.com/testcontainers/testcontainers-node/issues/66

src/network.test.ts
```diff
@@ -0,0 +1,50 @@
...
+
+    expect(exitCode).toBe(0);
+
+    await container1.stop();
+    await container2.stop();
+    await network.close();
+  });
+
+  it("instantiating Network should throw", () => {
+    expect(() => new Network()).toThrowError("use static newNetwork() method to instantiate network");
```
cristianrgreco, line 48: Why not use the constructor instead of having an additional method?

src/network.ts
```diff
@@ -0,0 +1,47 @@
...
+
+class StartedNetwork {
+  constructor(
+    private readonly id: string,
+    private readonly options: CreateNetworkOptions,
+    private readonly dockerClientFactory: DockerodeClientFactory
+  ) {}
+
+  public getId(): string {
+    return this.id!;
```
cristianrgreco, line 13: Is the `!` required?

src/network.test.ts
jayarjo: That's how it was implemented in `testcontainers-java` - i thought we were meant to take inspiration from there.

src/network.ts
jayarjo: I wasn't sure how to resolve TypeScript complaint here - the problem is that in the beginning `id` is undefined, so it requests to support that case in typings as well I guess.

jayarjo:
> This looks great, thank you! Could you update the README with some examples on how to use it?

Ok, will do.

jayarjo:
@cristianrgreco  I've added an example to README.  It keeps auto-formatting it into the different form though. Let me know if you expected anything else.

Also in my fork, I have whole lot of [other useful methods](https://github.com/jayarjo/testcontainers-node/blob/dist/src/network.ts) added to the Network. Like ability to grab an existing Network byId or byName, inspect it and check if there are any containers attached to it.

cristianrgreco:
@jayarjo Are you happy for me to merge this PR? 
I will make the two changes I suggested once merged

jayarjo:
@cristianrgreco sure.

jayarjo:
Btw, running tests on every commit is a bit excessive I think, could run on push maybe?

PR #80, opened 2020-05-23
Adding rimraf instead of rm -rf

fokion:
The rm -rf  is a unix style command that windows users wont be able to use.
 If we switch to the rimraf the clean command will work.


cristianrgreco:
Thanks @fokion

Issue #100, opened 2020-07-17
Unable to resolve container name


rhyek:
I'm trying out the docker compose functionality and for this `docker-compose.yaml`:
```yaml
version: "3"
services:
  db:
    container_name: e2e_webapi_db
    image: mysql:8.0
    command: --default-authentication-plugin=mysql_native_password
    environment:
      - MYSQL_ROOT_PASSWORD=root
      - MYSQL_USER=test
      - MYSQL_PASSWORD=test
      - MYSQL_DATABASE=test
    volumes:
      - ./test_20200624_1428AST.sql.gz:/docker-entrypoint-initdb.d/dump.sql.gz
    ports:
      - 10000:3306
```
and this code:
```js
const path = require('path');
const {
  DockerComposeEnvironment,
  StartedDockerComposeEnvironment,
  Wait,
} = require('testcontainers');

(async () => {
  const composeFilePath = path.resolve(__dirname, '..');
  const composeFile = 'docker-compose.yaml';
  const environment = await new DockerComposeEnvironment(
    composeFilePath,
    composeFile
  )
    .up();
})();
```
I am getting this error:
```
(node:21220) UnhandledPromiseRejectionWarning: Error: Unable to resolve container name for: "/e2e_webapi_db"
    at Object.exports.resolveDockerComposeContainerName (/.../node_modules/testcontainers/dist/docker-compose-container-name-resolver.js:8:15)
    at DockerComposeEnvironment.<anonymous> (/.../node_modules/testcontainers/dist/docker-compose-environment.js:74:80)
    at Generator.next (<anonymous>)
    at fulfilled (/.../node_modules/testcontainers/dist/docker-compose-environment.js:24:58)
    at processTicksAndRejections (internal/process/task_queues.js:94:5)
```
This is using version `2.18.3`.

This also happens if I remove the `container_name` in the compose yaml.

cristianrgreco:
@rhyek I'll be able to take a proper look later, but I think a quick fix might be to change the `_` to something else in the container_name in the `docker-compose.yaml`

rhyek:
@cristianrgreco , thanks. as mentioned, even without `container_name` specified, i still get the same error (albeit, the error message mentions a different container name). one thing that looks odd is it always includes a leading slash. "/e2e_webapi_db"

cristianrgreco:
@rhyek Could you try 2.19.0? The release included a couple other things so hopefully I haven't created more work for myself :smile:

[...]

PR #101, opened 2020-07-17
Support docker-compose with container name

cristianrgreco:
Addresses #100


rhyek:
@cristianrgreco, that was quick. thanks! any idea when this might be released?

PR #102, opened 2020-07-18
Add Kafka module

cristianrgreco:
Addresses #17.

This will be a major release as it changes the signature of `GenericContainer.isCreating(BoundPorts): void` to `GenericContainer.preCreate(DockerClient, BoundPorts): Promise<void>`.

Please see the proposed README for more info: https://github.com/testcontainers/testcontainers-node/blob/add-kafka-module/src/modules/kafka/README.md


src/modules/kafka/kafka-container.ts
```diff
@@ -0,0 +1,77 @@
...
+    } else {
+      const zooKeeperHost = this.uuid.nextUuid();
+      const zooKeeperPort = await this.portClient.getPort();
+
+      this.withEnv("KAFKA_ZOOKEEPER_CONNECT", `${zooKeeperHost}:${zooKeeperPort}`);
+
+      const zookeeperContainer = await new GenericContainer("confluentinc/cp-zookeeper", "latest")
+        .withName(zooKeeperHost)
+        .withEnv("ZOOKEEPER_CLIENT_PORT", zooKeeperPort.toString())
+        .withExposedPorts(zooKeeperPort);
```
katerina-semikina, line 62: For which purpose exposing of port is done here?

cristianrgreco: Good point, this isn't needed!

PR #110, opened 2020-08-07
neo4j module was added

ltwlf:
I've added a module for the graph database neo4j


cristianrgreco:
Thanks for the PR @ltwlf 🙂

package.json
```diff
@@ -39,6 +39,7 @@
     "dockerode": "^3.2.0",
     "get-port": "^5.1.1",
     "glob": "^7.1.6",
+    "neo4j-driver": "^4.1.1",
```
cristianrgreco, line 42: dev dependency?

src/modules/neo4j/README.md
```diff
@@ -0,0 +1,50 @@
...
+Start a neo4j container and create a person node
+
+```typescript
+import neo4j from "neo4j-driver";
+import { Neo4jContainer } from "testcontainers";
+
+const container = await new Neo4jContainer().start();
+
+const driver = neo4j.driver(
+  "bolt://localhost:" + container.getMappedPort(Neo4jContainer.defaultBoltPort),
```
cristianrgreco, line 17: `localhost` -> `container.getContainerIpAddress()`

src/modules/neo4j/README.md
```diff
@@ -0,0 +1,50 @@
...
+  neo4j.auth.basic(Neo4jContainer.defaultUserName, Neo4jContainer.defaultPassword)
```
cristianrgreco, line 18: instead of exposing constants, it'd be nice if the container itself could expose these, eg:
`container.getUsername()` and `container.getPassword()`
that way the constants can be made private and users of the container don't need to know where to find these constants

src/modules/neo4j/README.md
```diff
@@ -0,0 +1,50 @@
...
+```
+
+Example installing APOC plugin and enabling ttl.
+
+```typescript
+import neo4j from "neo4j-driver";
+
+container = await new Neo4jContainer()
+  .withEnv("NEO4JLABS_PLUGINS","[\"apoc\"]")
+  .withEnv("NEO4J_apoc_ttl_enabled","true")
```
cristianrgreco, line 46: expose a builder method in the container which does this? `container.withApoc()`

src/modules/neo4j/README.md
cristianrgreco: how about a method in the container like `container.getUri()`?

src/modules/neo4j/neo4j-container.ts
```diff
@@ -0,0 +1,33 @@
...
+  public static readonly defaultUserName = "neo4j";
+  public static readonly defaultPassword = "test";
+
+  private readonly uuid: Uuid = new RandomUuid();
+  private password = Neo4jContainer.defaultPassword;
+
+  constructor(readonly image: Image = "neo4j", readonly tag: Tag = "latest", readonly host?: Host) {
+    super(image, tag);
+    this.host = host === undefined ? this.uuid.nextUuid() : host;
+    this.withName(this.host).withExposedPorts(Neo4jContainer.defaultBoltPort, Neo4jContainer.defaultHttPort);
```
cristianrgreco, line 20: setting `host` and `withName(this.host)` isn't required - it was required for kafka because it started two containers, kafka and the zookeeper, which had to be on the same network

src/modules/neo4j/README.md
ltwlf: There are plenty of config possibilities. This is just an example, but your are right it would make it easier for users. I will add for this case withApocPlugin. Good idea!

package.json
ltwlf: Indeed! Thanks for checking.

src/modules/neo4j/README.md
ltwlf: I had it exactly like you suggested before and then changed it to the current implementation. Didn't make me happy either. 
It thought it is maybe confusing with container and startedContainer. My preferred way would be to have it on the startedContainer like getMappedPort, but I guess this more complicated because the base class returns a StartedContainer instance.  But for know I revert it back to my initial approach like you suggested to.

src/modules/neo4j/neo4j-container.ts
ltwlf: Thanks for clarifying. I wasn't sure about this.

ltwlf:
@cristianrgreco with the last commit I'm much more happy. Looking forward for your feedback,

src/modules/neo4j/README.md
```diff
@@ -0,0 +1,48 @@
...
+Install APOC plugin:
+
+```typescript
+container = await new Neo4jContainer().withApoc().start();
+  ```
+
+Set custom password
+
+```typescript
+container = await new Neo4jContainer().withApoc().start();
```
cristianrgreco, line 47: missing `withPassword()`?

src/modules/neo4j/neo4j-container.ts
```diff
@@ -0,0 +1,68 @@
...
+import { StartedTestContainer } from "../../test-container";
+
+export interface StartedNeo4jTestContainer extends StartedTestContainer {
+  getPassword: () => string;
+  getUsername: () => string;
+  getBoltUri: () => string;
+}
+
+export class Neo4jContainer extends GenericContainer {
+  public static readonly defaultBoltPort = 7687;
```
cristianrgreco, line 15: any reason why not to make these private? my only thinking is that as soon as they're public, we won't be able to hide/change them - it seems they're all internal to the class

src/modules/neo4j/neo4j-container.ts
```diff
@@ -0,0 +1,68 @@
...
+    );
+  }
+
+  public async start(): Promise<StartedNeo4jTestContainer> {
+    this.startedContainer = await super.start();
+    (this.startedContainer as any).getPassword = this.getPassword.bind(this);
+    (this.startedContainer as any).getUsername = this.getUsername.bind(this);
+    (this.startedContainer as any).getBoltUri = this.getBoltUri.bind(this);
+
+    return this.startedContainer as StartedNeo4jTestContainer;
```
cristianrgreco, line 52: i think this should return a `new StartedNeo4jTestContainer(username, password, boltUri)`, so changing from an interface to a class, and the class itself will define the getters - that way we don't need to keep the instance as a field on this class (line 21)

src/modules/neo4j/neo4j-container.ts
```diff
@@ -0,0 +1,68 @@
+import { GenericContainer, Wait } from "../..";
+import { Image, Tag } from "../../repo-tag";
+import { Host } from "../../docker-client-factory";
+import { DockerClient } from "../../docker-client";
+import { BoundPorts } from "../../bound-ports";
+import { StartedTestContainer } from "../../test-container";
+
+export interface StartedNeo4jTestContainer extends StartedTestContainer {
```
cristianrgreco, line 8: we'll need to export this class from the index

src/modules/neo4j/README.md
ltwlf: withPassword was missing

src/modules/neo4j/neo4j-container.ts
ltwlf: Agree, it is better to expose less to avoid breaking changes

src/modules/neo4j/neo4j-container.ts
ltwlf: Like you wish :)

ltwlf:
@cristianrgreco next round...

src/modules/neo4j/neo4j-container.ts
```diff
@@ -0,0 +1,112 @@
...
+
+  public getId(): ContainerId {
+    return this.startedContainer.getId();
+  }
+
+  public exec(command: Command[]): Promise<ExecResult> {
+    return this.startedContainer.exec(command);
+  }
+
+  public logs(): Promise<NodeJS.ReadableStream> {
```
cristianrgreco, line 44: i'll take an action to create an abstract base class for StartedTestContainer which implements all these common methods (stop, getContainerIpAddress, getMappedPort, getName, getId, exec, logs, etc)

ltwlf: The abstract class would definitely make sense.

ltwlf:
Fine for me.

ltwlf:
Clicked accidentally on close.  Reopened and fine for me to release. I want to integrate this as a best practice for unit testing in the GRAND Stack https://grandstack.io/

cristianrgreco:
looks like a real nice site, and with a name like GRAND i gotta take a look!

cristianrgreco:
released in 3.3.0

ltwlf:
Great!

PR #127, opened 2020-09-17
Fixes arango test - Have fun!

ltwlf:
Uses log message wait strategy to ensure the db is ready


cristianrgreco:
Thanks @ltwlf ! ❤️

Issue #133, opened 2020-09-18
Docker Compose for Node not properly finishing the spinup


aixellent:
I tried to spin up my Testsetup in my globalSetup file (JEST) with the following docker-compose.yml:
```
version: '3.7'

services:
  rabbitmq:
    image: rabbitmq:3-management
    container_name: rabbitmq
    environment:
      RABBITMQ_DEFAULT_USER: test
      RABBITMQ_DEFAULT_PASS: test
    volumes:
      - './rabbitmq/etc/:/etc/rabbitmq/'
      - './rabbitmq/etc/definitions.json:/etc/rabbitmq/definitions.json:ro'
    ports:
      - 5672
      - 15672
      - 25672
  api:
    image: gcr.io/gke-utility/mocks
    container_name: api
    ports:
      - 5000
    command: api/v1.yaml
```

This is just to ensure my dependencies for the service i want to Test.

Global Setup file
```
import { Duration, TemporalUnit } from 'node-duration'
import { DockerComposeEnvironment, StartedTestContainer } from 'testcontainers'
import path = require('path')

export default async (): Promise<void> => {
  const composeFilePath: string = path.basename('../.docker')
  const composeFile = 'docker-compose.yml'
  const environment = await new DockerComposeEnvironment(composeFilePath, composeFile)
    .withStartupTimeout(new Duration(200, TemporalUnit.SECONDS))
    .up()

  const rabbitmq: StartedTestContainer = environment.getContainer('rabbitmq')
  const bully: StartedTestContainer = environment.getContainer('bully')

  // THE REST OF THE SETUP

}
```

The following problem occurs after the timeout:

`
$ NODE_ENV=test jest --config=__integration-tests__/jest.config.js --testRunner='jest-circus/runner' --ci --verbose --maxWorkers=7
Error: Port 5671 not bound after 200000ms
    at /PATH_TO_PROJECT/node_modules/testcontainers/dist/wait-strategy.js:61:23
    at IntervalRetryStrategy.<anonymous> /PATH_TO_PROJECT/node_modules/testcontainers/dist/retry-strategy.js:37:28)
    at Generator.next (<anonymous>)
    at fulfilled (/PATH_TO_PROJECT/testcontainers/dist/retry-strategy.js:5:58)
    at processTicksAndRejections (internal/process/task_queues.js:85:5)
error Command failed with exit code 1.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
`

[...]

PR #140, opened 2020-09-22
Add support for stopping a docker-compose environment

cristianrgreco:
Addresses #133 

Supports stopping a docker-compose environment. Calling `stop` returns a `StoppedDockerComposeEnvironment` which can also be `down`'d which returns a `DownedDockerComposeEnvironment`.


aixellent (approved):
Great that you could implement this so fats. Thanks 🎉

cristianrgreco:
Released in 4.1.0

PR #155, opened 2020-10-27
mention specific version of node-duration in readme

helgridly:
Hi there. I got briefly confused by conflicting versions of `node-duration` when following through the health-check example: I just went ahead and installed the latest `node-duration` but ran into the change in API. So I figured the readme could mention some use of this.

### DIFF
--- a/README.md
+++ b/README.md
@@ -186,7 +186,7 @@ const container = await new GenericContainer("alpine")
   .start();
 ```
 
-Creating a container with a custom health check command. 
+Creating a container with a custom health check command. You will need to `npm i -D node-duration@1.0.4` for this.
 
 Note that `interval`, `timeout`, `retries` and `startPeriod` are optional; the values will be inherited from the image or parent image if omitted. Also note that the wait strategy should be set to `Wait.forHealthCheck()` for this option to take effect:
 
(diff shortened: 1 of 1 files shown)


cristianrgreco:
Hi @helgridly, thanks for raising the PR. Just for my understanding is the problem that you have an explicit dependency on node-duration, perhaps a 2.x.x version, which isn't working well with testcontainer's reliance on 1.x.x?

helgridly:
No, I don't have a dependency on node-duration at all, outside the use of testcontainers. My `npm i -D testcontainers` did not provide node-duration, so I had to manually install it myself. So I went ahead and ran `npm i -D node-duration` and got the 2.x.x version.

cristianrgreco:
Hmm, I don't see how that can happen considering node-duration is a dependency: https://github.com/testcontainers/testcontainers-node/blob/master/package.json#L41

I just tried `npm install -D testcontainers` and I can see `node-duration` is there in the `node_modules`, do you think there could be something else going on?

helgridly:
On my install it's in `node_modules/testcontainers/node_modules`. It works fine if I `import { Duration, TemporalUnit } from "testcontainers/node_modules/node-duration";`, but that seems off to me. I am new to nodejs though so if this is normal, I'm happy to just do that.

cristianrgreco:
@helgridly Could you pls share a repro and let me know what version of node/testcontainers you're using?

helgridly:
package-lock says `"testcontainers": "^4.7.0"`. I'll try to whip something up for you now.

helgridly:
A reinstall of testcontainers fixed it for me. Weird. Thanks!

vedantroy:
@cristianrgreco I am also running into this issue. I did "yarn add --dev node-duration", and the version that was installed was version 2.0.0, which seems to be incompatible with the version of node duration being used in testcontainers. Is there a way for testcontainers to re-export the node-duration module?

Maybe this library needs to mark a peer dependency on node duration version < 2?

cristianrgreco:
Hi @vedantroy, I've decided to remove the dependency on node-duration: https://github.com/testcontainers/testcontainers-node/pull/157

cristianrgreco:
This has been released in v5.0.0

PR #180, opened 2021-01-25
Add ability to copy files to a container before it is started

dmitri-gb:
This adds two functions to `GenericContainer`:
```typescript
withCopyFileToContainer(sourcePath: string, containerPath: string);

withCopyContentToContainer(content: string | Buffer | Readable, containerPath: string);
```
Similar API also exists in the [Java library](https://javadoc.io/static/org.testcontainers/testcontainers/1.15.1/org/testcontainers/containers/GenericContainer.html#withCopyFileToContainer-org.testcontainers.utility.MountableFile-java.lang.String-).

### DIFF
--- a/src/container.ts
+++ b/src/container.ts
@@ -55,6 +55,7 @@ export interface Container {
   exec(options: ExecOptions): Promise<Exec>;
   logs(): Promise<Readable>;
   inspect(): Promise<InspectResult>;
+  putArchive(stream: Readable, containerPath: string): Promise<void>;
 }
 
 export class DockerodeContainer implements Container {
@@ -117,6 +118,15 @@ export class DockerodeContainer implements Container {
       healthCheckStatus: this.getHealthCheckStatus(inspectResult),
       networkSettings: this.getNetworkSettings(inspectResult),
     };
+  }
+
+  public async putArchive(stream: Readable, containerPath: string): Promise<void> {
+    const respStream = await this.container.putArchive(stream, { path: containerPath });
+    await new Promise((resolve, reject) =>
+      this.container.modem.followProgress(respStream, (err: unknown, res: unknown) =>
+        err ? reject(err) : resolve(res)
+      )
+    );
   }
 
   private getName(inspectInfo: ContainerInspectInfo): ContainerName {
--- a/src/container.ts
+++ b/src/container.ts
@@ -3,6 +3,7 @@ import { log } from "./logger";
 import { Command, ContainerName, ExitCode } from "./docker-client";
 import { Port } from "./port";
 import { Duplex, Readable } from "stream";
+import streamToArray from "stream-to-array";
 
 export type Id = string;
 
@@ -122,11 +123,7 @@ export class DockerodeContainer implements Container {
 
   public async putArchive(stream: Readable, containerPath: string): Promise<void> {
     const respStream = await this.container.putArchive(stream, { path: containerPath });
-    await new Promise((resolve, reject) =>
-      this.container.modem.followProgress(respStream, (err: unknown, res: unknown) =>
-        err ? reject(err) : resolve(res)
-      )
-    );
+    await streamToArray(respStream);
   }
 
   private getName(inspectInfo: ContainerInspectInfo): ContainerName {
--- a/src/generic-container.ts
+++ b/src/generic-container.ts
@@ -90,7 +90,7 @@ export class GenericContainer implements TestContainer {
   protected privilegedMode = false;
   protected daemonMode = false;
   protected pullPolicy: PullPolicy = new DefaultPullPolicy();
-  protected tarToCopy: archiver.Archiver | undefined;
+  protected tarToCopy?: archiver.Archiver;
 
   private extraHosts: ExtraHost[] = [];
 
--- a/src/generic-container.ts
+++ b/src/generic-container.ts
@@ -1,3 +1,4 @@
+import archiver from "archiver";
 import { BoundPorts } from "./bound-ports";
 import { Container, Id as ContainerId, InspectResult } from "./container";
 import { ContainerState } from "./container-state";
@@ -89,6 +90,7 @@ export class GenericContainer implements TestContainer {
   protected privilegedMode = false;
   protected daemonMode = false;
   protected pullPolicy: PullPolicy = new DefaultPullPolicy();
+  protected tarToCopy: archiver.Archiver | undefined;
 
   private extraHosts: ExtraHost[] = [];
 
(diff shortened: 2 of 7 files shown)


cristianrgreco:
Hi @dmitri-gb, thanks for raising!

Couple of questions:
1. What is the difference between `withBindMount` and `withCopyFileToContainer`?
2. We're using a library called `archiver` where we already have a `tar` library ([see here](https://github.com/testcontainers/testcontainers-node/blob/f58fe65acbbf1959c61416ad23a8727a7c8ca4e0/src/docker-client.ts#L232). Could we perhaps use the same library?

src/container.ts
```diff
@@ -119,6 +120,15 @@ export class DockerodeContainer implements Container {
     };
   }
 
+  public async putArchive(stream: Readable, containerPath: string): Promise<void> {
+    const respStream = await this.container.putArchive(stream, { path: containerPath });
+    await new Promise((resolve, reject) =>
+      this.container.modem.followProgress(respStream, (err: unknown, res: unknown) =>
```
cristianrgreco, line 126: Is it necessary to follow progress here?

dmitri-gb: I tought at first that it would be suitable to use `followProgress` here, but on second look, it seems I can also just do `await streamToArray(respStream)` instead.

dmitri-gb:
Hi @cristianrgreco!

> 1. What is the difference between the existing `withBindMount` and `withCopyFileToContainer`?

The big difference is that `withCopyFileToContainer` will actually _copy_ the file into the container, where as `withBindMount` gives the container access to an existing file or directory on the host. This also implies that
1) the container can modify the file without affecting the original file on the host;
2) since the file resides on the container's file system, read operations can be significantly faster in certain scenarios (e.g. when using Docker for Windows with WSL 2 based engine, which has [notoriously slow performance for mounted volumes](https://github.com/microsoft/WSL/issues/4197#issuecomment-604592340)).

> 2. We're using a library called `archiver` where we already have a `tar` library ([see here](https://github.com/testcontainers/testcontainers-node/blob/f58fe65acbbf1959c61416ad23a8727a7c8ca4e0/src/docker-client.ts#L232)). Could we use the same library?

As far as I know , the primary purpose of the `tar-fs` library is to pack a single directory on the file system. In this PR I need to add arbitrary files from different locations into a single archive, so `tar-fs` API is not very useful. Actually `tar-fs` and `archiver` both use the `tar-stream` library underneath, so the same code is used to create the actual stream in both cases 🙂 

> Please add an example for each operation to the README 🙂

Will do!

src/test-container.ts
```diff
@@ -29,6 +29,8 @@ export interface TestContainer {
   withDefaultLogDriver(): this;
   withPrivilegedMode(): this;
   withPullPolicy(pullPolicy: PullPolicy): this;
+  withCopyFileToContainer(sourcePath: string, containerPath: string): this;
+  withCopyContentToContainer(content: string | Buffer | Readable, containerPath: string): this;
```
cristianrgreco, line 33: What do you think about creating a `Content` type? `type Content = string | Buffer | Readable`

src/generic-container.ts
```diff
@@ -239,11 +246,28 @@ export class GenericContainer implements TestContainer {
...
+    this.getTarToCopy().append(content, { name: containerPath });
+    return this;
+  }
+
   public async isImageCached(dockerClient: DockerClient): Promise<boolean> {
     const dockerImageNames = await dockerClient.fetchDockerImageNames();
     return dockerImageNames.some((dockerImageName) => dockerImageName.equals(this.dockerImageName));
   }
 
+  protected getTarToCopy(): archiver.Archiver {
```
cristianrgreco, line 264: Any reason to make this protected or could it be private?

src/generic-container.ts
```diff
@@ -145,6 +147,11 @@ export class GenericContainer implements TestContainer {
       }
     }
 
+    if (this.tarToCopy) {
+      this.tarToCopy.finalize();
+      await container.putArchive(this.tarToCopy, "/");
```
cristianrgreco, line 152: Would a user ever want to specify the target path?

src/generic-container.ts
```diff
@@ -89,6 +90,7 @@ export class GenericContainer implements TestContainer {
   protected privilegedMode = false;
   protected daemonMode = false;
   protected pullPolicy: PullPolicy = new DefaultPullPolicy();
+  protected tarToCopy: archiver.Archiver | undefined;
```
cristianrgreco, line 93: Could do `protected tarToCopy?: archiver.Archiver`

cristianrgreco:
I left a couple of nitpicks and a question, other than that LGTM! 
Thanks again and thanks for explaining how those bits worked 🙂

src/test-container.ts
dmitri-gb: Hmm, I'm concerned that this will make the API less readable. For instance here's how the IDE behaves currently:

image

Where as with the type alias it will look like this:

image

The second screenshot doesn't really tell what can I actually pass as the argument, and I have to go and look up the definition of `Content`. In the first screenshot it is immediately clear that I am free to pass strings, buffers and readable streams. Does that make sense?

src/generic-container.ts
dmitri-gb: Well, the `tarToCopy` field itself is protected (in line with all the other fields), and if some subclass intends to use it, then I think they should also have access to this getter which also acts as the lazy initializer for the field.

src/test-container.ts
cristianrgreco: You're right, definitely better without the type, I hadn't considered the lack of IntelliSense, thanks!

src/generic-container.ts
dmitri-gb: It is the path at which the TAR archive gets extracted. This means it is also the path against which user provided container paths get resolved. For example if a user invokes `withCopyFileToContainer('file.txt', 'dir/bar.txt')`, then the file will end up in `/dir/bar.txt`. Typically though, it is expected that the user already provides an absolute path (`withCopyFileToContainer('file.txt', '/dir/bar.txt')`), in which case this parameter should have no effect.

src/generic-container.ts
dmitri-gb: Sure.

cristianrgreco (approved):

cristianrgreco:
@dmitri-gb Are you happy for me to merge/release?

dmitri-gb:
@cristianrgreco Absolutely! 👍

cristianrgreco:
Released in https://github.com/testcontainers/testcontainers-node/releases/tag/v7.1.0, thanks @dmitri-gb !

PR #183, opened 2021-01-26
Propagate a clear error when a container with LogWaitStrategy fails

dmitri-gb:
Right now if a container that uses `Wait.forLogMessage` fails ot start, then `undefined` is thrown and the only thing that gets printed by e.g. Jest is
```
thrown: undefined
```


cristianrgreco (approved):

cristianrgreco:
Thanks @dmitri-gb, you're on a roll!

cristianrgreco:
Released in v7.2.0 🙂

PR #191, opened 2021-03-08
Port allocation doesn't work well, when containers are created by jest

xxBedy:
Port allocation doesn't work well, when containers are created by jest in parallel.

Due to limitations in 'get-port' library, port conflict occurred too often to be ignored (Documented in get-port: "There is a very tiny chance of a race condition if another process starts using the same port number as you in between the time you get the port number and you actually start using it.") .


cristianrgreco:
Thanks @xxBedy !

PR #192, opened 2021-03-08
Port host override feature from testcontainers-java

alexeagle:
This is used when a (possibly remote) docker daemon brings up the container at a network address other than localhost.

In testcontainers-java the logic appears here https://github.com/testcontainers/testcontainers-java/blob/c3a8ca76d0ed69f71c94e5bc2f7f1e0eef1a2259/core/src/main/java/org/testcontainers/dockerclient/DockerClientProviderStrategy.java#L264
similarly to this testcontainers-node change, this is checked immediately before looking into the "bridge" network to find a gateway address to the container.

We use this environment variable for our Java tests and want to have the Node tests behave the same way. This is because we run our CI on Buildkite, where agents running the tests are already containerized, so they reach out to a docker daemon running outside this container on the physical machine. Therefore the testcontainers are not on the local network address.

### DIFF
--- a/README.md
+++ b/README.md
@@ -23,6 +23,7 @@ The following environment variables are supported:
 | `DEBUG` | `testcontainers*` | See all output |
 | `DOCKER_HOST` | `tcp://docker:2375` | Override the Docker host |
 | `TESTCONTAINERS_RYUK_DISABLED` | `true` | Disable [ryuk](#ryuk) |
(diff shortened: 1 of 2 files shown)


cristianrgreco:
Thanks @alexeagle! 

Could you also update the Configuration section of the README to show this environment var is supported?

alexeagle:
Done, thanks!

cristianrgreco:
Thanks @alexeagle, released in 7.5.0

PR #201, opened 2021-04-08
Improve docker config auth loader registry matching

crickford:
Allows auth config to be loaded from the docker config file when the entry for a given registry is keyed under the registry's URL.

Implementation based on corresponding logic in `testcontainers-java`: https://github.com/testcontainers/testcontainers-java/blob/b3cf0624c8f251af0e5361d3a92d155d3563b372/core/src/main/java/org/testcontainers/utility/RegistryAuthLocator.java#L208

### DIFF
--- a/src/registry-auth-locator/auths.ts
+++ b/src/registry-auth-locator/auths.ts
@@ -1,4 +1,4 @@
-import { DockerConfig } from "./types";
+import { Auth, DockerConfig } from "./types";
 import { AuthConfig } from "../docker-client";
 import { RegistryAuthLocator } from "./registry-auth-locator";
 
@@ -7,15 +7,28 @@ export class Auths implements RegistryAuthLocator {
     return "Auths";
   }
 
+  findAuthEntry(registry: string, dockerConfig: DockerConfig): Auth | null {
+    const authEntries = dockerConfig.auths ?? {};
+    for (const key in authEntries) {
+      if (key.includes(`://${registry}`) || key === registry) {
+        return authEntries[key];
+      }
+    }
+    return null;
+  }
+
   isApplicable(registry: string, dockerConfig: DockerConfig): boolean {
-    return dockerConfig.auths !== undefined && dockerConfig.auths[registry] !== undefined;
+    return this.findAuthEntry(registry, dockerConfig) !== null;
   }
 
   async getAuthConfig(registry: string, dockerConfig: DockerConfig): Promise<AuthConfig | undefined> {
     const authConfig: Partial<AuthConfig> = { registryAddress: registry };
 
-    // @ts-ignore
-    const auth: Auth = dockerConfig.auths[registry];
+    const auth = this.findAuthEntry(registry, dockerConfig);
+
+    if (auth === null) {
+      return undefined;
+    }
 
     if (auth.email) {
       authConfig.email = auth.email;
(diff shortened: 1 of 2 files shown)


src/registry-auth-locator/auths.ts
```diff
@@ -7,15 +7,28 @@ export class Auths implements RegistryAuthLocator {
...
   }
 
   async getAuthConfig(registry: string, dockerConfig: DockerConfig): Promise<AuthConfig | undefined> {
     const authConfig: Partial<AuthConfig> = { registryAddress: registry };
 
-    // @ts-ignore
-    const auth: Auth = dockerConfig.auths[registry];
+    const auth = this.findAuthEntry(registry, dockerConfig);
+
+    if (auth === null) {
```
cristianrgreco, line 29: this check shouldn't be needed as it's the same as what's in the `isApplicable` method, so if it's `null` there then `getAuthConfig` shouldn't be called

cristianrgreco:
Thanks again for raising @crickford 🙂 Just had one question but other than that LGTM

src/registry-auth-locator/auths.ts
crickford: Gotcha, that makes sense. I see it as more of a convenience really, since it acts as a type guard in this situation.

Is what you're describing the defined behavior of any class implementing the RegistryAuthLocator interface? If that is the case, perhaps the signature of `getAuthConfig` could omit `undefined` in the return type [here](https://github.com/testcontainers/testcontainers-node/blob/2ec9b50aeecf700017eebedb03fe9127e950076d/src/registry-auth-locator/registry-auth-locator.ts#L7) - what do you think?

cristianrgreco: > Is what you're describing the defined behavior of any class implementing the RegistryAuthLocator interface?

Yep, if the auth locator returns false for `isApplicable` then the `getAuthConfig` of that locator shouldn't be called. 

Yes ideally we could drop `undefined` from the type, I cannot remember exactly why it was needed but maybe you will have better luck 🙂 If the type can't be changed _I think_ you could still drop the `if auth === null check`, but would be good to double check/test

crickford: Ah, I see. I noticed that the `testcontainers-java` works a little differently; it calls each auth locator just once in order of preference, and each auth locator attempts to returns credentials - bailing out early and returning null if it is inapplicable or if no creds are found. I think this pattern lends itself nicely to static typing, since there is no elegant way to express "this method will never be called unless this other method has previously returned true" which I am aware of.

Would you be open to (could be in this or a separate PR if you preferred) a refactor of the registry auth locator code to follow the single method pattern in `testcontainers-java`?

cristianrgreco: I think that'd be fine, thanks @crickford

crickford: Hey @cristianrgreco - the null check is needed for now, either that or a type assertion something like this:
`const auth = this.findAuthEntry(registry, dockerConfig) as Auth;` I'm still planning to work on a refactor which would eliminate the need for either, but I won't have the chance to work on it for a little bit so I'll open a separate PR for that if you want to go ahead and merge this one :) Thanks!

cristianrgreco: Ah my bad, I misunderstood 🙂 I will merge/release this now then, thanks again @crickford

cristianrgreco: Released in 7.8.0

PR #260, opened 2021-08-15
Add TS type for bind mounts in SELinux-enabled systems

aldofunes:
Add the 'z' bind mount option to the TypeScript type alias for SELinux-enabled systems.

Here is the documentation for the label: https://docs.docker.com/storage/bind-mounts/#configure-the-selinux-label

My current workaround is to use type casting:

```typescript
  const genericContainer = new GenericContainer('fedora')
    .withBindMount(`/path/to/mount`, '/mount', 'z' as 'rw')
    .withCmd('ls /mount');
```

### DIFF
--- a/src/docker/types.ts
+++ b/src/docker/types.ts
@@ -10,7 +10,7 @@ export type Env = { [key in EnvKey]: EnvValue };
 
 export type Dir = string;
 
-export type BindMode = "rw" | "ro";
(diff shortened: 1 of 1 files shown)


cristianrgreco:
Thanks @aldofunes, released in [7.19.1](https://github.com/testcontainers/testcontainers-node/releases/tag/v7.19.1) 🙂

PR #266, opened 2021-09-06
add volume feature

thalleslmF:
Signed-off-by: thalles <lopthalles@gmail.com>

### DIFF
--- a/README.md
+++ b/README.md
@@ -332,6 +332,15 @@ const container = await new GenericContainer("alpine")
   .start();
 ```
 
+Creating a container with volumes associated:
+
+```javascript
+const { GenericContainer } = require("testcontainers");
+
+const container = await new GenericContainer("alpine")
+  .withVolumes("volumes/new-volume")
+  .start();
+```
 Creating a container with [IPC mode](https://docs.docker.com/engine/reference/run/#ipc-settings---ipc):
 
 ```javascript
--- a/src/docker/functions/container/create-container.ts
+++ b/src/docker/functions/container/create-container.ts
@@ -23,12 +23,12 @@ export type CreateContainerOptions = {
   extraHosts: ExtraHost[];
   ipcMode?: string;
   user?: string;
+  volumes?: string | string[];
 };
 
 export const createContainer = async (options: CreateContainerOptions): Promise<Dockerode.Container> => {
   try {
     log.info(`Creating container for image: ${options.imageName}`);
-
     return await dockerode.createContainer({
       name: options.name,
       User: options.user,
@@ -50,6 +50,7 @@ export const createContainer = async (options: CreateContainerOptions): Promise<
         LogConfig: getLogConfig(options.useDefaultLogDriver),
         Privileged: options.privilegedMode,
       },
+      Volumes: getVolumes(options.volumes),
     });
   } catch (err) {
     log.error(`Failed to create container for image ${options.imageName}: ${err}`);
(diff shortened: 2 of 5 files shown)


src/docker/functions/container/create-container.ts
```diff
@@ -59,6 +60,19 @@ export const createContainer = async (options: CreateContainerOptions): Promise<
 
 type DockerodeEnvironment = string[];
 
+type Volumes = { [volumes in string]: Record<string, unknown> };
+const getVolumes = (volumeNames: string | string[] | undefined) => {
```
cristianrgreco, line 64: I think using varargs would simplify this type and its validations (`volumes: ...string`)

src/generic-container/generic-container.ts
```diff
@@ -185,6 +186,11 @@ export class GenericContainer implements TestContainer {
     return this;
   }
 
+  public withVolumes(volumes: string | string[]): this {
```
cristianrgreco, line 189: Need to add this method to the base interface `TestContainer` (https://github.com/testcontainers/testcontainers-node/blob/91d9127f3e157a59f7371bec59effc5889096b1d/src/test-container.ts#L19)

cristianrgreco (commented):
What's the use case that https://github.com/testcontainers/testcontainers-node/blob/55c560e7b4c9477666736f19184f485173b905e2/src/generic-container/generic-container.ts#L188 does not address?

Also please add example usage to the README

src/generic-container/generic-container.ts
thalleslmF: done

src/docker/functions/container/create-container.ts
thalleslmF: agreed! done

thalleslmF:
@cristianrgreco i think that a use case can be a user wants to create a volume without binding a source. Example in java:
(The bind method is independent of the method of volume creation)
```
final Volume varVolume = new Volume("/var/lib/containerd");
          cmd.withEntrypoint("/usr/local/bin/entrypoint", "/sbin/init")
                  .withName(CONTAINER_NAME)
                  .withHostName("kindcontainer-control-plane")
                  .withVolumes(varVolume)
                  .withBinds(new Bind("kindcontainer-volume", varVolume, true)); 
```

Also volumes behave differently than bind mounts: 
`Docker volumes are completely handled by Docker itself and therefore independent of both your directory structure and the OS of the host machine. `
(https://blog.logrocket.com/docker-volumes-vs-bind-mounts/)

src/docker/functions/container/create-container.ts
```diff
@@ -59,6 +60,12 @@ export const createContainer = async (options: CreateContainerOptions): Promise<
 
 type DockerodeEnvironment = string[];
 
+type Volumes = { [volumes in string]: Record<string, unknown> };
+const getVolumes = (...volumeNames: string[]) => {
```
cristianrgreco, line 64: this can just be a `string[]` without varargs to remove the need to spread on line 53

README.md
```diff
@@ -332,6 +332,15 @@ const container = await new GenericContainer("alpine")
   .start();
 ```
 
+Creating a container with volumes associated:
+
+```javascript
+
+  .withVolumes("volumes/new-volume")
```
cristianrgreco, line 341: could the example show that you can provide multiple volumes, similar to the `withExposedPorts` example? else people will ask how to specify multiple volumes

cristianrgreco:
> @cristianrgreco i think that a use case can be a user that wants to create a volume without binding a source. Example in java:
> (The bind method is independent of the method of volume creation)
> 
> ```
> final Volume varVolume = new Volume("/var/lib/containerd");
>           cmd.withEntrypoint("/usr/local/bin/entrypoint", "/sbin/init")
>                   .withName(CONTAINER_NAME)
>                   .withHostName("kindcontainer-control-plane")
>                   .withVolumes(varVolume)
>                   .withBinds(new Bind("kindcontainer-volume", varVolume, true)); 
> ```
> 
> Also volumes behave differently than bind mounts:
> `Docker volumes are completely handled by Docker itself and therefore independent of both your directory structure and the OS of the host machine. `
> (https://blog.logrocket.com/docker-volumes-vs-bind-mounts/)

gotcha, thanks for that. One thing I'm not sure about is that in the Java example, the `Volume` is its own type. testcontainers-node currently doesn't support creating volumes but I wonder if it should as part of this, that way `withVolumes(volumes: string[])` could become `withVolumes(volumes: Volume[])`. thoughts?

PR #285, opened 2021-10-31
introduce withPortMappings

slaviqueue:
Introducing `withPortMappings` to let users explicitly specify target host port.

### DIFF
--- a/README.md
+++ b/README.md
@@ -115,7 +115,7 @@ Creating a container explicit port mapping:
 ```javascript
 const { GenericContainer } = require("testcontainers");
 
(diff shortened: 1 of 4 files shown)


cristianrgreco:
https://bsideup.github.io/posts/testcontainers_fixed_ports/

slaviqueue:
@cristianrgreco but they have something similar in [java version of test containers](https://github.com/testcontainers/testcontainers-java/blob/master/core/src/main/java/org/testcontainers/containers/FixedHostPortGenericContainer.java). And also they say:

> Callers are responsible for ensuring that this fixed port is actually available

Why can't we have such an API for cases where it is required? If I create a fixed mapping, most probably I know what I'm doing.

PR #316, opened 2022-01-06
Add withEnvFile to support the --env-file option of docker-compose

### LINKED ISSUE #184: Support for custom env-file in docker-compose
Hi there, and thanks for your great work.

I am trying to use a different `.env` file configuration for my test environment, e.g. like in [the official docs for compose cli](https://docs.docker.com/compose/environment-variables/#substitute-environment-variables-in-compose-files).

Specifically, I'd like to

```bash
 docker-compose --env-file .env.test -f docker-compose.yml -f docker-compose.test.yml up
```

I want to stress on the option `--env-file .env.test`, which is the missing bit here.

Looking at [the implementation](https://github.com/testcontainers/testcontainers-node/blob/873ace5cc730d608ddb69bdbdde5b99f8997c9f1/src/docker-compose-environment.ts#L173), it does not seem possible to pass it to the env file correctly.

I use a lot variable interpolation in my compose files, which helps me to have similar but separated configurations.

## Proposal

add a `envFile?: string` option to the `DockerComposeEnvironment` class and pass the option down to the compose engine.

christopherjwang:
Hey!

I'd like to add support for the `--env-file` docker-compose option. This is useful in environments where a developer has `.env` file that is used for local development (this .env file gets picked up automatically from docker-compose) and the CI/test environment wants to use different values (i.e. ports or environment vars).

fixes #184 

[0] https://github.com/testcontainers/testcontainers-node/issues/184

### DIFF
--- a/README.md
+++ b/README.md
@@ -547,6 +547,16 @@ const environment = await new DockerComposeEnvironment(composeFilePath, composeF
   .up();
 ```
 
+Specify the [environment file](https://docs.docker.com/compose/environment-variables/#using-the---env-file--option):
+
+```javascript
+const { DockerComposeEnvironment } = require("testcontainers");
+
+const environment = await new DockerComposeEnvironment(composeFilePath, composeFile)
+    .withEnvFile(".env.custom")
+    .up();
+```
+
 Specify [profiles](https://docs.docker.com/compose/profiles/):
 
 ```javascript
--- a/fixtures/docker-compose/docker-compose-with-env-file/.env
+++ b/fixtures/docker-compose/docker-compose-with-env-file/.env
@@ -1,0 +1,1 @@
+ENV_VAR=default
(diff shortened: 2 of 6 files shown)


cristianrgreco:
Hi @christopherjwang, looks great, thanks for the contribution! Could you please add an example in the README?

christopherjwang:
@cristianrgreco I've updated the PR and added an example. It's now `withEnvFile` (without the s). I am wracking my brain trying to figure out why I thought there could be multiple --env-files (my initial implementation was faulty and incorrect).

cristianrgreco (approved):

christopherjwang:
Thanks for the quick turnaround! What's the protocol and timeline for cutting a new version/release?

cristianrgreco:
@christopherjwang released in [v8.2.0](https://github.com/testcontainers/testcontainers-node/releases/tag/v8.2.0) 🙂

PR #325, opened 2022-02-17
Feat/323/add env var startup timeout

### LINKED ISSUE #323: Feature: add environment variable for scaling "withStartupTimeout"
**Issue :** Test will fail if the container has not started in allocated time. Depending on environment the tests will "randomly" fail.

**Why :** Some environments might have limited performances (low spec machines, slow CI runner, etc) or limited connectivity (inital pull of images, new image, etc). Some might even have sufficient performances and only encounter issues when multitasking with other applications using too much ressources (eg slack, teams, vscode, web browsers...)

**Current solution :** Current way to handle this is to include a [.withStartupTimeout(ms)](https://www.testcontainers.org/features/startup_and_waits/) option on container startup to replace the default 60s of waiting time. However this value could have to change over time, and it's not optimal to include a big value because your CI runner is slow. It seems like it would be easier if the library could handle it OOB.

**Proposition :** Include an environment variable to replace the current default (60s) timeout

eg. `TESTCONTAINERS_DEFAULT_STARTUP_TIMEOUT` = 120

It could also be in the form of a scaling value 

eg. `TESTCONTAINERS_STARTUP_TIMEOUT_SCALING` = 2

Would a PR on this be accepted ?

mderrien29:
I added the timeout in WaitStrategy and generic-container, so I needed a way to keep env pulling in one place, took the liberty to move others `process.env` configs to the same spot.

closes #323

### DIFF
--- a/src/docker/docker-client.ts
+++ b/src/docker/docker-client.ts
@@ -7,6 +7,7 @@ import fs from "fs";
 import { runInContainer } from "./functions/run-in-container";
 import { logSystemDiagnostics } from "../log-system-diagnostics";
 import "../testcontainers-properties-file";
+import { EnvConfig } from "../config";
 
 type DockerClient = {
   host: Host;
@@ -150,8 +151,8 @@ class NpipeSocketStrategy implements DockerClientStrategy {
 }
 
 const resolveHost = async (dockerode: Dockerode, uri: string): Promise<string> => {
-  if (process.env.TESTCONTAINERS_HOST_OVERRIDE !== undefined) {
-    return process.env.TESTCONTAINERS_HOST_OVERRIDE;
+  if (EnvConfig.isHostOverriden()) {
+    return EnvConfig.isHostOverriden();
   }
 
   const { protocol, hostname } = new URL(uri);
--- a/src/docker/docker-client.ts
+++ b/src/docker/docker-client.ts
@@ -151,8 +151,8 @@ class NpipeSocketStrategy implements DockerClientStrategy {
 }
 
 const resolveHost = async (dockerode: Dockerode, uri: string): Promise<string> => {
-  if (EnvConfig.isHostOverriden()) {
-    return EnvConfig.isHostOverriden();
+  if (EnvConfig.getHostOverride()) {
+    return EnvConfig.getHostOverride();
   }
 
   const { protocol, hostname } = new URL(uri);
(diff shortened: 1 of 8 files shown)


src/images.ts
```diff
@@ -1,2 +1,4 @@
-export const REAPER_IMAGE = process.env["RYUK_CONTAINER_IMAGE"] ?? "testcontainers/ryuk:0.3.2";
-export const SSHD_IMAGE = process.env["SSHD_CONTAINER_IMAGE"] ?? "testcontainers/sshd:1.0.0";
+import { EnvConfig } from "./config";
+
+export const REAPER_IMAGE = EnvConfig.riukContainerImage();
```
cristianrgreco, line 3: typo `riuk` -> `ryuk`

src/docker/docker-client.ts
```diff
@@ -150,8 +151,8 @@ class NpipeSocketStrategy implements DockerClientStrategy {
 }
 
 const resolveHost = async (dockerode: Dockerode, uri: string): Promise<string> => {
-  if (process.env.TESTCONTAINERS_HOST_OVERRIDE !== undefined) {
-    return process.env.TESTCONTAINERS_HOST_OVERRIDE;
+  if (EnvConfig.isHostOverriden()) {
```
cristianrgreco, line 154: perhaps 2 separate methods, `isHostOverridden` and `getHostOverride` ?

src/generic-container/generic-container.ts
```diff
@@ -58,7 +59,7 @@ export class GenericContainer implements TestContainer {
   protected tmpFs: TmpFs = {};
   protected healthCheck?: HealthCheck;
   protected waitStrategy?: WaitStrategy;
-  protected startupTimeout = 60_000;
+  protected startupTimeout = EnvConfig.defaultSetupTimeout();
```
cristianrgreco, line 62: this to me reads as startupTimeout = the default, perhaps it should be `getStartupTimeout` ?

src/reaper.ts
```diff
@@ -66,11 +67,11 @@ export class ReaperInstance {
   }
 
   private static isEnabled(): boolean {
-    return process.env.TESTCONTAINERS_RYUK_DISABLED !== "true";
+    return EnvConfig.isRyukDisabled();
```
cristianrgreco, line 70: should be flipped

src/reaper.ts
```diff
@@ -79,7 +80,7 @@ export class ReaperInstance {
   }
 
   private static async createRealInstance(): Promise<Reaper> {
-    const dockerSocket = process.env["TESTCONTAINERS_DOCKER_SOCKET_OVERRIDE"] ?? "/var/run/docker.sock";
+    const dockerSocket = EnvConfig.dockerSocketOverride();
```
cristianrgreco, line 83: same here, i think it should just be `getDockerSocket`

cristianrgreco:
thank you @mderrien29, this is great, just some comments/suggestions

mderrien29:
Thank you for your review, i am a bit tired as you could read. PR updated.

mderrien29:
@cristianrgreco closed by mistake. sorry.

PR #326, opened 2022-02-26
Add env DOCKER_CONFIG

yubinTW:
Hey, I'd like to add support for the environment variable `DOCKER_CONFIG`.

Since I had a java project and a node project, both of them used testcontainers in test cases.
I found some different behavior, when using the docker config file, testcontainers-java will find `$DOCKER_CONFIG` folder first, and if `$DOCKER_CONFIG` is not defined, use `$HOME/.docker` as well, but testcontainers-node only use `$HOME/.docker`

Please check [RegistryAuthLocator.java#L64](https://github.com/testcontainers/testcontainers-java/blob/master/core/src/main/java/org/testcontainers/utility/RegistryAuthLocator.java#L64)
```java
final String dockerConfigLocation = System.getenv().getOrDefault("DOCKER_CONFIG",
    System.getProperty("user.home") + "/.docker");
```

In my case, environment have the variable `DOCKER_CONFIG`, I hope testcontainers-node can have the same behavior with testcontainers-java.

Please review my commit : )

### DIFF
--- a/src/registry-auth-locator/index.ts
+++ b/src/registry-auth-locator/index.ts
@@ -11,7 +11,9 @@ import { AuthConfig } from "../docker/types";
 
 const DEFAULT_REGISTRY = "https://index.docker.io/v1/";
 
-const dockerConfigFile = path.resolve(os.homedir(), ".docker", "config.json");
+const dockerConfigLocation = process.env.DOCKER_CONFIG || `${os.homedir()}/.docker`;
+
+const dockerConfigFile = path.resolve(dockerConfigLocation, "config.json");
 
 const readDockerConfig = async (): Promise<DockerConfig> => {
   if (!existsSync(dockerConfigFile)) {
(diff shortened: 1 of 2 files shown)


src/registry-auth-locator/index.ts
```diff
@@ -11,7 +11,9 @@ import { AuthConfig } from "../docker/types";
 
 const DEFAULT_REGISTRY = "https://index.docker.io/v1/";
 
-const dockerConfigFile = path.resolve(os.homedir(), ".docker", "config.json");
+const dockerConfigLocation = process.env.DOCKER_CONFIG || `${os.homedir()}/.docker`;
```
cristianrgreco, line 14: i don't think the `/` will work on Windows

cristianrgreco (commented):
thanks for raising, could you please also update the `configuration` section of the readme?

cristianrgreco:
@yubinTW Released in 8.3.0, thanks!

PR #328, opened 2022-03-01
Allow specifying host port to bind to

### LINKED ISSUE #314: Features to specify the host port in port mapping
I use testcontainers for running my integration tests. On my work computer, a lot of port ranges are blocked. That causes the `docker-modem` library to fail with the following error on around every third run of tests:

```
/home/abdullah/abb/WB-IdentityServer/node_modules/docker-modem/lib/modem.js:354
      callback(null, json);
      ^
Error: (HTTP code 500) server error - Ports are not available: listen tcp 0.0.0.0:58773: bind: An attempt was made to access a socket in a way forbidden by its access permissions.
 
    at /home/abdullah/abb/WB-IdentityServer/node_modules/docker-modem/lib/modem.js:324:17
    at getCause (/home/abdullah/abb/WB-IdentityServer/node_modules/docker-modem/lib/modem.js:354:7)
    at Modem.buildPayload (/home/abdullah/abb/WB-IdentityServer/node_modules/docker-modem/lib/modem.js:323:5)
    at IncomingMessage.<anonymous> (/home/abdullah/abb/WB-IdentityServer/node_modules/docker-modem/lib/modem.js:295:14)
    at IncomingMessage.emit (node:events:402:35)
    at IncomingMessage.emit (node:domain:475:12)
    at endReadableNT (node:internal/streams/readable:1343:12)
    at processTicksAndRejections (node:internal/process/task_queues:83:21)
```

A possible solution is to add to able to specify a host port as well for each port mapping e.g. in the `GenericContainer.withExposedPorts` method.

codetheweb:
Closes #314.

Open to renaming `PortWithOptionalBinding` if there's something else that makes more sense. :)

### DIFF
--- a/README.md
+++ b/README.md
@@ -301,10 +301,10 @@ expect((await container.exec(["nslookup", "foo"])).exitCode).toBe(0);
 expect((await container.exec(["nslookup", "bar"])).exitCode).toBe(0);
 ```
 
-Specifying a pull policy. Note that if omitted will use the `DefaultPullPolicy` which will use a locally cached image 
-if one already exists, this is usually the preferred option. In cases where there is a local image for a given tag 
-but the remote image with the same tag may have changed (for example when using the `latest` tag), you can tell 
-testcontainers to pull the image again by specifying an `AlwaysPullPolicy`: 
+Specifying a pull policy. Note that if omitted will use the `DefaultPullPolicy` which will use a locally cached image
+if one already exists, this is usually the preferred option. In cases where there is a local image for a given tag
+but the remote image with the same tag may have changed (for example when using the `latest` tag), you can tell
+testcontainers to pull the image again by specifying an `AlwaysPullPolicy`:
 
 ```javascript
 const { GenericContainer, AlwaysPullPolicy } = require("testcontainers");
@@ -326,7 +326,7 @@ const container = await GenericContainer.fromDockerfile(buildContext)
 
 Specifying a default log driver. You can override the logging driver used by Docker to be the default one (json-file).
 This might be necessary when the driver of your docker host does not support reading logs
-and you want to use the `Wait.forLogMessage` wait strategy. This is the same as 
+and you want to use the `Wait.forLogMessage` wait strategy. This is the same as
 [--log-driver json-file on docker run](https://docs.docker.com/config/containers/logging/configure/#configure-the-logging-driver-for-a-container).
 
 ```javascript
@@ -376,8 +376,8 @@ const container = await new GenericContainer("postgres")
   .withExposedPorts(5432)
   .start();
 
-await container.stop({ 
-  timeout: 10000 
+await container.stop({
+  timeout: 10000
 });
 ```
 
@@ -391,7 +391,7 @@ const container = await new GenericContainer("postgres")
   .withExposedPorts(5432)
   .start();
 
-await container.stop({ 
+await container.stop({
   removeVolumes: false
 });
 ```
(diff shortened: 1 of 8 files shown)


src/port.ts
```diff
@@ -1,2 +1,12 @@
 export type Port = number;
 export type PortString = string;
+
+export type PortWithOptionalBinding =
+  | Port
```
cristianrgreco, line 5: What's the trailing `|` do? Also below `number` -> `Port`

cristianrgreco:
Thanks for raising @codetheweb. Could you please add an example to the README on how to use it, preferably with some disclaimer that it should be avoided where possible?

src/port.ts
codetheweb: Partly fixed in [8777232](https://github.com/testcontainers/testcontainers-node/pull/328/commits/8777232642285c9060c2a51f8a5604b0fba98b39). Prettier is adding the leading `|`.

codetheweb:
Added an example to the readme. Thanks for the quick review. 😃

README.md
```diff
@@ -433,6 +433,20 @@ const { output } = await container.exec(["curl", `http://host.testcontainers.int
 assert(output === "hello world");
 ```
 
+Specifying an exact host port to bind to (not recommended, Testcontainers will automatically find an available port):
+
+```javascript
+const { GenericContainer, Network } = require("testcontainers");
```
cristianrgreco, line 439: could you please remove the network part from this example?

src/docker/functions/container/create-container.ts
```diff
@@ -79,10 +79,14 @@ const getExtraHosts = (extraHosts: ExtraHost[]): string[] => {
   return extraHosts.map((extraHost) => `${extraHost.host}:${extraHost.ipAddress}`);
 };
 
-const getPortBindings = (exposedPorts: Port[]): DockerodePortBindings => {
+const getPortBindings = (exposedPorts: PortWithOptionalBinding[]): DockerodePortBindings => {
   const dockerodePortBindings: DockerodePortBindings = {};
   for (const exposedPort of exposedPorts) {
-    dockerodePortBindings[exposedPort] = [{ HostPort: "0" }];
+    if (typeof exposedPort === "number") {
```
cristianrgreco, line 85: in the same way that we have `getContainerPort` which does the type checking, perhaps we could also add `getHostPort`? that way we fully encapsulate the type differences

codetheweb: I think that makes sense. Should it just return null if `typeof port === 'number'`?

README.md
codetheweb: My bad, fixed in 6360c4b.

src/docker/functions/container/create-container.ts
codetheweb: Ah, I see. Added in a27f11b.

src/generic-container/generic-container.ts
```diff
@@ -206,12 +206,12 @@ export class GenericContainer implements TestContainer {
     return this;
   }
 
-  public withExposedPorts(...ports: Port[]): this {
+  public withExposedPorts(...ports: PortWithOptionalBinding[]): this {
```
cristianrgreco, line 209: the base interface `testcontainer.ts` also needs to be updated

cristianrgreco:
last comment from me, thanks for addressing the feedback so quickly

src/generic-container/generic-container.ts
codetheweb: Fixed in 88d7ca8.

cristianrgreco (approved):

cristianrgreco:
@codetheweb released in 8.4.0, thanks again

ab316:
Thank you guys for fixing this :)

PR #343, opened 2022-04-06
Support for health checks using CMD rather than CMD-SHELL

### LINKED ISSUE #342: Support for health checks using CMD rather than CMD-SHELL
Currently only `CMD-SHELL` health checks are supported which require a shell on the docker container to run. `distroless` docker images like [mockserver](https://hub.docker.com/r/mockserver/mockserver/) don't have a shell so that mechanism doesn't work, see https://github.com/GoogleContainerTools/distroless/issues/183.

To support docker containers without a shell the health check needs to be in the format
```
test: ['CMD', '/path/to/program', 'arg1', 'arg2', 'arg3']
```

richjharris:
Currently only `CMD-SHELL` health checks are supported which require a shell on the docker container to run. `distroless` docker images like [mockserver](https://hub.docker.com/r/mockserver/mockserver/) don't have a shell so that mechanism doesn't work, see https://github.com/GoogleContainerTools/distroless/issues/183.

To support docker containers without a shell the health check needs to be in the format
```
test: ['CMD', '/path/to/program', 'arg1', 'arg2', 'arg3']
```

Fixes #342


src/docker/functions/container/create-container.ts
```diff
@@ -111,7 +111,7 @@ const getHealthCheck = (healthCheck?: HealthCheck): DockerodeHealthCheck | undef
   }
 
   return {
-    Test: ["CMD-SHELL", healthCheck.test],
+    Test: typeof healthCheck.test === "string" ? ["CMD-SHELL", healthCheck.test] : ["CMD", ...healthCheck.test],
```
cristianrgreco, line 114: it's not expected for a user that the behaviour of the healthcheck should change if they provide their test as a string or an array of strings. I propose we change the type of the `test` field in `HealthCheck` to be as clear as possible and make this a breaking change. I'm thinking:

```typescript
export type HealthCheck = {
  test: ["CMD" | "CMD-SHELL", string];
  ...
}
```

as well as adding more information in the README on what CMD/CMD-SHELL means. What do you think?

src/generic-container/generic-container.test.ts
```diff
@@ -380,6 +380,22 @@ describe("GenericContainer", () => {
     expect(await getRunningContainerNames()).not.toContain(containerName);
   });
 
+  it("should handle health checks without the need of a shell", async () => {
```
cristianrgreco, line 383: i think the new test should be based off `should wait for custom health check`, with the use of `CMD` being called out very clearly

src/docker/functions/container/create-container.ts
richjharris: I would be disinclined to make it a breaking change, how about
```typescript
type HealthCheckCommon = {
  interval?: number;
  timeout?: number;
  retries?: number;
  startPeriod?: number;
};

type HealthCheckCommandShell = {
  test: string;
} & HealthCheckCommon;

type HealthCheckCommand = {
  testCmd: string[];
} & HealthCheckCommon;

export type HealthCheck = HealthCheckCommandShell | HealthCheckCommand;
```
and
```suggestion
    Test: 'testCmd' in healthCheck ? ["CMD", ...healthCheck.testCmd] : ["CMD-SHELL", healthCheck.test],
```
and to use it
```typescript
   const container = await customGenericContainer
      .withExposedPorts(8080)
      .withHealthCheck({
        testCmd: ["/usr/bin/wget", "--no-verbose", "--tries=1", "--spider", "http://localhost:8080/hello-world"],
      })
```
that way you can continue to use
```typescript
      customGenericContainer
        .withName(containerName)
        .withExposedPorts(8080)
        .withHealthCheck({ test: "exit 1" })
```
for the CMD-SHELL version.

src/generic-container/generic-container.test.ts
richjharris: 👍

src/docker/functions/container/create-container.ts
cristianrgreco: i would rather make a breaking change if it means having a simpler API and codebase. A user could provide both `test` and `testCmd` here and the behaviour is unexpected. They would also need to dig into the code to figure out what it means which isn't great. I think the code you've shared here compared to the interface change I proposed we could agree the latter is far simpler?

richjharris: ok, I'll make the breaking change

richjharris:
@cristianrgreco I've updated the PR to make the syntax for the health check test syntax more specific.

README.md
```diff
@@ -260,6 +260,8 @@ const container = await new GenericContainer("alpine")
   .start();
 ```
 
+`test` can either be `["CMD-SHELL", "some-shell --command || exit 1"]` which executes the health check using the shell in the container or `["CMD", "/path/to/command", "--arg1", "arg2", "--arg3"]` which executes the health check without using a shell to execute it. Note that this is a breaking change from v8.x.x, to upgrade from v8.x.x replace `test: "curl -f http://localhost || exit 1"` with `test: ["CMD-SHELL", "curl -f http://localhost || exit 1"]`.
```
cristianrgreco, line 263: I think something like this reads better, note I've removed the mention of the breaking change as this will be in the release notes..

---

To execute the `test` in a shell. use the form `["CMD-SHELL", "command"]`, for example: 
```javascript
["CMD-SHELL", "curl -f http://localhost:8000 || exit 1"]
```

To execute the `test` without a shell, use the form: `["CMD", "command", "arg1", "arg2"]`, for example:

```javascript
["CMD", "/usr/bin/wget", "--no-verbose", "--tries=1", "--spider", "http://localhost:8080/hello-world"]
```

src/generic-container/generic-container.test.ts
```diff
@@ -372,14 +372,32 @@ describe("GenericContainer", () => {
...
     ).rejects.toThrowError("Health check failed");
 
     expect(await getRunningContainerNames()).not.toContain(containerName);
   });
 
+  it("should wait for custom health check using CMD to run the command directly without a shell", async () => {
+    const container = await new GenericContainer("cristianrgreco/testcontainer:1.1.12")
+      .withExposedPorts(8080)
+      .withHealthCheck({
+        test: ["CMD", "/usr/bin/wget", "--no-verbose", "--tries=1", "--spider", "http://localhost:8080/hello-world"],
```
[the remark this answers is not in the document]
cristianrgreco: not to worry, this explained it for me: https://devops.stackexchange.com/questions/11501/healthcheck-cmd-vs-cmd-shell

README.md
richjharris: done

cristianrgreco (approved):
Thanks @richjharris. I will merge and deploy this once I've prepared the other major changes

cristianrgreco:
Will be looking to implement/merge this PR with https://github.com/testcontainers/testcontainers-node/pull/405

PR #344, opened 2022-04-06
Be able to set own project name on withNoRecreate

tiloio:
If you want to use testcontainers with multiple projects the "testcontainers-node" project name may already be used on your system.
So to be able to use your own projectName in that function would make sense to me.


cristianrgreco:
Hi @tiloio, will you be able to add a test for this change?

PR #347, opened 2022-04-11
feat: add `withLabel`

nitsanavni:


src/generic-container/generic-container.ts
```diff
@@ -221,6 +224,11 @@ export class GenericContainer implements TestContainer {
     return this;
   }
 
+  public withLabel({ key, value }: { key: string; value: string }): this {
```
cristianrgreco, line 227: i think that `withLabels({ [key:string]: value })` works better with JS. 

So instead of:

```typescript
withLabel({ key: 'a', value: 'a' })
withLabel({ key: 'b', value: 'b' })
```

You could simply do:

```typescript
withLabels({ a: 'a', b: 'b' })
```

nitsanavni: BTW, I see `withEnv` is structured similarly:
```ts
public withEnv(key: EnvKey, value: EnvValue)
```

cristianrgreco: Yes I did think that's where you got it from 🙂  I'm planning on changing that API as well

src/generic-container/generic-container.ts
```diff
@@ -221,6 +224,16 @@ export class GenericContainer implements TestContainer {
     return this;
   }
 
+  public withLabels(labels: Labels): this {
+    this.labels = { ...labels };
+    return this;
+  }
+
+  public addLabels(labels: Labels): this {
```
cristianrgreco, line 232: I am not sure about adding `addX` methods 🙂  
I think that `withX` can satisfy all cases, and halves the API we expose to users, makes changes easier, etc.

nitsanavni: got it 👍🏼
makes sense

README.md
```diff
@@ -243,6 +243,20 @@ const container = await new GenericContainer("alpine")
...
+
+```javascript
+const { GenericContainer } = require("testcontainers");
+
+const container = await new GenericContainer("alpine")
+  .withLabels({
+    label1: "value1",
+    label2: "value2",
+  })
+  .addLabels({ label3: "value3" })
```
cristianrgreco, line 256: Just this needs to be removed

nitsanavni: oops
sure thing

cristianrgreco:
Thanks @nitsanavni, I will look at releasing this tomorrow, will update once done.

cristianrgreco:
Released in 8.7.0. Thanks again!

PR #352, opened 2022-04-15
fix: make kafka containers reusable

### LINKED ISSUE #351: Kafka containers are not reusable
**Expected Behaviour**
Kafka containers should be reusable when the `withReuse` method is called before the container is started.

**Actual Behaviour**
Containers are kept alive after the test is done, but new ones are created each time.
This is because a random zookeeper port is being selected and exposed each time.

**Testcontainer Logs**
None

**Steps to Reproduce**
Run the following test:
```typescript
describe("KafkaContainer", () => {
  jest.setTimeout(240_000);

  it("should be reusable", async () => {
    const originalKafkaContainer = await new KafkaContainer().withReuse().start();
    const newKafkaContainer = await new KafkaContainer().withReuse().start();

    expect(newKafkaContainer.getId()).toBe(originalKafkaContainer.getId());

    await originalKafkaContainer.stop()
  });
})
```

**Environment Information**
- Operating System: macOS 12.2.1
- Docker Version: 20.10.6
- Node version: v17.9.0
- Testcontainers version: 8.7.0

apokralipsa:
This fixes #351 by making the zookeeper port stable.
This will also simplify calling commands to create a user so that the container can be used to test a connection to a cluster that requires SASL.

Alternatively the randomly generated port could not be exposed anymore. It should fix the reusability issue, but will not help with running commands within the container.

### DIFF
--- a/src/modules/kafka/kafka-container.test.ts
+++ b/src/modules/kafka/kafka-container.test.ts
@@ -59,6 +59,15 @@ describe("KafkaContainer", () => {
     await network.stop();
   });
 
+  it("should be reusable", async () => {
+    const originalKafkaContainer = await new KafkaContainer().withReuse().start();
+    const newKafkaContainer = await new KafkaContainer().withReuse().start();
+
+    expect(newKafkaContainer.getId()).toBe(originalKafkaContainer.getId());
+
+    await originalKafkaContainer.stop();
+  });
+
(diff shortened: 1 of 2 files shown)


cristianrgreco:
Great thank you. I see the same issue exists for at least the MySQL/ArangoDB modules, I will fix those too.

Issue #353, opened 2022-04-16
Feature request: Make setting up SASL and SSL in Kafka containers easier.


apokralipsa:
Production Kafka clusters may require the application to use encryption and to authenticate using the SASL mechanism.
In such a case it might be best to also configure the test container to require it, so that the test and production environments are as similar as possible.

Currently setting it up using the provided module is a bit cumbersome. It requires extending the `KafkaContainer` class so that the listener list set up in the `postStart` method can be changed, setting multiple environment variables and calling commands on the running container to create a user that can be used in the SASL authentication.

It would be nice to introduce a method to configure a separate SASL_SSL listener. Something like:
```typescript
const kafkaContainer = await new KafkaContainer()
      .withSaslSslListener({
        port: 9094,
        user: {
          name: "app-user",
          password: "userPassword"
        },
        keystore: {
          pfx: fs.readFileSync(path.resolve(pathToCertificates, "kafka.server.keystore.pfx")),
          passphrase: "serverKeystorePassword"
        },
        truststore: {
          pfx: fs.readFileSync(path.resolve(pathToCertificates, "kafka.server.truststore.pfx")),
          passphrase: "serverTruststorePassword"
        }
      })
      .withExposedPorts(9093, 9094).start();
``` 

Keeping the existing 9093 port would be useful so that the tests can produce messages without having to provide additional config.

I can provide a PR that will introduce this.

cristianrgreco:
Released in [v8.8.0](https://github.com/testcontainers/testcontainers-node/releases/tag/v8.8.0), thanks again @apokralipsa !

PR #354, opened 2022-04-17
feat: expose method to setup a SASL_SSL listener in a Kafka container

apokralipsa:
Introduces #353 

Currently only SCRAM-SHA-256 and SCRAM-SHA-512 mechanisms are supported.
Other mechanisms would require further complicating the `SaslSslListenerOptions` interface.
I am not sure how popular the remaining mechanisms are.

`SaslSslListenerOptions` accepts the keystore and truststore in PKCS12 format.
The default for the Kafka brokers is still JKS, but that is a Java specific format and can be transformed to PKCS12 using the tools provided by the Java Development Kit.


fixtures/docker/docker-kafka-generate-ssl-certs/Dockerfile
```diff
@@ -0,0 +1,4 @@
+FROM openjdk:8u322-jdk-buster
```
cristianrgreco, line 1: Is it possible to use an alpine-slim image to speed up the build or is there a reason not to?

src/modules/kafka/kafka-container.ts
```diff
@@ -6,27 +6,48 @@ import { StartedTestContainer } from "../..";
...
+  user: User;
+}
+
+export interface User {
+  name: string;
+  password: string;
+}
+
+export interface PKCS12CertificateStore {
+  content: Parameters<GenericContainer["withCopyContentToContainer"]>[0];
```
cristianrgreco, line 34: What does this mean? 🙂

cristianrgreco:
Thanks muchly for the PR @apokralipsa, looks like a great addition to the Kafka module!

src/modules/kafka/kafka-container.ts
apokralipsa: It means that this field accepts the same type as the first parameter of the `withCopyContentToContainer` method. Useful when we want to be secure for the future (in case the mentioned method changes), and if the method accepts a union of many types we don't want to copy all over the place :)

fixtures/docker/docker-kafka-generate-ssl-certs/Dockerfile
apokralipsa: Not sure if there is an alpine-slim image for JDK 8. Will have a look.

I need to use JDK 8, because the Kafka container uses that version underneath and certificates generated in newer versions don't seem to be supported.

apokralipsa: Switched to using `openjdk:8-slim` image.

cristianrgreco:
@apokralipsa the test is failing to access the certificates dir:

```
FAIL src/modules/kafka/kafka-container.test.ts (89.99 s)
  ● KafkaContainer › should expose SASL_SSL listener if configured

    EACCES: permission denied, open '/tmp/certs-1650368792034/kafka.client.truststore.pem'

      107 |       },
      108 |       ssl: {
    > 109 |         ca: [fs.readFileSync(path.resolve(pathOfCertificatesDir, "kafka.client.truststore.pem"))],
          |                 ^
      110 |       },
      111 |     });
      112 |

      at src/modules/kafka/kafka-container.test.ts:109:17
      at fulfilled (src/modules/kafka/kafka-container.test.ts:28:58)
          at runMicrotasks (<anonymous>)
```

apokralipsa:
I was not aware that on some machines in the CI the build will not have access to the OS temp directory. Switched to using a directory within the project and cleaning it afterwards.

cristianrgreco:
@apokralipsa Interestingly the problem is the same:

```
Summary of all failing tests
FAIL src/modules/kafka/kafka-container.test.ts (87.764 s)
  ● KafkaContainer › when a set of certificates is provided › should expose SASL_SSL listener if configured

    EACCES: permission denied, open '/home/runner/work/testcontainers-node/testcontainers-node/.certs-1650444214036/kafka.client.truststore.pem'

      116 |         },
      117 |         ssl: {
    > 118 |           ca: [fs.readFileSync(path.resolve(pathOfCertificatesDir, "kafka.client.truststore.pem"))],
          |                   ^
      119 |         },
      120 |       });
      121 |       await kafkaContainer.stop();

      at src/modules/kafka/kafka-container.test.ts:118:19
      at fulfilled (src/modules/kafka/kafka-container.test.ts:28:58)
          at runMicrotasks (<anonymous>)
```

apokralipsa:
This time the file could not be created within the project directory on a different CI server. 

Simplified the test to avoid generating certificates altogether. Left the docker file and some md so that users can generated their own certificates if they need them.

src/modules/kafka/test-certs/Dockerfile
```diff
@@ -0,0 +1,5 @@
+FROM openjdk:8-slim
```
cristianrgreco, line 1: The `test-certs` directory will need to be excluded in the `tsconfig.json`, to avoid it being packaged

src/modules/kafka/kafka-container.ts
```diff
@@ -6,27 +6,48 @@ import { StartedTestContainer } from "../..";
...
 import { InspectResult } from "../../docker/functions/container/inspect-container";
-import { dockerClient } from "../../docker/docker-client";
 
 const KAFKA_PORT = 9093;
 const KAFKA_BROKER_PORT = 9092;
 const DEFAULT_ZOOKEEPER_PORT = 2181;
 
 export const KAFKA_IMAGE = "confluentinc/cp-kafka:5.5.4";
 
+export interface SaslSslListenerOptions {
```
cristianrgreco, line 16: Currently only the containers are exported from the Kafka module: https://github.com/testcontainers/testcontainers-node/blob/51654f5cceb7c8e2757d82102761d345fbe24979/src/index.ts#L17. I think it's OK not to export these SASL interfaces but maybe users would want them? What do you think?

src/modules/kafka/test-certs/Dockerfile
apokralipsa: Good point. Done.

src/modules/kafka/kafka-container.ts
apokralipsa: I think it's a good idea. The intention is to provide the config inline. Done.

cristianrgreco (approved):

cristianrgreco:
Thanks muchly @apokralipsa !

PR #357, opened 2022-04-24
feat: add support for specifying host port on module container instan…

### LINKED ISSUE #356: Feature request: allow configuration of exposed ports for PostgreSQL container
When calling `start` on an instance of `PostgreSQLContainer` https://github.com/testcontainers/testcontainers-node/blob/7580ad550b2e4a2221f55c75a7bb0774de6c07f0/src/modules/postgresql/postgresql-container.ts#L32 the exposed ports are hardcoded to `5432` so calling `withExposedPorts` while configuring the container does not work. 

A workaround is using an instance of `GenericContainer`.

Happy to create a PR if this issue is accepted.

Robin-Hoodie:
…ces just like you can on GenericContainer. Fixes #356

Consumers can now use `withHostPort` on modules to specify a specific host port. (except for `KafkaContainer` where this is specified in the ctor)
If `withHostPort` is not called, the current behaviour is kept where a random port is assigned on the host system.

### DIFF
--- a/src/generic-container/generic-container.ts
+++ b/src/generic-container/generic-container.ts
@@ -214,6 +214,10 @@ export class GenericContainer implements TestContainer {
     boundPorts: BoundPorts
   ): Promise<void>;
 
+  protected get hasExposedPorts(): boolean {
+    return Boolean(this.ports.length);
+  }
+
   public withCmd(cmd: Command[]): this {
     this.cmd = cmd;
     return this;
--- a/src/generic-container/generic-container.ts
+++ b/src/generic-container/generic-container.ts
@@ -215,7 +215,7 @@ export class GenericContainer implements TestContainer {
   ): Promise<void>;
 
   protected get hasExposedPorts(): boolean {
-    return Boolean(this.ports.length);
+    return this.ports.length !== 0;
   }
 
   public withCmd(cmd: Command[]): this {
--- a/src/modules/kafka/kafka-container.ts
+++ b/src/modules/kafka/kafka-container.ts
@@ -1,6 +1,6 @@
 import { GenericContainer } from "../../generic-container/generic-container";
 import { BoundPorts } from "../../bound-ports";
-import { Port } from "../../port";
+import { Port, PortWithOptionalBinding } from "../../port";
 import { RandomUuid, Uuid } from "../../uuid";
 import { StartedTestContainer } from "../..";
 import { AbstractStartedContainer } from "../abstract-started-container";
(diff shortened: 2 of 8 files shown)


Robin-Hoodie:
@cristianrgreco You're right, I hadn't considered that. I did check for `this.ports.length` instead of relying on `this.ports` being `null|undefined` as `this.ports` is [initialized with an empty array](https://github.com/testcontainers/testcontainers-node/blob/e11f0b24b9dfbe15ec0bd39179347dd25b3b66f3/src/generic-container/generic-container.ts#L58)

src/modules/kafka/kafka-container.ts
```diff
@@ -43,10 +43,10 @@ export class KafkaContainer extends GenericContainer {
   private zooKeeperPort?: Port;
   private saslSslConfig?: SaslSslListenerOptions;
 
-  constructor(image = KAFKA_IMAGE) {
+  constructor(image = KAFKA_IMAGE, ports?: PortWithOptionalBinding[]) {
```
cristianrgreco, line 46: Why add `ports` to the constructor when there's the `withExposedPorts` method?

src/generic-container/generic-container.ts
```diff
@@ -214,6 +214,10 @@ export class GenericContainer implements TestContainer {
     boundPorts: BoundPorts
   ): Promise<void>;
 
+  protected get hasExposedPorts(): boolean {
+    return Boolean(this.ports.length);
```
cristianrgreco, line 218: This could be more explicit, instead of type coercion from number->boolean, this could just be `this.ports.length ===0`

Robin-Hoodie: Sure, I can change that, but what's the advantage of doing it that way?

src/modules/kafka/kafka-container.ts
Robin-Hoodie: I figured there was a reason `withExposedPorts` was being called in the ctor here instead of in `start` like the other modules. Are you suggesting we just don't make any changes to `KafkaContainer`?

src/generic-container/generic-container.ts
cristianrgreco: > This could be more explicit, instead of type coercion from number->boolean
🙂

src/modules/kafka/kafka-container.ts
cristianrgreco: Let's omit this change from the Kafka module for now, the ports are intertwined with the SASL configuration and the ZK. It's probably more complexity than it's worth at the moment.

src/generic-container/generic-container.ts
Robin-Hoodie: Fair enough. Refactored in dfd591af969b814dde593b73d7abb87fdadec4c9

cristianrgreco (approved):

Robin-Hoodie:
@cristianrgreco Can we merge this one in?

cristianrgreco:
@Robin-Hoodie Released in [8.9.0](https://github.com/testcontainers/testcontainers-node/releases/tag/v8.9.0)

PR #364, opened 2022-05-06
Implement restart container for `StartedTestContainer`

### LINKED ISSUE #340: Feature request: Ability to restart container
As of right now, there is no possible way provided to restart started containers. I would like to see a exposed method to restart a given container.

The use case for this may be specific to me (or may not be), but generally there is a need for me to change the configuration file and push for a restart after the config change. As of right now, I am trying to work it through via the use of `Dockerode`, which this is built on top of, but found no success.

I am willing to implement this myself for this feature, but would like to know the project maintainer's thought on this feature, considering it does not exists in `testcontainer-java`.

Related: https://github.com/testcontainers/testcontainers-java/issues/606

cwkang1998:
Adds functionality to allow restarting of `StartedTestContainers`.

Closes #340

### DIFF
--- a/README.md
+++ b/README.md
@@ -99,16 +99,6 @@ const startedContainer: StartedTestContainer = await container.start();
 const stoppedContainer: StoppedTestContainer = await startedContainer.stop();
 ```
 
-Restarting a container:
-
-```javascript
-const { GenericContainer } = require("testcontainers");
-
-const container = await new GenericContainer("alpine")
-  .start();
-const restartedContainer = container.restart();
-```
-
 Using a specific image version:
 
 ```javascript
@@ -170,6 +160,16 @@ const container2 = await new GenericContainer("alpine")
   .start();
 
 assert(container1.getId() === container2.getId());
+```
+
+Restarting a container:
+
+```javascript
+const { GenericContainer } = require("testcontainers");
+
+const container = await new GenericContainer("alpine")
+  .start();
+const restartedContainer = await container.restart();
 ```
 
 Creating a container with a specified name:
--- a/README.md
+++ b/README.md
@@ -99,6 +99,16 @@ const startedContainer: StartedTestContainer = await container.start();
 const stoppedContainer: StoppedTestContainer = await startedContainer.stop();
 ```
 
+Restarting a container:
+
+```javascript
+const { GenericContainer } = require("testcontainers");
+
+const container = await new GenericContainer("alpine")
+  .start();
+const restartedContainer = container.restart();
+```
+
 Using a specific image version:
 
 ```javascript
--- a/src/generic-container/started-generic-container.ts
+++ b/src/generic-container/started-generic-container.ts
@@ -18,8 +18,8 @@ export class StartedGenericContainer implements StartedTestContainer {
   constructor(
     private readonly container: Dockerode.Container,
     private readonly host: Host,
-    private readonly inspectResult: InspectResult,
-    private readonly boundPorts: BoundPorts,
+    private inspectResult: InspectResult,
+    private boundPorts: BoundPorts,
     private readonly name: ContainerName,
     private readonly waitStrategy: WaitStrategy
   ) {}
@@ -28,23 +28,16 @@ export class StartedGenericContainer implements StartedTestContainer {
     return this.stopContainer(options);
   }
 
-  public async restart(options: Partial<RestartOptions> = {}): Promise<StartedGenericContainer> {
+  public async restart(options: Partial<RestartOptions> = {}): Promise<void> {
     const resolvedOptions: RestartOptions = { timeout: 0, ...options };
     await restartContainer(this.container, resolvedOptions);
-    const newInspectRes = await inspectContainer(this.container);
-    const newBoundPorts = await BoundPorts.fromInspectResult(newInspectRes).filter(
+
+    // Get the restart container with its new bound ports and update it
+    this.inspectResult = await inspectContainer(this.container);
+    this.boundPorts = await BoundPorts.fromInspectResult(this.inspectResult).filter(
       Array.from(this.boundPorts.iterator()).map((port) => port[0])
     );
-    await this.waitForContainer(this.container, newBoundPorts);
-
-    return new StartedGenericContainer(
-      this.container,
-      this.host,
-      newInspectRes,
-      newBoundPorts,
-      this.name,
-      this.waitStrategy
-    );
+    await this.waitForContainer(this.container, this.boundPorts);
   }
 
   private async stopContainer(options: Partial<StopOptions> = {}): Promise<StoppedGenericContainer> {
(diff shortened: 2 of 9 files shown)


src/generic-container/started-generic-container.ts
```diff
@@ -11,20 +11,42 @@ import { Readable } from "stream";
...
 
   public async stop(options: Partial<StopOptions> = {}): Promise<StoppedTestContainer> {
     return this.stopContainer(options);
   }
 
+  public async restart(options: Partial<RestartOptions> = {}): Promise<StartedGenericContainer> {
+    const resolvedOptions: RestartOptions = { timeout: 0, ...options };
+    await restartContainer(this.container, resolvedOptions);
+    const newInspectRes = await inspectContainer(this.container);
+    const newBoundPorts = await BoundPorts.fromInspectResult(newInspectRes).filter(
```
cristianrgreco, line 35: Do the bound ports change when restarting the container?

src/generic-container/started-generic-container.ts
```diff
@@ -35,6 +57,24 @@ export class StartedGenericContainer implements StartedTestContainer {
     return new StoppedGenericContainer();
   }
 
+  private async waitForContainer(container: Dockerode.Container, boundPorts: BoundPorts): Promise<void> {
```
cristianrgreco, line 60: This is duplicated from `generic-container`, perhaps we should extract this to a dedicated `wait-for-container` module, which accepts the wait strategy

cristianrgreco:
This is great, thanks @cwkang1998. I left a couple of comments. Could you please add an example of restarting the container to the README?

cristianrgreco:
I think you'll need to support `restart` for the modules. Hopefully it's as easy as adding it to the `AbstractStartedContainer`

src/generic-container/started-generic-container.ts
cwkang1998: Yes, after restarting it, it seems that a new port is assigned to them.

src/generic-container/started-generic-container.ts
cwkang1998: Maybe we should open another PR for this instead? Since it make changes across multiple files, would like to keep the PRs focused. I can open a seperate one for it.

cwkang1998:
Hmm, things still failing. will get it fixed asap.  @cristianrgreco do give me one more week for this

cwkang1998:
> Hmm, things still failing. will get it fixed asap. @cristianrgreco do give me one more week for this

@cristianrgreco Finally got back to this, should be ready to review now!

Thanks!

src/generic-container/started-generic-container.ts
```diff
@@ -11,20 +11,42 @@ import { Readable } from "stream";
...
+  public async restart(options: Partial<RestartOptions> = {}): Promise<StartedGenericContainer> {
+    const resolvedOptions: RestartOptions = { timeout: 0, ...options };
+    await restartContainer(this.container, resolvedOptions);
+    const newInspectRes = await inspectContainer(this.container);
+    const newBoundPorts = await BoundPorts.fromInspectResult(newInspectRes).filter(
+      Array.from(this.boundPorts.iterator()).map((port) => port[0])
+    );
+    await this.waitForContainer(this.container, newBoundPorts);
+
+    return new StartedGenericContainer(
```
cristianrgreco, line 40: I wonder if the return type should be `Promise<void>` instead of `Promise<StartedGenericContainer>`. I say that because the new `StartedGenericContainer` is the same as the previous. Also, if I have a reference to the previous instance and for example `stop` it, the restarted container instance will also be stopped. I think therefore it should return `void` to make clear that you are working with the same container

src/modules/abstract-started-container.ts
```diff
@@ -1,15 +1,19 @@
-import { StartedTestContainer, StopOptions, StoppedTestContainer } from "../test-container";
+import { RestartOptions, StartedTestContainer, StopOptions, StoppedTestContainer } from "../test-container";
 import { Port } from "../port";
 import { ContainerName, Host, Id, Command, ExecResult, Labels } from "../docker/types";
 import { Readable } from "stream";
 
 export class AbstractStartedContainer {
-  constructor(protected readonly startedTestContainer: StartedTestContainer) {}
+  constructor(protected startedTestContainer: StartedTestContainer) {}
```
cristianrgreco, line 7: This could remain `readonly`

cristianrgreco:
Nice one @cwkang1998, the build's passed. I left a couple of comments, other than that LGTM

README.md
```diff
@@ -99,6 +99,16 @@ const startedContainer: StartedTestContainer = await container.start();
...
 ```
 
+Restarting a container:
+
+```javascript
+
+  .start();
+const restartedContainer = container.restart();
```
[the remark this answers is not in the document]
cristianrgreco: Could we also move this further down, perhaps above `Creating a container with a specified name:` ? At this point we haven't even covered using an image with a specific version 🙂

src/generic-container/started-generic-container.ts
cwkang1998: Hmm, I've decided to return a `StartedGenericContainer` because the bound ports was `readonly` and I kinda thought above conserving it to be that way, but now that I thought about it, it does make a lot more sense. Will make this change now.

src/generic-container/started-generic-container.ts
```diff
@@ -28,23 +28,16 @@ export class StartedGenericContainer implements StartedTestContainer {
...
-  public async restart(options: Partial<RestartOptions> = {}): Promise<StartedGenericContainer> {
+  public async restart(options: Partial<RestartOptions> = {}): Promise<void> {
     const resolvedOptions: RestartOptions = { timeout: 0, ...options };
     await restartContainer(this.container, resolvedOptions);
-    const newInspectRes = await inspectContainer(this.container);
-    const newBoundPorts = await BoundPorts.fromInspectResult(newInspectRes).filter(
+
+    // Get the restart container with its new bound ports and update it
+    this.inspectResult = await inspectContainer(this.container);
+    this.boundPorts = await BoundPorts.fromInspectResult(this.inspectResult).filter(
```
cristianrgreco, line 37: Do the ports change when the container is restarted?

cwkang1998: Yup. That's the reason why directly using `dockerrode` to perform a restart fails.

This is how I work around this limitation before:

```typescript
// Explicitly change the typing to any
const container: any = this.requireContainer();
const inspectRes = await inspectContainer(dockerrodeContainer);
const boundPorts: BoundPorts = (BoundPorts as any).fromInspectResult(inspectRes).filter(DeFiDContainer.DefiDPorts[this.network]);

// Since its a any type, typescript does not scream at us
// Javascript allows for this during run time
container.inspectResult = inspectRes;
container.boundPorts = boundPorts;
```

cristianrgreco (approved):

Issue #367, opened 2022-05-22
No Docker client strategy found caused by invoked exported dockerClient


calvintwr:
**Expected Behaviour**
No error should occur when setup files involving `dockerClient` are called.
...

**Actual Behaviour**
`No Docker client strategy found` error appears.
...

**Testcontainer Logs**
With `DEBUG=testcontainers*` on, testcointainers will run:
```
  testcontainers DEBUG Found applicable Docker client strategy: UnixSocketStrategy +0ms
  testcontainers DEBUG Testing Docker client strategy URI: unix:///var/run/docker.sock +1s
  testcontainers WARN  Docker daemon is not reachable: Error: connect ENOENT /var/run/docker.sock +118ms
  testcontainers WARN  Docker client strategy UnixSocketStrategy is not reachable +0ms
```
And the error:
```
No Docker client strategy found

      at node_modules/testcontainers/dist/docker/docker-client.js:46:11
      at fulfilled (node_modules/testcontainers/dist/docker/docker-client.js:5:58)
```
...

**Steps to Reproduce**
1. Write a generic setup "boilerplate" file like:

```ts
// setup.ts:
export const container = await new GenericContainer('postgres').withExposedPorts(5432).start()
```

```ts
// index.ts
import { container } from './setup.ts'

function start() {
    return null
}
start()
```

2. When you run `$ ts-node index.ts`, notice that you will get the error. But `container` was never used, only imported.

**Environment Information**
- Operating System: MacOS Monterey 12.4
- Docker Version: 4.8.1
- Node version: 16.13.1
- Testcontainers version: v8.10.1

cristianrgreco:
@calvintwr share the test containers logs

calvintwr:
@cristianrgreco I have created the PR for the fix.

cristianrgreco:
@calvintwr I'm more interested in fixing why the Docker client was not found instead of making it lazy

calvintwr:
@cristianrgreco the issue is not that it’s not found. The issue has more to do the search (`dockerClient`) being invoked when it shouldn’t have been. You can check my PR and advise how best to resolve this further.

PR #368, opened 2022-05-22
fix for "No Docker client strategy found" despite not invoked

calvintwr:
fix for https://github.com/testcontainers/testcontainers-node/issues/367

### DIFF
--- a/src/docker-compose-environment/docker-compose-environment.ts
+++ b/src/docker-compose-environment/docker-compose-environment.ts
@@ -142,7 +142,7 @@ export class DockerComposeEnvironment {
 
           return new StartedGenericContainer(
             container,
-            (await dockerClient).host,
+            (await dockerClient()).host,
             inspectResult,
             boundPorts,
             containerName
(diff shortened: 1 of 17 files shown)


cristianrgreco:
@calvintwr seems the PR is failing the tests

calvintwr:
@cristianrgreco yea let me check it out.

calvintwr:
@cristianrgreco can I get some help?

calvintwr:
@cristianrgreco great got it thanks. i think in your suggestion, given that `getDockerClient` will be invoked, it might still result in the same issue. i'll see if I can come up with something, if not there might be practical fix for this. And honestly it doesn't require a fix. It just stands in the way when one tries to create abstractions.

cristianrgreco:
> i think in your suggestion, given that getDockerClient will be invoked, it might still result in the same issue

That shouldn't be the case, `getDockerClient` is only invoked when `dockerClient()` is called, whereas before it was invoked when imported, so unless you call testcontainers code it shouldn't invoke, or am I missing something?

PR #389, opened 2022-08-28
Fix withCopyFileToContainer for large files

dmitri-gb:
This was broken in v7.13.0 after the changes in [#224](https://github.com/testcontainers/testcontainers-node/pull/224/files#diff-e9908fa461de235d8997a04799b787ee3b57d9682d8262c0ed65b9af47255086R213).

`Archiver.finalize` will simply wait for the `end` event which cannot fire until the stream has been consumed by the following `putContainerArchive` call.


cristianrgreco:
@dmitri-gb Thanks!

PR #395, opened 2022-09-14
Pass Dockerode ExecStartOptions to StartedGenericContainer.exec

Ovich:
For now the StartedGenericContainer.exec only takes Command[] as argument. startExec sends a hardcoded Dockerode.ExecStartOptions to Dockerode.Exec. 

It might be intressting to send an optional exec options object to StartedGenericContainer.exec. In our use case we need the container to run in "tty : false" mode.

### DIFF
--- a/src/docker/functions/container/exec-container.ts
+++ b/src/docker/functions/container/exec-container.ts
@@ -4,9 +4,9 @@ import Dockerode from "dockerode";
 import { log } from "../../../logger";
 
 type ExecContainerOptions = {
-  Tty: boolean;
+  tty: boolean;
   stdin: boolean;
-  Detach: boolean;
+  detach: boolean;
 };
 
 export const execContainer = async (
--- a/src/docker/functions/container/exec-container.ts
+++ b/src/docker/functions/container/exec-container.ts
@@ -41,6 +41,7 @@ const startExec = async (exec: Dockerode.Exec, options?: Dockerode.ExecStartOpti
       stdout: true,
       stderr: true,
     };
+
     const stream = await exec.start(options || defaultOptions);
     stream.setEncoding("utf-8");
     return stream;
--- a/src/docker/functions/container/exec-container.ts
+++ b/src/docker/functions/container/exec-container.ts
@@ -3,8 +3,10 @@ import { Command, ExecResult, ExitCode } from "../../types";
 import Dockerode from "dockerode";
 import { log } from "../../../logger";
 
-export type ExecContainerOptions = {
+type ExecContainerOptions = {
   Tty: boolean;
+  stdin: boolean;
+  Detach: boolean;
 };
 
 export const execContainer = async (
--- a/src/docker/functions/container/exec-container.ts
+++ b/src/docker/functions/container/exec-container.ts
@@ -3,7 +3,11 @@ import { Command, ExecResult, ExitCode } from "../../types";
 import Dockerode from "dockerode";
 import { log } from "../../../logger";
 
-export const execContainer = async (container: Dockerode.Container, command: Command[]): Promise<ExecResult> => {
+export const execContainer = async (
+  container: Dockerode.Container,
+  command: Command[],
+  options?: Dockerode.ExecStartOptions
+): Promise<ExecResult> => {
   try {
     const exec = await container.exec({
       Cmd: command,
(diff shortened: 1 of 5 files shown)


src/docker/functions/container/exec-container.ts
```diff
@@ -27,17 +31,18 @@ export const execContainer = async (container: Dockerode.Container, command: Com
   }
 };
 
-const startExec = async (exec: Dockerode.Exec): Promise<Readable> => {
+const startExec = async (exec: Dockerode.Exec, options?: Dockerode.ExecStartOptions): Promise<Readable> => {
   try {
-    const options = {
+    const defaultOptions = {
```
cristianrgreco, line 36: I guess we should merge `options` and `defaultOptions`, else if someone only specifies `Tty`, then the other fields wouldn't be set

src/generic-container/started-generic-container.ts
```diff
@@ -100,8 +100,8 @@ export class StartedGenericContainer implements StartedTestContainer {
     return this.inspectResult.networkSettings[networkName].ipAddress;
   }
 
-  public exec(command: Command[]): Promise<ExecResult> {
-    return execContainer(this.container, command);
+  public exec(command: Command[], options?: Dockerode.ExecStartOptions): Promise<ExecResult> {
```
cristianrgreco, line 103: We don't expose dockerode options directly, instead we create/export our own so that people don't have to import transitive dependencies

src/docker/functions/container/exec-container.ts
Ovich: Thank you very much for your prompt response.

That’s a great idea. Inspired by how StopOptions are implemented for the stopContainer method, I made following changed:
#### interface ExecOptions is created in "test-containers.ts"
```
// from "test-containers.ts"
export interface ExecOptions {
  Tty: boolean;
  Detach: boolean;
  stdin: boolean;
}
```
- Merging of default and specific ExecOptions properties is done within "started-generic-container.ts" in a private method
- ExecContainerOptions type is declared in "exec-container" for local use -> based on StopContainerOptions from "stop-container.ts"
- execContainer in "exec-container.ts" now receives options as mandatory argument, merging is done by the caller.

## Unused properties
Some of the previously hardcoded options, such as stdout and stderr, are inexistent in the Dockerode.ExecStartOptions specification. So, I also removed these unused options in interface and type definitions.
```
// From Dockerode
interface ExecStartOptions {
        // hijack and stdin are used by docker-modem
        hijack?: boolean | undefined;
        stdin?: boolean | undefined;
        // Detach and Tty are used by Docker's API
        Detach?: boolean | undefined;
        Tty?: boolean | undefined;
        abortSignal?: AbortSignal;
}
```

## Hardcoded argument

"port-check.ts" makes a direct call to "execContainer", by-passing the merging of options that is done via "StartedGenericContainer ". 
That’s the only usage of “execContainer” function from outside of the “StartedGenericContainer” instance. 
We have 2 possibilities: 
-  keep it as is, with the hardcoded argument in "port-check.ts" 
- do the merging of options within “startExec” function, exporting the “ExecContainerOptions” for use in "started-generic-container” and removing the ExecOptions interface.

src/test-container.ts
```diff
@@ -63,6 +63,12 @@ export interface StopOptions {
   removeVolumes: boolean;
 }
 
+export interface ExecOptions {
+  Tty: boolean;
```
cristianrgreco, line 67: let's use camelCase for any publicly exposed definitions. You'll also need to update the `exec` method in the `StartedTestContainer` interface below

Ovich: the idea was to keep the same case as the Dockerode specification. 

interface ExecOptions and type ExecContainerOptions now have camelCase attributs. StartedTestContainer updated accordingly

cristianrgreco: you'll also need to update the exec method in the StartedTestContainer to add the new options

Ovich: I included  options? in the exec method signature in StartedTestContainer

src/generic-container/started-generic-container.ts
```diff
@@ -100,8 +111,8 @@ export class StartedGenericContainer implements StartedTestContainer {
     return this.inspectResult.networkSettings[networkName].ipAddress;
   }
 
-  public exec(command: Command[]): Promise<ExecResult> {
-    return execContainer(this.container, command);
+  public exec(command: Command[], options: Partial<ExecOptions> = {}): Promise<ExecResult> {
+    return this.execContainer(command, options);
```
cristianrgreco, line 115: could you please inline this private method? having 2 functions called `execContainer` is a bit confusing

cristianrgreco:
thanks for addressing the feedback @Ovich

cristianrgreco (approved):

PR #399, opened 2022-10-18
add flag to log exec output stream

abhinav-byahut:


src/docker/functions/container/exec-container.ts
```diff
@@ -24,7 +25,10 @@ export const execContainer = async (
     const stream = await startExec(exec, options);
 
     const chunks: string[] = [];
-    stream.on("data", (chunk) => chunks.push(chunk));
+    stream.on("data", (chunk) => {
+      if (options.debug) log.debug(chunk);
```
cristianrgreco, line 29: instead of supporting an extra argument, it would probably be better to always log the chunks in the catch block below, something like:

log.error(`Failed to exec container ${container.id} with command "${command.join(" ")}": ${err}, output: ${chunks.join("")}`);

abhinav-byahut: @cristianrgreco main purpose of this flag is to print debug logs for execution & not wait for execution to finish before being able to see the logs. It would be helpful to have a feature that would provide logs for exec as they come.

cristianrgreco: @abhinav-byahut only thing I don't like is that because the chunks may not be newline terminated, it would be hard to read along with the other logs. I wonder if it'd be better to create a separate logger (e.g `DEBUG=testcontainers:exec`) and create a new stream `byline` which outputs to it?

abhinav-byahut: @cristianrgreco sounds good to me, will update the PR.

abhinav-byahut: @cristianrgreco updated the code to support separate logger, rather that a flag.

cristianrgreco:
I've created a separate PR which achieves this: https://github.com/testcontainers/testcontainers-node/pull/404, it's easier than going back and forth via comments 🙂 The 2 main diffs are making the logs optional, we run a lot of exec while polling the containers to be ready. Also the `byline` is used to create a separate stream to ensure the existing functionality works when chunks don't contain newlines

PR #408, opened 2022-10-25
Define required node version

ilopezluna:
## What does this PR do?

Defines the minimum node version (`>= 10.16`) to build the project.
Configures [engine-strict](https://docs.npmjs.com/cli/v8/using-npm/config#engine-strict) to prevent builds using any previous version.
The minimum node version for the current dependency graph has been calculated using [ls-engines](https://www.npmjs.com/package/ls-engines).

This PR also updates the preferred node version through `.nvmrc` file.

## Why is it important?

To define the minimum/preferred node version will help to prevent errors running the project with an unsupported node version. 

## How to test this PR

- Run `npm install` with a previous node version, the command should fail because an invalid node version
- Run `nvm use` and then `npm install` the package-lock.json should be identical as it is


cristianrgreco:
Thanks @ilopezluna. I have a PR in progress which will be a major release. As this is a breaking change I will merge it into that PR as soon as the build passes. Thanks for raising

PR #414, opened 2022-11-30
Update ryuk and sshd version images

eddumelendez:
There are some improvements such as sshd is now built in order to run
in multi-platform.


cristianrgreco:
Hi @eddumelendez, thanks for raising. Looks like the port-forwarder tests, which use the sshd container, are failing with the new version. Could you take a look?

PR #417, opened 2022-12-12
adding new volumes option and tests

SadSadko:
Adding volume feature, according to docker doc, Volumes are the preferred mechanism for persisting data generated by and used by Docker containers.

I exposed the volume parameter as a new Volume type to match the other options

### DIFF
--- a/src/docker/functions/container/create-container.ts
+++ b/src/docker/functions/container/create-container.ts
@@ -4,7 +4,7 @@ import { dockerClient } from "../../docker-client";
 import Dockerode, { PortMap as DockerodePortBindings } from "dockerode";
 import { getContainerPort, hasHostBinding, PortWithOptionalBinding } from "../../../port";
 import { createLabels } from "../create-labels";
-import { BindMount, Environment, ExtraHost, HealthCheck, Labels, TmpFs, Ulimits } from "../../types";
+import { BindMount, Environment, ExtraHost, HealthCheck, Labels, TmpFs, Ulimits, Volume } from "../../types";
 
 export type CreateContainerOptions = {
   imageName: DockerImageName;
@@ -28,6 +28,7 @@ export type CreateContainerOptions = {
   addedCapabilities?: string[];
   droppedCapabilities?: string[];
   user?: string;
+  volumes?: Volume[];
 };
 
 export const createContainer = async (options: CreateContainerOptions): Promise<Dockerode.Container> => {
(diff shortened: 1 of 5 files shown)


src/docker/functions/container/create-container.ts
```diff
@@ -161,3 +163,9 @@ const getUlimits = (ulimits: Ulimits | undefined): DockerodeUlimit[] | undefined
     Soft: value.soft,
   }));
 };
+
+function getVolumes(volumes: Volume[]): { [volume: string]: Record<string, unknown> } | undefined {
```
cristianrgreco, line 167: This won't return `undefined` because you're checking if `volumes` is truthy before calling this. 
This function can also be refactored:

```javascript
const getVolumes = (volumes: Volume[]): { [volume: string]: Record<string, never> } => {
  return volumes.reduce((prev, next) => ({ ...prev, [next.name]: {} }), {});
};
```

src/generic-container/generic-container.test.ts
```diff
@@ -509,6 +510,39 @@ describe("GenericContainer", () => {
...
+
+    const container = await new GenericContainer("cristianrgreco/testcontainer:1.1.12")
+      .withExposedPorts(8080)
+      .withVolumes(volume)
+      .start();
+
+    const dockerContainer = await getContainerById(container.getId());
+    const containerInfo = await dockerContainer.inspect();
+    await container.stop();
+    expect(containerInfo.Mounts.length).toBe(1);
```
cristianrgreco, line 524: Could this be tested similar to the bind mounts, where the actual behaviour is tested, instead of checking the container info? I.e you should be able to mount a volume and interact with it and assert its values

src/generic-container/generic-container.test.ts
```diff
@@ -509,6 +510,39 @@ describe("GenericContainer", () => {
...
+    expect(containerInfo.Mounts[0].Destination).toBe(volume.name);
+  });
+
+  it("should create container with array of volumes", async () => {
```
cristianrgreco, line 528: There isn't any difference in behaviour if you pass one or multiple, so this test can be removed

src/generic-container/generic-container.ts
```diff
@@ -366,4 +368,9 @@ export class GenericContainer implements TestContainer {
     }
     return this.tarToCopy;
   }
+
+  public withVolumes(...volumes: Volume[]): this {
```
cristianrgreco, line 372: This may be more intuitive without using varargs, i.e:

```javascript
.withVolumes(
  { name: "vol1" },
  { name: "vol2" },
)
```

Is less intuitive than:

```javascript
.withVolumes([
  { name: "vol1" },
  { name: "vol2" },
])
```

cristianrgreco:
@SadSadko few comments, thanks for raising the PR!

src/generic-container/generic-container.test.ts
SadSadko: Good idea, on it.

redjolr:
Hi @cristianrgreco @SadSadko 
Will we have this feature any time soon? I would love to bind volumes to containers, especially host volumes (volumes that we can use to share data between the container and the host).

PR #427, opened 2022-12-18
Added a `buildAs` method to GenericContainerBuilder to allow custom b…

leonwilly:
I would like to add custom methods to the GenericContainer without having to encapsulate the entire container. The `buildAs` accepts a type that should be a subclass of GenericContainer.

### DIFF
--- a/src/generic-container/generic-container-builder.ts
+++ b/src/generic-container/generic-container-builder.ts
@@ -39,6 +39,13 @@ export class GenericContainerBuilder {
   }
 
   public async build(image = `${this.uuid.nextUuid()}:${this.uuid.nextUuid()}`): Promise<GenericContainer> {
+    return this.buildAs(GenericContainer, image);
+  }
+
+  public async buildAs<T extends GenericContainer>(
+    type: new (image: string) => T,
+    image = `${this.uuid.nextUuid()}:${this.uuid.nextUuid()}`
+  ): Promise<T> {
     const imageName = DockerImageName.fromString(image);
 
     await ReaperInstance.getInstance();
@@ -57,7 +64,7 @@ export class GenericContainerBuilder {
       cache: this.cache,
       registryConfig,
     });
-    const container = new GenericContainer(imageName.toString());
+    const container = new type(imageName.toString());
 
     if (!(await imageExists((await dockerClient()).dockerode, imageName))) {
       throw new Error("Failed to build image");


leonwilly:
If you all end up allowing this I'll gladly add the required documentation to the Readme.

leonwilly:
The reasoning for this is I wrote a library that wraps the `tc` command for the  `iproute2` package. I would like to extend the generic container to include a `withTrafficControlSetting` that will execute the command similar to `withCommand` at start. If I were to take the same approach as `AbstractStartedContainer` I would have to write a wrapper method for every method in the GenericContainer or do something less elegant. Since there isn't a way to return a custom GenericContainer from the build command. I do agree this method isn't ideal.  Adding generics throughout the library to be used with a single `build` method would be the right way to do it. If it's something you'd welcome I would be up for the task. Otherwise this solution will suffice even if it's our own  forked version. I just thought I'd share it and see if you could see it being useful. As always there are several ways to solve this problem. I generally prefer inheritance as a first class citizen over composition.

cristianrgreco:
I see, it's a pain to compose the GenericContainer/StartedGenericContainer because you'd need to re-declare the methods. 

Let me have a think about this, whether it's exposing an AbstractGenericContainer/AbstractStartedGenericContainer that people can extend, or even simplifying the constructors of the GenericContainer/StartedGenericContainer such that they can be easily extended, there should be some solution which benefits all. I think we'd only need generics/buildAs if we accept that these classes are too messy to be extended, which I hope we can avoid

cristianrgreco:
@leonwilly I'm looking into this at the moment, I don't think you have the same issue as with the `AbstractStartedContainer`. Take a look at how existing modules extend `GenericContainer`. You can add whatever methods you'd like, e.g `withTrafficControlSetting`, and override the `start` method to create whatever `StartedGenericContainer` you'd like. For example:

```typescript
export class MySqlContainer extends GenericContainer {
  private database = "test";

  constructor(image = "mysql:8.0.31") {
    super(image);
  }

  public withDatabase(database: string): this {
    this.database = database;
    return this;
  }

  public override async start(): Promise<StartedMySqlContainer> {
    return new StartedMySqlContainer(await super.start(), this.database);
  }
}

export class StartedMySqlContainer extends AbstractStartedContainer {
  constructor(startedTestContainer: StartedTestContainer, private readonly database: string) {
    super(startedTestContainer);
  }

  public getDatabase(): string {
    return this.database;
  }
}

```

Issue #428, opened 2022-12-20
Authenticate using gitlab DOCKER_AUTH_CONFIG env variable


owodunni:
**Expected Behaviour**
When using gitlab CI and running `testcontainers` I would like it to use the gitlab environment variable config. It is set on the environment variable DOCKER_AUTH_CONFIG. It already needs to be setup for gitlab to be able to fetch from [private registries](https://docs.gitlab.com/ee/ci/docker/using_docker_images.html#access-an-image-from-a-private-container-registry) so being able to reuse it for `testcontainers` to authenticate seams pretty plesent. 

**Actual Behaviour**
testcontainers fail to fetch containers.

**Testcontainer Logs**
```No Docker client strategy found
    at ../../node_modules/testcontainers/dist/docker/docker-client.js:46:11
    at fulfilled (../../node_modules/testcontainers/dist/docker/docker-client.js:5:58)
```

**Steps to Reproduce**
1. In a gitlab ci runner try to fetch from a private registry

**Environment Information**
- Operating System: Docker in docker executing in node:16 container
- Docker version: unclear
- Node version: 16
- Testcontainers version: 8.13.0

owodunni:
I managed to get a simple test repository up and failing! https://gitlab.com/alexanderp2/testcontainers-dind-test

Looking at it some more I'm not confident that the fixes that I proposed on github will actually solve the issue. It might just be that the authentication needs to be configured outside of testcontainers when running in this type of nested CI environment. Ohh well, more debugging I guess.

owodunni:
I got some more feedback, 

"From a quick look it seems you’re missing DOCKER_HOST config.

If I had to guess I’d say you’re mounting docker socket from the underlying system and running directly on it and not the dind service:
 2022-12-22T05:42:13.735606282Z failed to load listeners: can't create unix socket /var/run/docker.sock: device or resource busy
You can access the host docker through /var/run/docker_host.sock on shared runners although"

I guess my test repository is not configured correctly.

cristianrgreco:
Will be supported by https://github.com/testcontainers/testcontainers-node/pull/476, thanks for raising @owodunni

PR #429, opened 2022-12-20
Enable reading docker config from environment variable (#428)

owodunni:
When using `testcontainers` in a docker in docker environment it can be a bit stainious to setup config files. GitLab alread suggests using a environment variable DOCKER_AUTH_CONFIG to configure docker.

With this fix `testcontainers` checks if the DOCKER_AUTH_CONFIG is available after trying all other means of authentication.

### DIFF
--- a/src/registry-auth-locator/config.test.ts
+++ b/src/registry-auth-locator/config.test.ts
@@ -1,0 +1,57 @@
+import { DockerConfig } from "./types";
+
+import fs from "fs";
+import { readDockerConfig } from "./config";
+
+const spyExists = jest.spyOn(fs, "existsSync");
+const spyRead = jest.spyOn(fs.promises, "readFile");
+
+describe("Config", () => {
+  const dockerConfig: DockerConfig = {
+    auths: {
+      "https://registry.example.com": {
+        email: "user@example.com",
+        username: "user",
+        password: "pass",
+      },
+    },
+  };
+
+  afterEach(() => (process.env.DOCKER_AUTH_CONFIG = undefined));
+
+  it("should return empty object if neither config nor environment variable exists", async () => {
+    spyExists.mockReturnValueOnce(false);
+
+    const config = await readDockerConfig();
+    expect(config).toEqual({});
+  });
+
+  it("should use the docker environment variable if it exists", async () => {
+    process.env.DOCKER_AUTH_CONFIG = JSON.stringify(dockerConfig);
+
+    spyExists.mockReturnValueOnce(false);
+
+    const config = await readDockerConfig();
+    expect(config).toEqual(dockerConfig);
+  });
+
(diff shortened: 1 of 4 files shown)


cristianrgreco:
Thanks @owodunni. Would you mind updating the `Authentication` section of the README with this new behaviour? https://github.com/testcontainers/testcontainers-node#authentication

src/registry-auth-locator/config.test.ts
```diff
@@ -0,0 +1,35 @@
...
+        },
+      },
+    };
+    process.env.DOCKER_AUTH_CONFIG = JSON.stringify(dockerConfig);
+
+    spy.mockReturnValue(false);
+
+    const config = await readDockerConfig();
+    expect(config).toEqual(dockerConfig);
+  });
```
cristianrgreco, line 34: I don't suppose you could add the remaining case, where the file does exist? I guess you could spy on `readFile`? I know it's not introduced by your change but it'd complete the test spec 🙂

src/registry-auth-locator/config.ts
```diff
@@ -0,0 +1,37 @@
...
+  const dockerAuthConfig = process.env.DOCKER_AUTH_CONFIG;
+  if (!dockerAuthConfig) {
+    return {};
+  }
+  return JSON.parse(dockerAuthConfig);
+};
+
+export const readDockerConfig = async (): Promise<DockerConfig> => {
+  if (!existsSync(dockerConfigFile)) {
+    return readDockerConfigFromEnv();
```
cristianrgreco, line 26: Could this be refactored to show the 3 different paths at a high level? Something like:

```javascript
export const readDockerConfig = async () => {
  if (existsSync(dockerConfigFile)) {
    return fromDockerConfigFile();
  } else if (process.env.DOCKER_AUTH_CONFIG !== undefined) {
    return fromDockerAuthConfigEnv();
  } else {
    return {};
  }
};
```

src/registry-auth-locator/config.test.ts
owodunni: Ofcourse, I didn't want to do it before I knew your thoughts on these changes in general. :)

src/registry-auth-locator/config.ts
owodunni: Check out the latest solution and see if looks better.

kiview:
The only other implementation that currently supports this is `testcontainers-dotnet`. This is the commit that added the support to .NET: https://github.com/testcontainers/testcontainers-dotnet/commit/ced98a9db6c70c188b113938decc25d12d03e6fc

cristianrgreco:
Excellent, thanks @owodunni. Just waiting on your last Slack message before merging

cristianrgreco:
Will be fixed by #476

PR #433, opened 2022-12-23
Remove path from default registry

cristianrgreco:


src/registry-auth-locator/index.ts
```diff
@@ -9,7 +9,7 @@ import { RegistryAuthLocator } from "./registry-auth-locator";
 import { log } from "../logger";
 import { AuthConfig } from "../docker/types";
 
-const DEFAULT_REGISTRY = "https://index.docker.io/v1/";
+const DEFAULT_REGISTRY = "https://index.docker.io";
```
HofmeisterAn, line 12: I think `https://index.docker.io/v1/` is correct. That's what Docker Desktop resolves to, see:

- https://github.com/testcontainers/testcontainers-dotnet/issues/666#issuecomment-1313558944
- https://github.com/testcontainers/testcontainers-java/pull/5347

What I meant by my comment in Slack is, that the GitLab example uses `https://index.docker.io` as key, but our default key is `https://index.docker.io/v1/`. Might be addressed due to `key.includes(registry)`.

cristianrgreco: @HofmeisterAn Understood, and I found that for the credential provider it didn't work in this case. We check the substring when listing the credentials, but when we go to get the credential we lookup the original registry, which isn't found

cristianrgreco:
Superseded by https://github.com/testcontainers/testcontainers-node/pull/484

PR #437, opened 2023-01-06
make unit for startup timeout more obvious

mdunisch:
there is no documentation in what unit the parameter for withStartupTimeout is. Thats why i made it more obvious buy:
- adding a comment in the readme example
- adding "Ms" to the Param-Name (so you see it in your IDE directly)

### DIFF
--- a/README.md
+++ b/README.md
@@ -793,7 +793,7 @@ const { GenericContainer } = require("testcontainers");
 
 const container = await new GenericContainer("redis")
   .withExposedPorts(6379)
-  .withStartupTimeout(120000)
+  .withStartupTimeout(120000) // wait 120s 
   .start();
 ```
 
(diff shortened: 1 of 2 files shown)


cristianrgreco:
Thanks for raising @mdunisch, would you mind also updating the variable name in the interface?

https://github.com/testcontainers/testcontainers-node/blob/08454582aee7efd74312236040dc76d6c907d3a5/src/test-container.ts#L41

mdunisch:
> Thanks for raising @mdunisch, would you mind also updating the variable name in the interface?
> 
> https://github.com/testcontainers/testcontainers-node/blob/08454582aee7efd74312236040dc76d6c907d3a5/src/test-container.ts#L41

@cristianrgreco fixed!

PR #447, opened 2023-02-09
fix: make sure log based waiting strategy respects container restarts

### LINKED ISSUE #446: Log based waiting strategy immediately fires after restarting container
**Expected Behaviour**
After restarting a container, it should wait until the log message appears again, not consider previous log lines that already have appeared.

**Actual Behaviour**
Log based waiting strategy immediately considers the container as ready after restart, as the logs are inspected from the beginning of time (i.e when the container originally got started).

**Testcontainer Logs**
```
  testcontainers DEBUG Found applicable Docker client strategy: UnixSocketStrategy +0ms
  testcontainers DEBUG Testing Docker client strategy URI: unix:///var/run/docker.sock +0ms
  testcontainers INFO  Using Docker client strategy: UnixSocketStrategy, Docker host: localhost +9ms
  testcontainers DEBUG Fetching system diagnostics +0ms
  testcontainers DEBUG Executing Docker credential provider: docker-credential-desktop +1ms
  testcontainers DEBUG No credential found for registry: "https://index.docker.io/v1/" +179ms
  testcontainers DEBUG No registry auth locator found for registry: "https://index.docker.io/v1/" +0ms
  testcontainers DEBUG Checking if image exists: frolvlad/alpine-bash:latest +1ms
  testcontainers DEBUG Not pulling image as it already exists: frolvlad/alpine-bash:latest +19ms
  testcontainers DEBUG Creating new Reaper for session: e8776e00e57e6c50a4995d53df593137 +0ms
  testcontainers DEBUG Re-using cached auth for registry https://index.docker.io/v1/ +0ms
  testcontainers DEBUG Checking if image exists: testcontainers/ryuk:0.3.2 +0ms
  testcontainers DEBUG Not pulling image as it already exists: testcontainers/ryuk:0.3.2 +6ms
  testcontainers INFO  Creating container for image: testcontainers/ryuk:0.3.2 +0ms
  testcontainers INFO  Starting container testcontainers/ryuk:0.3.2 with ID: 1ac84c6df2a56259caa9bdeffb7e8e382cc7dbea5eb4854d5eb7c4a357ca0ebf +18ms
  testcontainers DEBUG System diagnostics: {"node":{"version":"v14.21.2","architecture":"arm64","platform":"darwin"},"docker":{"serverVersion":"20.10.22","operatingSystem":"Docker Desktop","operatingSystemType":"linux","architecture":"aarch64","cpus":4,"memory":8337670144},"dockerCompose":{"version":"2.15.1"}} +1ms
  testcontainers DEBUG Waiting for container to be ready: 1ac84c6df2a56259caa9bdeffb7e8e382cc7dbea5eb4854d5eb7c4a357ca0ebf +207ms
  testcontainers DEBUG Waiting for log message "/.+ Started!/" +0ms
  testcontainers INFO  Container is ready +4ms
  testcontainers DEBUG Connecting to Reaper (attempt 1) 1ac84c6df2a56259caa9bdeffb7e8e382cc7dbea5eb4854d5eb7c4a357ca0ebf on localhost:32819 +1ms
  testcontainers DEBUG Connected to Reaper 1ac84c6df2a56259caa9bdeffb7e8e382cc7dbea5eb4854d5eb7c4a357ca0ebf +3ms
  testcontainers INFO  Creating container for image: frolvlad/alpine-bash:latest +0ms
  testcontainers INFO  Starting container frolvlad/alpine-bash:latest with ID: aa16240c50f2ac8080988d9018951ab9dafe0f6091e0bb3b2854b3ff23ded88c +17ms
  testcontainers DEBUG Waiting for container to be ready: aa16240c50f2ac8080988d9018951ab9dafe0f6091e0bb3b2854b3ff23ded88c +176ms
  testcontainers DEBUG Waiting for log message "Ready" +0ms
  testcontainers INFO  Container is ready +5s
  # Restart happened here
  testcontainers DEBUG Waiting for container to be ready: aa16240c50f2ac8080988d9018951ab9dafe0f6091e0bb3b2854b3ff23ded88c +280ms
  testcontainers DEBUG Waiting for log message "Ready" +0ms
  testcontainers INFO  Container is ready +4ms
```

**Steps to Reproduce**
Including a simple test-case that can be run in the project, with the current code it fails the second assertion (as the time between just before the restart and after the restart is not enough to have generated the second log line)
```ts
it('correctly waits for logs after restart', async () => {
  const container = new GenericContainer('frolvlad/alpine-bash')
    .withEntrypoint(['/bin/sh', '-c', 'sleep 5; echo "Ready"; tail -f /dev/null'])
    .withWaitStrategy(Wait.forLogMessage('Ready'));

  let date = new Date();
  const started = await container.start();
  expect(new Date().getTime() - date.getTime()).toBeGreaterThanOrEqual(5_000);

  date = new Date();
  await started.restart();
  expect(new Date().getTime() - date.getTime()).toBeGreaterThanOrEqual(5_000);
}, 20_000);
```

**Environment Information**
- Operating System: macOS Ventura 13.1
- Docker Version: 20.10.22
- Node version: 16
- Testcontainers version: 9.1.2

henrinormak:
Fixes #446

### DIFF
--- a/src/generic-container/generic-container.test.ts
+++ b/src/generic-container/generic-container.test.ts
@@ -71,6 +71,21 @@ describe("GenericContainer", () => {
     await container.stop();
   });
 
+  it("should wait for a new log after restart", async () => {
+    let date = new Date();
+    const container = await new GenericContainer("cristianrgreco/testcontainer:1.1.14")
+      .withCommand(["/bin/sh", "-c", 'sleep 2; echo "Ready"'])
+      .withWaitStrategy(Wait.forLogMessage("Ready"))
+      .start();
+
+    expect(new Date().getTime() - date.getTime()).toBeGreaterThanOrEqual(2_000);
+    date = new Date();
+    await container.restart();
+    expect(new Date().getTime() - date.getTime()).toBeGreaterThanOrEqual(2_000);
+
+    await container.stop();
+  });
+
   it("should wait for health check", async () => {
     const context = path.resolve(fixtures, "docker-with-health-check");
     const customGenericContainer = await GenericContainer.fromDockerfile(context).build();
--- a/src/generic-container/generic-container.test.ts
+++ b/src/generic-container/generic-container.test.ts
@@ -72,16 +72,20 @@ describe("GenericContainer", () => {
   });
 
   it("should wait for a new log after restart", async () => {
-    let date = new Date();
+    const start = new Date();
     const container = await new GenericContainer("cristianrgreco/testcontainer:1.1.14")
       .withCommand(["/bin/sh", "-c", 'sleep 2; echo "Ready"'])
       .withWaitStrategy(Wait.forLogMessage("Ready"))
       .start();
 
-    expect(new Date().getTime() - date.getTime()).toBeGreaterThanOrEqual(2_000);
-    date = new Date();
+    expect(new Date().getTime() - start.getTime()).toBeGreaterThanOrEqual(2_000);
+
+    // Add a short delay, as otherwise we'll still be too fast and restart the container
+    // on the same second as it previously got ready (which will mess with the log detection)
+    await new Promise((resolve) => setTimeout(resolve, 1_000));
     await container.restart();
-    expect(new Date().getTime() - date.getTime()).toBeGreaterThanOrEqual(2_000);
+
+    expect(new Date().getTime() - start.getTime()).toBeGreaterThanOrEqual(4_000);
 
     await container.stop();
   });
--- a/src/generic-container/generic-container.test.ts
+++ b/src/generic-container/generic-container.test.ts
@@ -79,12 +79,7 @@ describe("GenericContainer", () => {
       .start();
 
     expect(new Date().getTime() - start.getTime()).toBeGreaterThanOrEqual(2_000);
-
-    // Add a short delay, as otherwise we'll still be too fast and restart the container
-    // on the same second as it previously got ready (which will mess with the log detection)
-    await new Promise((resolve) => setTimeout(resolve, 1_000));
     await container.restart();
-
     expect(new Date().getTime() - start.getTime()).toBeGreaterThanOrEqual(4_000);
 
     await container.stop();
--- a/src/wait-strategy.ts
+++ b/src/wait-strategy.ts
@@ -74,7 +74,8 @@ export class LogWaitStrategy extends AbstractWaitStrategy {
 
   public async waitUntilReady(container: Dockerode.Container): Promise<void> {
     log.debug(`Waiting for log message "${this.message}"`);
-    const stream = await containerLogs(container);
+    const inspect = await inspectContainer(container);
+    const stream = await containerLogs(container, { since: inspect.state.startedAt.getTime() / 1000 });
 
     return new Promise((resolve, reject) => {
       const startupTimeout = this.startupTimeout;
--- a/src/wait-strategy.ts
+++ b/src/wait-strategy.ts
@@ -75,7 +75,8 @@ export class LogWaitStrategy extends AbstractWaitStrategy {
   public async waitUntilReady(container: Dockerode.Container): Promise<void> {
     log.debug(`Waiting for log message "${this.message}"`);
     const inspect = await inspectContainer(container);
-    const stream = await containerLogs(container, { since: Math.floor(inspect.state.startedAt.getTime() / 1000) });
+    const startedAt = inspect.state.startedAt.getTime();
+    const stream = await containerLogs(container, { since: Math.floor(startedAt / 1000), timestamps: true });
 
     return new Promise((resolve, reject) => {
       const startupTimeout = this.startupTimeout;
@@ -86,6 +87,12 @@ export class LogWaitStrategy extends AbstractWaitStrategy {
       }, startupTimeout);
 
       const comparisonFn: (line: string) => boolean = (line: string) => {
+        const timestamp = new Date(line.split(" ")[0]).getTime();
+        if (!isNaN(timestamp) && timestamp < startedAt) {
+          // We should not be getting this log line, ignore it
+          return false;
+        }
+
         if (this.message instanceof RegExp) {
           return this.message.test(line);
         } else {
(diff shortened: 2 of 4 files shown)


cristianrgreco:
Thanks @henrinormak, I was having a look to see how I'd implement it and I would've done the exact same 🙂

cristianrgreco:
@henrinormak seems this change has broken the tests

cristianrgreco:
@henrinormak seems some tests are failing

henrinormak:
Will take a look

henrinormak:
So at least one of those I managed to fix, the others I'm not 100% sure about, as they are not reproducing locally (the startup strategy one and the reaper one).

henrinormak:
Seems like the timing is quite finicky, I think the flooring of the time is generally a good idea, as that removes some weirdness that Docker(ode) might have with fetching logs after some timestamp. However, in the tests, the restart cycle is so fast, this flooring becomes a problem, as it still manages to read the log line from the previous start cycle.

Not sure whether this should be solved in code or rather should it be addressed in the test - i.e have some delay between initial container start and when we restart it (even just 1s is enough).

src/generic-container/generic-container.test.ts
```diff
@@ -71,6 +71,25 @@ describe("GenericContainer", () => {
...
+    const container = await new GenericContainer("cristianrgreco/testcontainer:1.1.14")
+      .withCommand(["/bin/sh", "-c", 'sleep 2; echo "Ready"'])
+      .withWaitStrategy(Wait.forLogMessage("Ready"))
+      .start();
+
+    expect(new Date().getTime() - start.getTime()).toBeGreaterThanOrEqual(2_000);
+
+    // Add a short delay, as otherwise we'll still be too fast and restart the container
+    // on the same second as it previously got ready (which will mess with the log detection)
+    await new Promise((resolve) => setTimeout(resolve, 1_000));
```
cristianrgreco, line 85: Why won't the same problem occur in production cases?

henrinormak: Because it would assume someone restarts the container on the same second that it was declared ready by the log waiting strategy - this seems highly unlikely, no? I could also manipulate the restarting to force a delay that way. I can't think of any other solution around this, any thoughts?

henrinormak: I refactored to not rely on this type of behaviour, it does come with a bit of performance overhead in the log waiting strategy, but perhaps that is not a big deal? Basically, it validates that the log lines it gets are actually after the expected start timestamp.

src/wait-strategy.ts
```diff
@@ -74,7 +74,9 @@ export class LogWaitStrategy extends AbstractWaitStrategy {
 
   public async waitUntilReady(container: Dockerode.Container): Promise<void> {
     log.debug(`Waiting for log message "${this.message}"`);
-    const stream = await containerLogs(container);
+    const inspect = await inspectContainer(container);
+    const startedAt = inspect.state.startedAt.getTime();
+    const stream = await containerLogs(container, { since: Math.floor(startedAt / 1000), timestamps: true });
```
cristianrgreco, line 79: Is the rounding still necessary?

src/wait-strategy.ts
```diff
@@ -85,6 +87,12 @@ export class LogWaitStrategy extends AbstractWaitStrategy {
       }, startupTimeout);
 
       const comparisonFn: (line: string) => boolean = (line: string) => {
+        const timestamp = new Date(line.split(" ")[0]).getTime();
+        if (!isNaN(timestamp) && timestamp < startedAt) {
```
cristianrgreco, line 91: Should this be OR?

src/generic-container/generic-container.test.ts
cristianrgreco: This looks great, I don't think there's any real performance impact here and it lets us be 100% accurate

src/wait-strategy.ts
henrinormak: Unfortunately, yes, as without it, we don't get back all logs that might be necessary (i.e one of the startup strategy tests fails)

src/wait-strategy.ts
henrinormak: I went with a conservative approach and if we don't get a valid timestamp at the start of the line, we do still test the line. I thought that this was better than perhaps accidentally never declaring a container as ready.

henrinormak:
Adjusted the approach one more time, I think this one is actually the most suitable one, in essence, instead of taking "logs after our container started", we now go with "discard logs before the container last finished".

I'm hoping that the tests will like this approach, it does not do any rounding any more (as semantically it is not possible to log something earlier than the last time the container stopped). The only hairy part is the discarding of the finished timestamp if it is in year 1 - I could also make it so that the inspect always returns a date, and the log waiting keeps that in mind?

henrinormak:
This is slowly making me lose my sanity, the approach that I added last night is quite obfuscated, but it might work - the check for whether the container is running or not is there for a simple reason, if it is not running, the finished timestamp is likely so new that no logs will have been sent after that, which would ruin the flow of containers that immediately exit. As such, it makes sense for those to consider all logs.

Ideally the container would give us a more concrete way of limiting the logs to some specific "run", but I could not think of any way to achieve that - at least not a way that would behave transparently from the user perspective (we could always log something as a sentinel that a restart was triggered, but that would show up in the logs, visible to the user).

src/wait-strategy.ts
```diff
@@ -74,7 +74,17 @@ export class LogWaitStrategy extends AbstractWaitStrategy {
...
+
+    // We should not take logs prior to our last finish time into account
+    // Otherwise we risk reporting the container as ready based on old logs
+    // This does not apply to containers that are currently not running (i.e already exited)
+    const finishedAt = inspect.state.finishedAt?.getTime();
+    const logsOptions =
+      finishedAt !== undefined && inspect.state.status === "running"
+        ? { since: finishedAt / 1000, timestamps: true }
+        : undefined;
+    const stream = await containerLogs(container, logsOptions);
```
cristianrgreco, line 87: I wonder if we could refactor this a bit, something like:

```typescript
const inspect = await inspectContainer(container);

const logOptions = this.hasContainerRestarted(inspect) 
  ? containerRestartedLogOptions(inspect) 
  : undefined
```

WDYT?

henrinormak: Sounds good, will do

cristianrgreco (approved):

cristianrgreco:
Thanks @henrinormak, I am sorry you lost your sanity but I appreciate the contribution 😄

cristianrgreco:

PR #449, opened 2023-02-13
fix: make sure to not create more than one reusable container

henrinormak:
Sorry for spamming you with PRs recently, but found one more potential issue.

Namely, if reusable containers are triggered in parallel (for example via jest's `it.concurrent`), it can lead to multiple of the reusable containers being spawned. This is generally not an issue, but in test suites that have dozens or more test cases, it can quite quickly rack up a large number of running containers, ultimately leading to resource constraints.

This proposed approach is hopefully simple enough - checking whether the container already exists or creating it are two separate async processes, but as they are dependent, we should do them in a locked fashion. I added a really simple async lock (no timeouts, no fancy options for re-entry etc.), which can be used to make sure each reusable container hash is checked and created as an atomic set of steps (at least within one node process).

Let me know if I need to add some further functionality or if I missed anything.


henrinormak:
@cristianrgreco sorry to bother you, but any way this could get reviewed, it is blocking us from adopting testcontainers - would prefer not to fork the library as it is otherwise an excellent fit!

cristianrgreco:
Hi @henrinormak, not a bother, I am grateful for your contributions! The PR looks great, would it be possible to use an existing lib for async-lock? https://www.npmjs.com/package/async-lock. Seems to have a similar API and is less to maintain. Other than that LGTM

henrinormak:
Sure, I originally used that, but thought that it was actually the opposite in terms of maintenance burden (i.e having to update it and what not). I'll swap back to that one.

cristianrgreco:

PR #450, opened 2023-02-13
fix: correctly evaluate predicates in HttpWaitStrategy

abendi:
`this.predicates.every((predicate) => predicate(response))` in `HttpWaitStrategy` always resolved to `true`, because Promise objects were tested, which are always truthy. Fixing it by looping over predicates with async `for..of`.

Not too sure about the timeout tests, they add a bit of time.


cristianrgreco:
Thanks for raising @abendi

cristianrgreco:

PR #466, opened 2023-02-20
Make sure container logs are output after container restart

### LINKED ISSUE #465: Container logs are no longer output after restart
**Expected Behaviour**
Container logs are output after a `container.restart()` when using `DEBUG=testcontainers:containers`

**Actual Behaviour**
Container logs are only output until a container is restarted

**Testcontainer Logs**
N/A

**Steps to Reproduce**
1. With `DEBUG=testcontainers:containers`
2. Start a container, logs are output
3. Restart the container via `container.restart()` API
4. No container logs are output after restart

**Environment Information**
- Operating System: Mac
- Docker Version: 20.10.21
- Node version: v18.13.0
- Testcontainers version: 9.1.2

justinwyer:
Expanding on the work done in #447, this fixes #465


src/generic-container/generic-container.test.ts
```diff
@@ -368,6 +371,21 @@ describe("GenericContainer", () => {
     await container.stop();
   });
 
+  it("should stream logs from a running container after restart", async () => {
+    const mock = jest.mocked(containerLog);
```
cristianrgreco, line 375: There is already a `FakeLogger` implementation in `logger.ts`, is it possible to use that?

cristianrgreco:
Thanks for raising this PR @justinwyer!

src/generic-container/generic-container.test.ts
justinwyer: I wanted to use it but could not see a way mock it out or replace it at test time, if you have any ideas I am happy to implement them.

cristianrgreco: Yes actually it's not straight forward. There's a mix of fakes+DI and jest mocking, it would be good to pick one at some point, but it is outside the scope of this PR

src/docker/functions/container/restart-container.ts
```diff
@@ -1,17 +1,44 @@
...
+
 export const restartContainer = async (
   container: Dockerode.Container,
   options: RestartContainerOptions
 ): Promise<void> => {
   try {
     await container.restart({ t: options.timeout / 1000 });
-  } catch (err) {
+
+    const inspect = await inspectContainer(container);
```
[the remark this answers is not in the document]
justinwyer: great idea, let me give that a try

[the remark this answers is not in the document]
justinwyer: I moved `containerRestartedLogOptions` & `hasContainerRestarted` into `inspect-container.js` not sure if this is the best place

src/generic-container/generic-container.test.ts
justinwyer: I feel that a good approach might be providing a `Logger` as an parameter to `GenericContainer` which can be propagated elsewhere, it could have a default value of the `logger.containerLog` but then be overridden in tests (or by users who want to override it), but this would be best as a seperate PR. Would be happy to take it further if you think the approach is sound.

cristianrgreco: It's a good idea, I don't know at the moment what's the right thing to do. I quite like that right now you can call functions like `containerLogs(container)` from anywhere without worrying about dependencies, and so I don't love the idea of propagating out its dependencies. 

In fact, I wonder if in this case our level of testing is too low, we're testing the actual `containerLogs` function here, instead of testing that we call it when a container is restarted. Perhaps then we should just spy on `containerLogs` and ensure it's called when the container is restarted (instead of mocking the logger, checking the actual logs, etc)?

justinwyer: Let me spend a little more time trying to get a spy working, it did not when I initially tried.

PR #486, opened 2023-03-08
Add support for rootless Docker

### LINKED ISSUE #434: No Docker client strategy found on Docker Desktop for linux
The linux version of this issue: https://github.com/testcontainers/testcontainers-node/issues/407#issuecomment-1367467107

It looks like the Docker devs realized the missing socket broke "all the things" on mac, so the [socket systemlink has been added back in on docker desktop for Mac](https://docs.docker.com/desktop/release-notes/#4140), but sadly not for linux.

image

Docker tooling like this will need a solution for handling the case where only the userspace socket exists and not `/var/run/docker.sock`, otherwise there will be a lot of issues like this one where mac users are ok but linux users are still broken. Does anyone know if there is a standard way to get the socket location, maybe from the active docker context?

cristianrgreco:
Fixes #434

Please ignore additions/removals, there is a change to the lockfile.


kiview (approved):

kiview:
In `tc-java`, we also have a CI job for rootless Docker, should be duplicate it here as well?
https://github.com/testcontainers/testcontainers-java/blob/main/.github/workflows/ci-rootless.yml

cristianrgreco:
Looking into adding a CI job. Rootless Docker does not seem to work on Ubuntu 22.04, appears related to https://github.com/moby/moby/issues/45014

PR #520, opened 2023-03-31
fix: directConnection flag should be present

ThallesP:
Hey! When not using the `directConnection=true`, mongoose (or nestjs mongoose) will not be able to connect.

Also, [mongosh](https://www.mongodb.com/docs/mongodb-shell/connect/#option-2--specify-members-in-connection-string) seems to be adding that automatically.

### DIFF
--- a/src/modules/mongodb/mongodb-container.ts
+++ b/src/modules/mongodb/mongodb-container.ts
@@ -70,6 +70,6 @@ export class StartedMongoDBContainer extends AbstractStartedContainer {
   }
 
   public getConnectionString(): string {
-    return `mongodb://${this.getHost()}:${this.getMappedPort(MONGODB_PORT)}`;
+    return `mongodb://${this.getHost()}:${this.getMappedPort(MONGODB_PORT)}?directConnection=true`;
   }
 }


cristianrgreco:
Hi @ThallesP, is it not a case of adding `directConnect` if/when needed? I'm aware that a) this is how it's implemented in tc-java (without `directConnect`), and b) this would be breaking change

ThallesP:
> Hi @ThallesP, is it not a case of adding `directConnect` if/when needed? I'm aware that a) this is how it's implemented in tc-java (without `directConnect`), and b) this would be breaking change

I don't know if that would be a breaking change as `testcontainers` creates the mongodb container and I connect to that using it's own `connection string`, if `connection string` doesn't work then it's a bug in my opinion.

cristianrgreco:
> I don't know if that would be a breaking change as testcontainers creates the mongodb container and I connect to that using it's own connection string which should be valid, if connection string doesn't work then it's a bug in my opinion.

Perhaps other users are already appending the `directConnect` query parameter, this change would break those cases.

ThallesP:
> > I don't know if that would be a breaking change as testcontainers creates the mongodb container and I connect to that using it's own connection string which should be valid, if connection string doesn't work then it's a bug in my opinion.
> 
> Perhaps other users are already appending the `directConnect` query parameter, this change would break those cases.

Yeah, it makes sense now.
Maybe a property in `getConnectionString` that adds the `directConnect` would be better?

ThallesP:
> > Maybe a property in getConnectionString that adds the directConnect would be better?
> 
> Maybe. I don't know if it would be used by many people though

If it's in the documentation, it would be used and also would save me a lot of time when debugging my code for finding that.

Maybe I should make a PR that changes only the documentation, makes more sense, what do you think? @cristianrgreco

PR #527, opened 2023-04-08
Added a compatible image for mysql for m1

ramigoldvarg:


cristianrgreco:
Hi @ramigoldvarg, the existing MySQL module already works on M1 macs

ramigoldvarg:
My bad, I was working with outdated version

PR #530, opened 2023-04-13
Add support for docker-compose pull policy

### LINKED ISSUE #529: add `--pull always` to docker compose command
It would be very useful to add a Pull Policy like `--pull always` to the docker compose environment. There is a way to add this parameter to single containers with 
```
const { GenericContainer, AlwaysPullPolicy } = require("testcontainers");

const container = await GenericContainer
  .withPullPolicy(new AlwaysPullPolicy())
```
but I can't see any way to add this to a Docker Compose Environment instance. The argument `--pull` exists in the [Docker API](https://docs.docker.com/engine/reference/commandline/compose_up/#options).

**Expected Behaviour**
there is a way to adjust the --pull parameter for the docker compose up command.

**Actual Behaviour**
there is no way to adjust the --pull parameter for the docker compose up command.

cristianrgreco:
Fixes #529

### DIFF
--- a/docs/features/compose.md
+++ b/docs/features/compose.md
@@ -41,6 +41,16 @@ const environment = await new DockerComposeEnvironment(composeFilePath, composeF
   .up();
 ```
 
+### With a pull policy
+
+```javascript
+const { DockerComposeEnvironment, AlwaysPullPolicy } = require("testcontainers");
+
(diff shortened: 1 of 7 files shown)


gibso:
Amazing, thank you!

PR #545, opened 2023-04-26
Update supported container runtimes doc to add Podman/MacOS

cristianrgreco:
See Netlify preview

### DIFF
--- a/docs/supported-container-runtimes.md
+++ b/docs/supported-container-runtimes.md
@@ -10,6 +10,20 @@ Works out of the box.
 export DOCKER_HOST=unix://${XDG_RUNTIME_DIR}/podman/podman.sock
 ```
 
+### Known issues
+
+**Resource reaper does not work on MacOS**
+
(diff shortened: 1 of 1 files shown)


docs/supported-container-runtimes.md
```diff
@@ -10,6 +10,20 @@ Works out of the box.
 export DOCKER_HOST=unix://${XDG_RUNTIME_DIR}/podman/podman.sock
 ```
 
+### Known issues
+
+**Resource reaper does not work on MacOS**
+
+The resource reaper is a process which runs in the background and cleans up containers and images after they are no longer needed. It is not possible to run the resource reaper on MacOS with Podman due to permission limitations, it fails with "Operation not supported" errors.
```
HofmeisterAn, line 17: Should we mention the rootful workaround?

cristianrgreco: I still had network related issues with the rootful workaround. Is it not the case in .NET?

cristianrgreco: I've just double checked, I do not see network issues, perhaps yesterday I confused something with the ryuk privileged flag. Disabling ryuk works for both rootful and rootless

HofmeisterAn (approved):

PR #553, opened 2023-04-29
Adding mysql connection uri utility function

jakeleventhal:


cristianrgreco:
Hi @jakeleventhal, looks like some tests are failing, are you able to have a look?

jakeleventhal:
I don't understand why the test is failing? expected and received look identical to me

cristianrgreco:
@jakeleventhal The test output says:

```
Expected: "***localhost:32870/testDB"
Received: "***localhost:32870/testDB"
```

Indeed they are identical. I've run the test locally on Docker (as that runtime failed) and see the following:

```
mysql://root:testPassword@localhost:32771/testDB
mysql://test:testPassword@localhost:32771/testDB
```

Looks like CI replaced the password and everything before with `***`. Looks like the user is invalid.

jakeleventhal:
@cristianrgreco I believe the push i just made ought to fix it. Having trouble getting tests running locally but this should be good to go I think

cristianrgreco:
@jakeleventhal thanks for the contribution!

PR #558, opened 2023-05-03
DEBUG mode for single container by name

AlexPtnn:
Hello!

Made DEBUG mode for logs for single container by its name.
It is convenient if I need logs only from one container instead of logs from all containers.
Also name of container displaying before log line, so it is easier to identify container by its name instead of id.

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -10,6 +10,7 @@ All possible environment variable configurations for Testcontainers are found he
 | DEBUG    | testcontainers            | Enable testcontainers logs |
 | DEBUG    | testcontainers:containers | Enable container logs      |
 | DEBUG    | testcontainers:exec       | Enable container exec logs |
+| DEBUG    | testcontainers:container:${container-name} | Enable logs for single container by name |
 
 Note that you can enable multiple loggers, e.g: `DEBUG=testcontainers,testcontainers:exec`.
 
--- a/src/generic-container/generic-container.ts
+++ b/src/generic-container/generic-container.ts
@@ -206,9 +206,9 @@ export class GenericContainer implements TestContainer {
     }
 
     if (containerLog.enabled()) {
-      await this.logContainer(container, containerLog, containerName);
+      await this.logContainer(container, containerLog);
     } else if (currentContainerLog?.enabled()) {
-      await this.logContainer(container, currentContainerLog, containerName);
+      await this.logContainer(container, currentContainerLog);
     }
 
     if (this.containerStarting) {
@@ -235,12 +235,10 @@ export class GenericContainer implements TestContainer {
     return startedContainer;
   }
 
-  private async logContainer(container: Dockerode.Container, logger: Logger, name?: string) {
-    const containerName = name ? `${name}, ` : "";
-
+  private async logContainer(container: Dockerode.Container, logger: Logger) {
     (await containerLogs(container))
-      .on("data", (data) => logger.trace(`${containerName}${container.id}: ${data.trim()}`))
-      .on("err", (data) => logger.error(`${containerName}${container.id}: ${data.trim()}`));
+      .on("data", (data) => logger.trace(`${container.id}: ${data.trim()}`))
+      .on("err", (data) => logger.error(`${container.id}: ${data.trim()}`));
   }
 
   /**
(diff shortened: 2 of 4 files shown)


AlexPtnn:
Do I need to add this logic for [docker-compose](https://github.com/testcontainers/testcontainers-node/blob/main/src/docker-compose-environment/docker-compose-environment.ts#L145) and [started-container](https://github.com/testcontainers/testcontainers-node/blob/main/src/generic-container/started-generic-container.ts#L56)?

I'm open to add some other tests if you think it's important

cristianrgreco:
Hi @AlexPtnn

We generally [discourage using container names](https://node.testcontainers.org/features/containers/#with-a-name), so I don't think it makes sense to add logging capabilities around it.

Is there a reason you cannot use the API to get logs for a given container?

```ts
(await container.logs()).on("data", line => ...);
```

AlexPtnn:
@cristianrgreco logs in your example available only after successful start of container. So I cann't debug starting process of single container if it failures. Reading logs of all containers is very difficult.

cristianrgreco:
> logs in your example available only after successful start of container. So I cann't debug starting process of single container if it failures.

This is a good point. 

Technically, the `container.logs()` method _already_ returns logs which include _before_ the container is ready, but the method is only available on the `StartedTestContainer`, so if the container fails to start, you cannot get logs other than via the DEBUG logger, which is messy.

I see in testcontainers-java, a container can be created `withLogConsumer(...)`, perhaps something like this would help? The result would be something like:

```ts
const container = new GenericContainer(...)
  .withLogConsumer((stdout, stderr) => ...)
```

What do you think?

AlexPtnn:
I think it will be better to have as less options as we can.
With your solution we will have 3 different options to access logs: DEBUG, containers.logs, logConsumer.
It may cause a confusion.

Can we run `container.logs()` before StartedTestContainer available?
But I also need option to debug my specific container without changes in code (in CI, for example). It will be easier by passing env variable.

cristianrgreco:
> Can we run container.logs() before StartedTestContainer available?

Yes!

> But I also need option to debug my specific container without changes in code (in CI, for example). It will be easier by passing env variable.

I understand that the DEBUG output is verbose, but if you need to debug a CI run, surely that's helpful? 

Is the issue instead that it's hard to find the container logs based on the container ID? Maybe we could change the logs to include the container name? For example before:

```
testcontainers:containers TRACE 34255903bfa44153063ba0b507ac12f2cb3e1fd61c39559fa65737c1a0cc390d: Listening on port 8080 +473ms
```

And after:

```
testcontainers:containers TRACE e24bed940b175bb1387e83fe4ff0c4789a25e7e26e178a764d9be328ebc33bfd (/customName): Listening on port 8080 +908ms
``` 

Would that help?

AlexPtnn:
> if you need to debug a CI run, surely that's helpful?

Yes, it will be helpful. Some problems appear only on CI, so its hard to find logs for specific container among others.
Also some CI environments have limitations of logs length, so if we enable logs of all containers it will fastly overflow logs limitations.

> it's hard to find the container logs based on the container ID?

It will be nice to have container name in log. But my previous point is still important and solve other problems (easy to work with, log length limitations).

cristianrgreco:
Thanks for the feedback @AlexPtnn. However as I mentioned using container names is not recommended, for this reason I do not think it's a good idea to add logging capabilities centred around it. I think this is one of those cases where a library shouldn't try to solve everyone's needs.

I am working on a PR which will help address a lot of your concerns, in a way beneficial to everyone: https://github.com/testcontainers/testcontainers-node/pull/559

1. Include container ID in all logs where applicable (I've chosen container ID instead of container name because when following recommendations they are both random, and container name relies on having inspected the container).
2. Added a new `withLogConsumer` method to the `TestContainer`. I started to implement a `logs` method as previously discussed, but the usage wasn't fluent, for example:

```ts
const container = new GenericContainer(...);
const stream = container.logs().on(...);
const startedContainer = await container.start();
```

Is more cumbersome than:

```ts
const startedContainer = await new GenericContainer(...)
  .withLogConsumer(stream => ...)
  .start()
```

3. Created more loggers to separate output, for example build stream and pull stream output is now opt-in. This was previously trace output. This will help with log limits by default.

AlexPtnn:
Added some comments on your PR. Closing my PR.

PR #559, opened 2023-05-05
Add withLogConsumer + general logging improvements

cristianrgreco:


src/logger.ts
```diff
@@ -2,79 +2,56 @@
...
-  public enabled(): boolean {
-    return true;
+  public trace(message: Message, options?: Options): void {
+    this.logger(`${this.showLevel ? "[TRAC] " : ""}${this.renderOptions(options)}${message}`);
   }
 
-  public trace(message: Message): void {
-    this.traceLogs.push(message);
+  public debug(message: Message, options?: Options): void {
+    this.logger(`${this.showLevel ? "[DEBU] " : ""}${this.renderOptions(options)}${message}`);
```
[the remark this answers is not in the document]
AlexPtnn: DEBUG isntead of DEBU

src/logger.ts
```diff
@@ -2,79 +2,57 @@ import debug, { IDebugger } from "debug";
...
 
-  public error(message: Message): void {
-    this.errorLogs.push(message);
+  private renderOptions(options?: Options): string {
+    let str = "";
+    if (options?.containerId) {
+      str += `[${options.containerId.substring(0, 12)}] `;
+    }
+    if (options?.imageName) {
+      str += `[${options.imageName}] `;
```
AlexPtnn, line 47: Can you set here name of container? And if it is not exists use imageName? Name will be shorter than imageName in output.

src/generic-container/generic-container.ts
```diff
@@ -203,10 +206,18 @@ export class GenericContainer implements TestContainer {
...
+    if (containerLog.enabled() || this.logConsumer !== undefined) {
+      const logStream = await containerLogs(container);
+
+      if (this.logConsumer !== undefined) {
+        this.logConsumer(logStream);
+      }
+
+      if (containerLog.enabled()) {
+        logStream
+          .on("data", (data) => containerLog.trace(data.trim(), { containerId: container.id }))
```
AlexPtnn, line 218: Will be nice to have containerName/imageName here (one of, name > image), as we discussed in my PR

src/logger.ts
cristianrgreco: This field is only set when pulling an image, at which point there isn't a container

src/generic-container/generic-container.ts
cristianrgreco: Actually see this comment on your PR where I explained why not to use the container name: https://github.com/testcontainers/testcontainers-node/pull/558#issuecomment-1536068467

PR #568, opened 2023-05-11
Fix rootless socket location for MacOS on Docker Desktop

cristianrgreco:


src/docker/remote-docker-unix-socket-path.ts
```diff
@@ -1,18 +1,21 @@
...
+  if (dockerClient.allowUserOverrides) {
     if (env["TESTCONTAINERS_DOCKER_SOCKET_OVERRIDE"] !== undefined) {
       return env["TESTCONTAINERS_DOCKER_SOCKET_OVERRIDE"];
     }
   }
 
   let socketPath: string;
-  if (uri.startsWith("unix://")) {
-    socketPath = uri.replace("unix://", "");
+  if (dockerClient.info.dockerInfo.operatingSystem === "Docker Desktop") {
```
HofmeisterAn, line 15: I think we should not just check Docker Desktop. The socket path (`/var/run/docker.sock`) is also applicable to Colima and Podman Desktop (on macOS; running inside a VM). We do not need to address this in the current pull request.

HofmeisterAn (approved):

src/docker/remote-docker-unix-socket-path.ts
cristianrgreco: Very good point. I will address this in a separate PR :+1:

PR #598, opened 2023-07-10
Add HiveMQ module

h2xd:
## WHAT

- Add HiveMQ module similar to the as seen implementation in the [Java based Test-Containers repository](https://github.com/testcontainers/testcontainers-java/tree/main/modules/hivemq/src/main/java/org/testcontainers/hivemq)
- Add HiveMQ module page
  - Add links to documentation and community forums of HiveMQ
  - Add note that Enterprise Edition will follow

## WHY

I'm working on multiple JavaScript / TypeScript based MQTT solutions at HiveMQ and the setup for testing is a burden. Then I came across [Test Containers for Java](https://github.com/testcontainers/testcontainers-java) and wanted to have similar handling in Nodejs. This is where I found out that luckily a Nodejs Test Containers community does also exists. 🤩

## HOW

- Adapt the Java code to Nodejs
- Checkout modules that already exist in the testcontainer-node repository
- Adapt codebase style to create the HiveMQ testcontainer-node module

## Additional Notes
- I'm working towards Enterprise Image support similar to [the Java implementation](https://github.com/testcontainers/testcontainers-java/blob/main/modules/hivemq/src/main/java/org/testcontainers/hivemq/HiveMQContainer.java#L36). This will take some time.


cristianrgreco:
@h2xd This is an excellent contribution, thank you! 

You'll need to export the new classes from the index as well, so users can `import { HiveMQContainer, StartedHiveMQContainer } from "testcontainers"`. Will check the rest of the PR now

src/modules/hivemq/hivemq-container.ts
```diff
@@ -0,0 +1,56 @@
...
+
+export enum HiveMQDockerImageName {
+  COMMUNITY_EDITION = "hivemq/hivemq-ce",
+}
+
+export class HiveMQContainer extends GenericContainer {
+  private temporaryFileSystemOptions = new Map<string, string>();
+
+  constructor() {
+    super(HiveMQDockerImageName.COMMUNITY_EDITION);
```
cristianrgreco, line 20: I would add a default image to the constructor like here. In this way we can drop the enum: https://github.com/testcontainers/testcontainers-node/blob/9f0b3afb6e361a0e87c9873e7417cb9d8f5058ef/src/modules/postgresql/postgresql-container.ts#L13

package.json
```diff
@@ -46,6 +46,7 @@
     "docker-compose": "^0.24.1",
     "dockerode": "^3.3.5",
     "get-port": "^5.1.1",
+    "mqtt": "^4.3.7",
```
cristianrgreco, line 49: This should be a dev dependency, so it is not there for all testcontainers users. In this way also users can choose whichever mqtt library they like (maybe not useful in this case, but may be for other vendors).

src/modules/hivemq/hivemq-container.ts
```diff
@@ -0,0 +1,56 @@
...
+import { StartedTestContainer } from "../../test-container";
+import { AbstractStartedContainer } from "../abstract-started-container";
+import { Wait } from "../../wait-strategy/wait";
+
+const START_LOG_MESSAGE_REGEX = /(.*)Started HiveMQ in(.*)/i;
+const HIVEMQ_BASE_PATH = "/opt/hivemq";
+const MQTT_PORT = 1883;
+
+export enum HiveMQDockerImageName {
+  COMMUNITY_EDITION = "hivemq/hivemq-ce",
```
cristianrgreco, line 13: I would pin a version here, else one day this module will break for users.

package.json
h2xd: Makes total sense, will move it in a second!

src/modules/hivemq/hivemq-container.test.ts
```diff
@@ -0,0 +1,28 @@
...
+    const testMqttClient = mqtt.connect(hiveMQContainer.getConnectionString());
+
+    testMqttClient.on("message", (topic, message) => {
+      expect(message.toString()).toEqual("Test Message");
+
+      testMqttClient.end();
+    });
+
+    testMqttClient.on("connect", () => {
+      testMqttClient.subscribe("test", function (error) {
```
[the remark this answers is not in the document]
h2xd: Arrow functions should not be a problem. Promises might be a bit harder. I can refactor it into something more functional like that. WDYT?

```ts
testMqttClient.on("connect", handleOnConnect)

function handleOnConnect() {
  testMqttClient.subscribe("test", handleSubscribeError)
}

function handleSubscribeError(error: MqttErrorType) {
  // ... error implementation
}
```

src/modules/hivemq/hivemq-container.ts
h2xd: Truly indeed. I check the repo to see how the user can also provide a custom image. Since their use case might also only work with a previous version.

src/modules/hivemq/hivemq-container.test.ts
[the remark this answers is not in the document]
h2xd: Ah I see what you mean. Arrow function in terms on consistency. 💡

src/modules/hivemq/hivemq-container.ts
h2xd: fixed version 👍

package.json
h2xd: Moved to `devDependencies`

src/modules/hivemq/hivemq-container.ts
h2xd: Removed `enum`

src/modules/hivemq/hivemq-container.test.ts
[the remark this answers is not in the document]
h2xd: Added arrow function

src/modules/hivemq/hivemq-container.test.ts
```diff
@@ -0,0 +1,28 @@
...
+describe("HiveMQContainer", () => {
+  jest.setTimeout(240_000);
+
+  // connect {
+  it("should connect to HiveMQ Community Edition via MQTT.js", async () => {
+    const hiveMQContainer = await new HiveMQContainer().start();
+
+    const testMqttClient = mqtt.connect(hiveMQContainer.getConnectionString());
+
+    testMqttClient.on("message", (topic, message) => {
```
cristianrgreco, line 13: This test will always pass because the expectation is within a callback. You could wrap it in a promise:

```js
const message = await new Promise((resolve) => 
  testMqttClient.on("message", (topic, message) => 
    resolve(message)));

expect(message.toString()).toEqual("Test Message");

testMqttClient.end();
```

h2xd: Hope to say gotcha this time around. Had to wrap couple of definitions into functions, since strait `await new Promise` definitions looked weird in my eyes. Hope that these changes suite the codebase.

cristianrgreco: Thanks for making the change @h2xd. 

I am nitpicking now but it seems long for a single test. Considering the mqtt library doesn't support promises, and wrapping connecting/subscribing/publishing in promises is a bit verbose, maybe the `done` callback reads better?

What do you think?

```js
it("should connect to HiveMQ Community Edition via MQTT.js", (done) => {
  new HiveMQContainer().start().then((hiveMQContainer) => {
    const testMqttClient = mqtt.connect(hiveMQContainer.getConnectionString());

    testMqttClient.on("message", (topic, message) => {
      expect(message.toString()).toEqual("Test Message");
      testMqttClient.end();
      done();
    });

    testMqttClient.on("connect", () => {
      testMqttClient.subscribe("test", (error) => {
        if (!error) {
          testMqttClient.publish("test", "Test Message");
        }
      });
    });
  });
});
```

h2xd: No problem @cristianrgreco nitpicks are just beneficial to the overall project. Will add your proposal in a second.

cristianrgreco (approved):

cristianrgreco:
Thanks again @h2xd, I'll do a release in a couple hours and will comment here when it's done

cristianrgreco:
Released in [v9.10.0](https://github.com/testcontainers/testcontainers-node/releases/tag/v9.10.0)

PR #627, opened 2023-08-05
Improve test coverage of container runtime strategies

cristianrgreco:


src/testcontainers/src/container-runtime/strategies/npipe-socket-strategy.ts
```diff
@@ -2,12 +2,14 @@ import { ContainerRuntimeClientStrategy } from "./strategy";
...
 export class NpipeSocketStrategy implements ContainerRuntimeClientStrategy {
+  constructor(private readonly env: NodeJS.ProcessEnv = process.env) {}
+
   getName(): string {
     return "NpipeSocketStrategy";
   }
 
   async getResult(): Promise<ContainerRuntimeClientStrategyResult | undefined> {
-    if (process.platform !== "win32") {
+    if (this.env.platform !== "win32") {
```
thenglong, line 12: @cristianrgreco May I ask why this changed from `process.platform` to `this.env.platform` since the **platform** value is in the process object itself and not in the `process.env`. I believe this cause testcontainers to never find runtime strategy on docker windows unless we set the `platform` env manually.

cristianrgreco: Thanks for this @thenglong, you are correct it's an issue. I will release a patch shortly..

PR #644, opened 2023-08-23
Export container runtime

cristianrgreco:
The underlying container runtime may be useful for users to use standalone. It is automatically configured to work with different container runtimes, handles authentication, provides nice API over common Docker/DockerCompose operations, as well as general information.

### DIFF
--- a/docs/features/advanced.md
+++ b/docs/features/advanced.md
@@ -32,7 +32,7 @@ await containerRuntimeClient.container.start(container);
 ### Starting a Docker Compose environment
 
 ```js
-const environment = await containerRuntimeClient.compose.up({ ...})
+const environment = await containerRuntimeClient.compose.up({ ... })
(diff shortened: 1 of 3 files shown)


mdelapenya (approved):
LGTM, thanks for being so quick! 🚀

docs/features/advanced.md
```diff
@@ -0,0 +1,42 @@
...
+
+```js
+const container = await containerRuntimeClient.container.create({ ... });
+await containerRuntimeClient.container.start(container);
+```
+
+### Starting a Docker Compose environment
+
+```js
+const environment = await containerRuntimeClient.compose.up({ ...})
```
cristianrgreco, line 35:
```suggestion
const environment = await containerRuntimeClient.compose.up({ ... })
```

PR #642, opened 2023-08-20
Add Redis module

djakielski:
Adds new testcontainer for redis

Task:
- [x] docs
- [x] write & read tests
- [x] execute redis cli command
- [x] container start
- [x] password authentication 
- [x] persistent volume

PR #645, opened 2023-08-24
Add MSSQLServer module

wjin17:
Add MS Server module

- [x] acceptLicense
- [x] select edition
- [x] connect and query, connection uri, and secure password tests
- [x] container start
- [x] update docs


packages/modules/mssqlserver/src/mssqlserver-container.ts
```diff
@@ -0,0 +1,127 @@
...
+  private readonly port: number;
+
+  constructor(
+    startedTestContainer: StartedTestContainer,
+    private readonly database: string,
+    private readonly username: string,
+    private readonly password: string
+  ) {
+    super(startedTestContainer);
+    this.port = startedTestContainer.getMappedPort(MSSQL_PORT);
```
djakielski, line 101: When I was writing the redis module (https://github.com/testcontainers/testcontainers-node/pull/642), I also oriented on this style from mysql. But when I start testing it by restaring the container, it will fail, because the container gots new ports after restart. So I moved it to the getter function.

packages/modules/mssqlserver/src/mssqlserver-container.ts
```diff
@@ -0,0 +1,127 @@
...
+  }
+
+  public getPort(): number {
+    return this.port;
```
djakielski, line 105: Better use getMappedPort in the getter, to get actual port, also when container was restarted.
```typescript
 return this.getMappedPort(MSSQL_PORT);
```

wjin17: Thanks for the review! I updated the PR to use getMappedPort in the getter and I also removed the wait strategy in the module to so users can define their own.

packages/modules/mssqlserver/src/mssqlserver-container.ts
```diff
@@ -0,0 +1,123 @@
...
+
+type ProductKey = `${string}-${string}-${string}-${string}-${string}`;
+
+type Edition = "Evaluation" | "Developer" | "Express" | "Web" | "Standard" | "Enterprise" | ProductKey;
+
+const PASSWORD_CATEGORY_VALIDATION_PATTERNS = [/[A-Z]+/, /[a-z]+/, /[0-9]+/, /[@$!%*?&]+/];
+
+export class MSSQLServerContainer extends GenericContainer {
+  private database = "master";
+  private username = "SA";
```
eddumelendez, line 13: See https://github.com/testcontainers/testcontainers-java/pull/7238#issuecomment-1620777580

```suggestion
  private username = "sa";
```

packages/modules/mssqlserver/src/mssqlserver-container.ts
```diff
@@ -0,0 +1,123 @@
+import { AbstractStartedContainer, GenericContainer, StartedTestContainer } from "testcontainers";
+
+const MSSQL_PORT = 1433;
+
+type ProductKey = `${string}-${string}-${string}-${string}-${string}`;
+
+type Edition = "Evaluation" | "Developer" | "Express" | "Web" | "Standard" | "Enterprise" | ProductKey;
+
+const PASSWORD_CATEGORY_VALIDATION_PATTERNS = [/[A-Z]+/, /[a-z]+/, /[0-9]+/, /[@$!%*?&]+/];
```
eddumelendez, line 9: Although, this is similar to what testcontainers-java does, the current approach is to leave it to the service itself. Because, it can change and the container implementation in Testcontainers would fail when it shouldn't

packages/modules/mssqlserver/src/mssqlserver-container.ts
```diff
@@ -0,0 +1,123 @@
...
+
+  public acceptLicense(): this {
+    this.acceptEula = "Y";
+    return this;
+  }
+
+  public withDatabase(database: string): this {
+    this.database = database;
+    return this;
+  }
```
eddumelendez, line 30: AFAIK, you can not set the database name in MSSQL and due to there is no test covering the case I wonder if that works, nowadays.

packages/modules/mssqlserver/src/mssqlserver-container.ts
```diff
@@ -0,0 +1,123 @@
...
+
+type Edition = "Evaluation" | "Developer" | "Express" | "Web" | "Standard" | "Enterprise" | ProductKey;
+
+const PASSWORD_CATEGORY_VALIDATION_PATTERNS = [/[A-Z]+/, /[a-z]+/, /[0-9]+/, /[@$!%*?&]+/];
+
+export class MSSQLServerContainer extends GenericContainer {
+  private database = "master";
+  private username = "SA";
+  private password = "Passw0rd";
+  private edition: Edition = "Developer";
```
eddumelendez, line 15: `Developer` is the default value for `MSSQL_PID`, so I don't think it is needed. Also, this can added `.withEnvironment({ "MSSQL_PID": "Express })` if required.

packages/modules/mssqlserver/src/mssqlserver-container.ts
```diff
@@ -0,0 +1,123 @@
+import { AbstractStartedContainer, GenericContainer, StartedTestContainer } from "testcontainers";
+
+const MSSQL_PORT = 1433;
+
+type ProductKey = `${string}-${string}-${string}-${string}-${string}`;
```
eddumelendez, line 5: why is this needed?

eddumelendez (commented):
Thanks for your contribution. I've left some comments. Also, please add more tests where is needed.

packages/modules/mssqlserver/src/mssqlserver-container.ts
wjin17: Thank you for your comments! This is actually my first time contributing so I really appreciate the feedback.

packages/modules/mssqlserver/src/mssqlserver-container.ts
wjin17: I left this in as a convenience method for the generating the connection string should the user want to seed the instance during initialization. I can also add a test/example for seeding the database with SQL files and bash scripts but I was worried about adding too many files. Let me know how I should proceed.

packages/modules/mssqlserver/src/mssqlserver-container.ts
wjin17: Product key comes from the official documentation here: 

[https://learn.microsoft.com/en-us/sql/linux/sql-server-linux-configure-environment-variables?view=sql-server-ver16#environment-variables](https://learn.microsoft.com/en-us/sql/linux/sql-server-linux-configure-environment-variables?view=sql-server-ver16#environment-variables)

>If specifying a product key, it must be in the form of #####-#####-#####-#####-#####, where '#' is a number or a letter.

However, if the edition should be added in by the user using withEnvironment, I can remove all PID/Edition code.

packages/modules/mssqlserver/src/mssqlserver-container.test.ts
```diff
@@ -0,0 +1,81 @@
...
+import { MSSQLServerContainer } from "./mssqlserver-container";
+
+describe("MSSqlServerContainer", () => {
+  jest.setTimeout(180_000);
+
+  // connect {
+  it("should connect and return a query result", async () => {
+    const container = await new MSSQLServerContainer()
+      .acceptLicense()
+      .withWaitStrategy(Wait.forLogMessage(/.*Recovery is complete.*/, 1))
```
cristianrgreco, line 12: Could this wait strategy be configured in the module itself?

packages/modules/mssqlserver/src/mssqlserver-container.ts
```diff
@@ -0,0 +1,123 @@
...
+    this.password = password;
+    return this;
+  }
+
+  public withEdition(edition: Edition): this {
+    this.edition = edition;
+    return this;
+  }
+
+  private checkPasswordStrength(password: string) {
```
cristianrgreco, line 43: For my own understanding, why is it necessary to enforce a strong password here?

packages/modules/mssqlserver/src/mssqlserver-container.test.ts
wjin17: I left the wait strategy out of the module to allow users to use their own. You can see an example below. I can merge this branch into the PR and update the docs to include this but I wanted to keep the changes to a minimum just in case. 

[https://github.com/wjin17/testcontainers-node/blob/40385d0af17ef2d9af75165ac99da83c96787470/packages/modules/mssqlserver/src/mssqlserver-container.test.ts#L84](https://github.com/wjin17/testcontainers-node/blob/40385d0af17ef2d9af75165ac99da83c96787470/packages/modules/mssqlserver/src/mssqlserver-container.test.ts#L84)

packages/modules/mssqlserver/src/mssqlserver-container.ts
wjin17: SQL Server requires a password to meet the following criteria:

The password doesn't contain the account name of the user.

The password is at least eight characters long.

The password contains characters from three of the following four categories:

1. Latin uppercase letters (A through Z)
2. Latin lowercase letters (a through z)
3. Base 10 digits (0 through 9)
4. Non-alphanumeric characters such as: exclamation point (!), dollar sign ($), number sign (#), or percent (%).

[source](https://learn.microsoft.com/en-us/sql/relational-databases/security/password-policy?view=sql-server-ver16)

This implementation comes from Java but as @eddumelendez stated above, it looks like they're changing it so that the service will handle it.

packages/modules/mssqlserver/src/mssqlserver-container.test.ts
eddumelendez: @wjin17 take into account that Testcontainers modules are pre-configured, so, it is totally fair to use the wait strategy in the `mssql-server-container.ts`.

wjin17: @eddumelendez Great point!

I've
1. Merged the change from SA to sa
2. Removed password validation
3. Created a default wait strategy* that can be overwritten
4. Added tests for password validation using the native MSSql instance and a test using the Express edition with environment variables

I had to allow the wait strategy regex to be overwritten because the Express edition requires a different wait regex before the db can be queried.

Let me know if I should add any more changes.

Thanks again for your review!

cristianrgreco:
Hi @wjin17, thanks for the contribution! The test are failing on my M1 mac with the following:

```
testcontainers:containers [0dbfd11c6c3f] /opt/mssql/bin/sqlservr: Invalid mapping of address 0x400395b000 in reserved address space below 0x400000000000. Possible causes: +65ms
testcontainers:containers [0dbfd11c6c3f] 1) the process (itself, or via a wrapper) starts-up its own running environment sets the stack size limit to unlimited via syscall setrlimit(2); +0ms
testcontainers:containers [0dbfd11c6c3f] 2) the process (itself, or via a wrapper) adjusts its own execution domain and flag the system its legacy personality via syscall personality(2); +0ms
testcontainers:containers [0dbfd11c6c3f] 3) sysadmin deliberately sets the system to run on legacy VA layout mode by adjusting a sysctl knob vm.legacy_va_layout. +0ms
```

Could it be an issue with ARM/is a fix possible?

wjin17:
> Hi @wjin17, thanks for the contribution! The test are failing on my M1 mac with the following:
> 
> ```
> testcontainers:containers [0dbfd11c6c3f] /opt/mssql/bin/sqlservr: Invalid mapping of address 0x400395b000 in reserved address space below 0x400000000000. Possible causes: +65ms
> testcontainers:containers [0dbfd11c6c3f] 1) the process (itself, or via a wrapper) starts-up its own running environment sets the stack size limit to unlimited via syscall setrlimit(2); +0ms
> testcontainers:containers [0dbfd11c6c3f] 2) the process (itself, or via a wrapper) adjusts its own execution domain and flag the system its legacy personality via syscall personality(2); +0ms
> testcontainers:containers [0dbfd11c6c3f] 3) sysadmin deliberately sets the system to run on legacy VA layout mode by adjusting a sysctl knob vm.legacy_va_layout. +0ms
> ```
> 
> Could it be an issue with ARM/is a fix possible?
> 
> Also looks like the Azure dependencies are incompatible with Node 20 which is LTS, perhaps it's supported with a newer version of the dependency?

Hi @cristianrgreco

Unfortunately, I do not have access to an M1 mac but trying to run mssql on M1 macs seems to be a known issue. Here's a link to a fix from the official MSSQL docker repo [mssql docker issue #668](https://github.com/microsoft/mssql-docker/issues/668#issuecomment-1436802153).

For the Azure dependency, I tried updating to the latest version of [node-mssql](https://www.npmjs.com/package/mssql) using Node 20 but it seems like the official library is still using the deprecated Azure module. 

Please let me know how to proceed.

cristianrgreco:
According to https://github.com/tediousjs/node-mssql/issues/1540, node 20 is now supported, will re-run the build now.

wjin17:
> According to [tediousjs/node-mssql#1540](https://github.com/tediousjs/node-mssql/issues/1540), node 20 is now supported, will re-run the build now.

Awesome! Thank you for the update.

cristianrgreco:
Thanks for the contribution @wjin17 !

PR #648, opened 2023-09-04
Add Clickhouse module

tomershafir:
Add basic clickhouse module


packages/modules/clickhouse/src/clickhouse-container.ts
```diff
@@ -0,0 +1,108 @@
...
+        CLICKHOUSE_PASSWORD: this.password,
+      })
+      .withWaitStrategy(Wait.forHttp(`/ping`, HTTP_PORT)
+        .forStatusCode(200))
+      .withStartupTimeout(120_000);
+    
+    if (this.hostConfigPathXml)
+      this.withCopyFilesToContainer([{source: this.hostConfigPathXml, target: CONFIG_PATH_XML, mode: CONFIG_FILE_MODE}]);
+    if (this.hostConfigPathYaml)
+      this.withCopyFilesToContainer([{source: this.hostConfigPathYaml, target: CONFIG_PATH_YAML, mode: CONFIG_FILE_MODE}]);
```
eddumelendez, line 63: Although, those are very convenient, sometimes, can confuse to users when having multiple configurations. I'm not a ClickHouse expert, so, for now my suggestion would be to do not support those OOTB but rather have examples using the `withCopyFilesToContainer` as part of the documentation.

packages/modules/clickhouse/src/clickhouse-container.test.ts
```diff
@@ -0,0 +1,102 @@
...
+      .start();
+    const client = createClickhouseContainerHttpClient(container);
+    await _test(client, container.getDatabase());
+    await client.close();
+    await container.stop();
+  });
+  
+  function createClickhouseContainerHttpClient(container: StartedClickhouseContainer) {
+    return createClient({
+      host: `http://${container.getHost()}:${container.getHostPorts().http}`,
```
eddumelendez, line 69: is this should be returned by the container implementation for easy access?

packages/modules/clickhouse/src/clickhouse-container.ts
```diff
@@ -0,0 +1,108 @@
...
+    private readonly password: string,
+  ) {
+    super(startedTestContainer);
+    this.httpHostPort = startedTestContainer.getMappedPort(httpPort);
+    this.nativeHostPort = startedTestContainer.getMappedPort(nativePort);
+  }
+
+  public getHostPorts(): {http: number; native: number} {
+    return {http: this.httpHostPort, native: this.nativeHostPort}
+  }
```
eddumelendez, line 95: I think the urls should be returned instead of the ports.

eddumelendez (commented):
Thanks for your contribution, @tomershafir ! I've left some comments.

packages/modules/clickhouse/src/clickhouse-container.test.ts
tomershafir: I think the client should be a different module rather than part the container abstraction
it also couples the dev to a dep he may not want (even though it may make sense to couple because its an official client), for example qryn project doesnt use this client atm

packages/modules/clickhouse/src/clickhouse-container.test.ts
```diff
@@ -0,0 +1,89 @@
...
+      .start();
+    const client = createClickhouseContainerHttpClient(container);
+    await _test(client, container.getDatabase());
+    await client.close();
+    await container.stop();
+  });
+
+  function createClickhouseContainerHttpClient(container: StartedClickhouseContainer) {
+    return createClient({
+      host: container.getHttpUrl("http"),
```
eddumelendez, line 56: Due to the current support the `getHttpUrl()` can return the `http` protocol. If, at some point in the future, https is supported another method/fn can be added or revisited

djakielski:
@tomershafir please add your module under modules add file https://github.com/testcontainers/testcontainers-node/blob/main/mkdocs.yml. Otherwise it will be missing in the documentation.

tomershafir:
done, @cristianrgreco hey, can you merge?

cristianrgreco:
Thanks @tomershafir, please address the linting issues. The security alerts could also be solved by replacing `Math.random()` with a uuid (we have a UUID generator). I know it's for test only but it'd be nice not to get alerted each time a change is made here.

tomershafir:
done

packages/modules/clickhouse/src/clickhouse-container.ts
```diff
@@ -0,0 +1,95 @@
...
+    this.hostNativePort = startedTestContainer.getMappedPort(nativePort);
+  }
+
+  public getHttpUrl(): string {
+    return this.toUrl("http", this.hostHttpPort)
+  }
+
+  public getNativeUrl(schema: string): string {
+    return this.toUrl(schema, this.hostNativePort)
+  }
```
eddumelendez, line 78: can you provide an example of this please? or better add a test :)

tomershafir: as clickhouse-js doesnt support native protocol I removed the js methods. Added https one

eddumelendez (approved):
LGTM

cristianrgreco:
Thanks @eddumelendez :slightly_smiling_face:  Will merge when the build passes

cristianrgreco:
Closing due to inactivity.

PR #649, opened 2023-09-05
Update elasticsearch module

bijela-gora:
Hello,

Thank you for such a great tool!

I would appreciate it if you could check if the changes suggested in this PR are useful to add.

1. The useful constant `ELASTIC_SEARCH_HTTP_PORT` was exported. This change may make the interface of the @testcontainers/elasticsearch package a little more convenient. I found out that in order to allow a microservice (inside a docker container) to communicate with Elasticsearch in the other docker container, it is necessary to use port 9200, not the one published to the host.
2. I spent some time figuring out the smallest change to run the elasticsearch v8 test container. In short, it's `this.withEnvironment({ "xpack.security.enabled": "false" });` I want to share this knowledge with this PR. I can add it to the README.md instead.
3. The @elastic/elasticsearch dependency was updated to the latest version.

### DIFF
--- a/packages/modules/elasticsearch/src/elasticsearch-container.ts
+++ b/packages/modules/elasticsearch/src/elasticsearch-container.ts
@@ -3,13 +3,13 @@ import { AbstractStartedContainer, GenericContainer, StartedTestContainer } from
 export const ELASTIC_SEARCH_HTTP_PORT = 9200;
 
 export class ElasticsearchContainer extends GenericContainer {
-  constructor(image = "elasticsearch:7.17.7") {
+  constructor(image = "elasticsearch:8.9.1") {
     super(image);
   }
 
   public override async start(): Promise<StartedElasticsearchContainer> {
     this.withExposedPorts(...(this.hasExposedPorts ? this.exposedPorts : [ELASTIC_SEARCH_HTTP_PORT]))
-      .withEnvironment({ "discovery.type": "single-node" })
+      .withEnvironment({ "discovery.type": "single-node", "xpack.security.enabled": "false" })
       .withCopyContentToContainer([
         {
           content: "-Xmx2G\n",
--- a/packages/modules/elasticsearch/src/elasticsearch-container.ts
+++ b/packages/modules/elasticsearch/src/elasticsearch-container.ts
@@ -1,6 +1,6 @@
 import { AbstractStartedContainer, GenericContainer, StartedTestContainer } from "testcontainers";
 
-const ELASTIC_SEARCH_HTTP_PORT = 9200;
+export const ELASTIC_SEARCH_HTTP_PORT = 9200;
 
 export class ElasticsearchContainer extends GenericContainer {
   constructor(image = "elasticsearch:7.17.7") {
--- a/packages/modules/elasticsearch/src/index.ts
+++ b/packages/modules/elasticsearch/src/index.ts
@@ -1,1 +1,5 @@
-export { ElasticsearchContainer, StartedElasticsearchContainer } from "./elasticsearch-container";
+export {
+  ELASTIC_SEARCH_HTTP_PORT,
+  ElasticsearchContainer,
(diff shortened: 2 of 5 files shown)


bijela-gora:
@cristianrgreco Thank you for your attention to this PR. All checks have passed.

packages/modules/elasticsearch/src/elasticsearch-container.ts
```diff
@@ -1,15 +1,15 @@
...
 export class ElasticsearchContainer extends GenericContainer {
-  constructor(image = "elasticsearch:7.17.7") {
+  constructor(image = "elasticsearch:8.9.1") {
     super(image);
   }
 
   public override async start(): Promise<StartedElasticsearchContainer> {
     this.withExposedPorts(...(this.hasExposedPorts ? this.exposedPorts : [ELASTIC_SEARCH_HTTP_PORT]))
-      .withEnvironment({ "discovery.type": "single-node" })
+      .withEnvironment({ "discovery.type": "single-node", "xpack.security.enabled": "false" })
```
eddumelendez, line 12: I don't think it will be a good idea to default to false. In this case we can let the user decide. So, as part of the Elasticsearch container definition, this en var can be set.

packages/modules/elasticsearch/src/index.ts
```diff
@@ -1 +1,5 @@
-export { ElasticsearchContainer, StartedElasticsearchContainer } from "./elasticsearch-container";
+export {
+  ELASTIC_SEARCH_HTTP_PORT,
```
eddumelendez, line 2: In Testcontainers for Java, those values are not exported. However, what we did lately is to document all the ports used by the containers implementations.

packages/modules/elasticsearch/src/elasticsearch-container.ts
bijela-gora: Any suggestions?

The motivation of this change was to prevent as much braking changes as possible. Without this change all users of `new ElasticsearchContainer()` (without argument passed) will get failed tests. Because I changed default value of `image` parameter.

Am I right that users can override environment by calling `.withEnvironment({ "xpack.security.enabled": "true" }`?

packages/modules/elasticsearch/src/index.ts
bijela-gora: I added this because it is convenient for me and it may be convenient for others. An alternative to exporting a constant is to copy and paste that constant into a project that uses `ElasticsearchContainer`.

By the way, thank you for your time for this review.

packages/modules/elasticsearch/src/elasticsearch-container.ts
eddumelendez: When testing against specific technology, you want to tests using the right version. This is something we really encourage to do. Using default images as you mention bring the drawback you mention. But, currently the default is version 7 so if you are moving to 8 and declaring the image version then you should be aware of what implies that upgrade, in this case, security is enabled by default.

> Am I right that users can override environment by calling .withEnvironment({ "xpack.security.enabled": "true" }?

It should work.

packages/modules/elasticsearch/src/index.ts
eddumelendez: Yep, totally make sense. I'm not saying this is a bad idea but just making sure we are aligned in the different language implementations :)

packages/modules/elasticsearch/src/elasticsearch-container.ts
bijela-gora: > you should be aware of what implies that upgrade, in this case, security is enabled by default.

Now I understand how you see the modules. It's a good idea to make them as close as possible to real-world use.

The parameter `"xpack.security.enabled": "true"` works for me because I don't care about security when testing queries/responses. **I thought testing queries/responses is the main use-case for this module.** Is there any statistic about how elastic module are used most of the time?

PR #657, opened 2023-09-25
Allow usage of --no-deps and others docker compose options in docker compose environment

### LINKED ISSUE #656: Support of --no-deps and others docker compose option
**Expected Behaviour**

In addition to the support of start only selected services present in docker compose stack with : 

```javascript
new DockerComposeEnvironment("docker-compose.yml").up(["my_service"]);
```

It will be useful to support the docker compose `--no-deps` option to allow start only specified service even if it declare dependencies

The syntax could be :

```javascript
new DockerComposeEnvironment("docker-compose.yml")
  .withNoDeps()
  .up(["my_service"]);
```

or :

```javascript
new DockerComposeEnvironment("docker-compose.yml")
  .withCommandOptions("--no-deps")
  .up(["my_service"]);
```

**Actual Behaviour**

Not supported

VanRoy:
Resolves #656

### DIFF
--- a/packages/testcontainers/src/docker-compose-environment/docker-compose-environment.test.ts
+++ b/packages/testcontainers/src/docker-compose-environment/docker-compose-environment.test.ts
@@ -255,4 +255,42 @@ describe("DockerComposeEnvironment", () => {
 
     await startedEnvironment.down();
   });
+
+  it("should start selected container by service without its dependencies", async () => {
+    const startedEnvironment = await new DockerComposeEnvironment(fixtures, "docker-compose-with-no-deps.yml")
+      .withNoDeps()
+      .up(["service_2"]);
+
+    await Promise.all(
+      [await composeContainerName("service_2")].map(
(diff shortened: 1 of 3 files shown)


packages/testcontainers/src/docker-compose-environment/docker-compose-environment.test.ts
```diff
@@ -255,4 +255,42 @@ describe("DockerComposeEnvironment", () => {
...
+
+    const service1 = await composeContainerName("service_1");
+    expect(() => startedEnvironment.getContainer(service1)).toThrow(
+      `Cannot get container "${service1}" as it is not running`
+    );
+
+    await startedEnvironment.down();
+  });
+
+  it("should start selected container by service without its dependencies by command options", async () => {
```
cristianrgreco, line 278: I'd rather separate supporting arbitrary command options for now because I'm not sure how to support it. E.g as we switch support from docker-compose v1 to v2 we could break user's options, whereas if testcontainers explicitly supports command like `withNoDeps` then we can unsure it works no matter what.

PR #663, opened 2023-10-25
Allow to start containers with raw image IDs

gzm0:
Docker supports starting containers with bare image IDs:

```
docker run aa285b773a2c042056883845aea893a743d358a5d40f61734fa228fde93dae6f
```

However, if a tag or digest is added to the ID, Docker will refuse it:

```
docker run aa285b773a2c042056883845aea893a743d358a5d40f61734fa228fde93dae6f:latest
Unable to find image 'aa285b773a2c042056883845aea893a743d358a5d40f61734fa228fde93dae6f:latest' locally
docker: Error response from daemon: invalid repository name (aa285b773a2c042056883845aea893a743d358a5d40f61734fa228fde93dae6f), cannot specify 64-byte hexadecimal strings.
See 'docker run --help'.
```

We adjust the string conversion of ImageName for this specific case to allow starting containers with bare image IDs.

More context:
https://github.com/testcontainers/testcontainers-java/discussions/7707

### DIFF
--- a/packages/testcontainers/src/container-runtime/image-name.test.ts
+++ b/packages/testcontainers/src/container-runtime/image-name.test.ts
@@ -41,7 +41,10 @@ describe("ContainerImage", () => {
     });
 
     it("should keep other tags (not `latest`) on image IDs", () => {
-      // Note that the resulting image ID will not be accepted by docker.
+      // Note that the resulting image ID will not be accepted by Docker:
+      //
+      // > "invalid repository name [...], cannot specify 64-byte hexadecimal strings"
+      //
       // However, not treating tags other than `latests` specially is probably less surprising.
       const imageName = new ImageName(
         undefined,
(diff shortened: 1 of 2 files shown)


packages/testcontainers/src/container-runtime/image-name.test.ts
```diff
@@ -96,5 +116,13 @@ describe("ContainerImage", () => {
       expect(imageName.image).toBe("image");
       expect(imageName.tag).toBe("sha256:1234abcd1234abcd1234abcd1234abcd");
     });
+
+    it("should work with image being an image ID", () => {
```
cristianrgreco, line 120: Similarly to `should keep other tags (not latest) on image IDs`, do we need a `fromString` test where the input has a non-latest tag (`aa285b773a2c042056883845aea893a743d358a5d40f61734fa228fde93dae6f:1`)?

gzm0: Yes, that's probably a good idea. I've added a commit with a test (and also added the docker error message to the text explaining that it won't accept the reference).

PR #668, opened 2023-11-10
Add support for setting `user`, `workingDir` and `env` when executing a command in a container

LNSD:
This pull request introduces support for specifying `User`, `WorkingDir`, and `Env` options during container execution in both Docker ([API docs](https://docs.docker.com/engine/api/v1.43/#tag/Exec)) and Podman ([API docs](https://docs.podman.io/en/latest/_static/api.html#tag/exec-(compat))), enhancing customization.

#### Changes:

Added support to `StartedTestContainer.exec(command, opts)` function to provide the following options:

1. **`user`:**
   - Users can set the user, and optionally, group to run the exec process inside the container. Format is one of: `user`, `user:group`, `uid`, or `uid:gid`.

2. **`workingDir`:**
   - The `WorkingDir` option allows users to define the working directory for the exec process inside the container.

3. **`env`:**
   - Users can now include the `Env` option when executing commands, allowing for environment variable customization.

#### Example:

```Typescript
const container = await new GenericContainer("alpine")
  .withCommand(["sleep", "infinity"])
  .start();

const { output, exitCode } = await container.exec(["echo", "hello", "world"], {
	workingDir: "/app/src/",
	user: "alpine",
	env: {
		"VAR1": "enabled",
		"VAR2": "full",
	}
});
```

---

Your feedback and testing contributions are appreciated.

### DIFF
--- a/docs/features/containers.md
+++ b/docs/features/containers.md
@@ -495,6 +495,9 @@ const container = await new GenericContainer("alpine")
 
 ## Running commands
 
+To run a command inside an already started container use the `exec` method. The command will be run in the container's 
+working directory, returning the command output and exit code:
+
 ```javascript
 const container = await new GenericContainer("alpine")
   .withCommand(["sleep", "infinity"])
@@ -502,6 +505,30 @@ const container = await new GenericContainer("alpine")
 
 const { output, exitCode } = await container.exec(["echo", "hello", "world"]);
 ```
+
+The following options can be provided to modify the command execution:
+
+1. **`user`:** The user, and optionally, group to run the exec process inside the container. Format is one of: `user`, `user:group`, `uid`, or `uid:gid`.
+
+2. **`workingDir`:** The working directory for the exec process inside the container.
+
+3. **`env`:** A map of environment variables to set inside the container.
+
+
+```javascript
+const container = await new GenericContainer("alpine")
+  .withCommand(["sleep", "infinity"])
+  .start();
+
+const { output, exitCode } = await container.exec(["echo", "hello", "world"], {
+	workingDir: "/app/src/",
+	user: "1000:1000",
+	env: {
+		"VAR1": "enabled",
+		"VAR2": "/app/debug.log",
+	}
+});
+````
 
 ## Streaming logs
 
--- a/packages/testcontainers/src/container-runtime/clients/container/container-client.ts
+++ b/packages/testcontainers/src/container-runtime/clients/container/container-client.ts
@@ -7,7 +7,7 @@ import Dockerode, {
   Network,
 } from "dockerode";
 import { Readable } from "stream";
-import { ExecResult } from "./types";
+import { ExecOptions, ExecResult } from "./types";
 
 export interface ContainerClient {
   dockerode: Dockerode;
@@ -22,7 +22,7 @@ export interface ContainerClient {
   stop(container: Container, opts?: { timeout: number }): Promise<void>;
   attach(container: Container): Promise<Readable>;
   logs(container: Container, opts?: ContainerLogsOptions): Promise<Readable>;
-  exec(container: Container, command: string[], opts?: { log: boolean }): Promise<ExecResult>;
+  exec(container: Container, command: string[], opts?: Partial<ExecOptions>): Promise<ExecResult>;
   restart(container: Container, opts?: { timeout: number }): Promise<void>;
   remove(container: Container, opts?: { removeVolumes: boolean }): Promise<void>;
   connectToNetwork(container: Container, network: Network, networkAliases: string[]): Promise<void>;
--- a/packages/testcontainers/src/container-runtime/clients/container/docker-container-client.ts
+++ b/packages/testcontainers/src/container-runtime/clients/container/docker-container-client.ts
@@ -4,11 +4,12 @@ import Dockerode, {
   ContainerInfo,
   ContainerInspectInfo,
   ContainerLogsOptions,
+  ExecCreateOptions,
   Network,
 } from "dockerode";
 import { PassThrough, Readable } from "stream";
 import { IncomingMessage } from "http";
-import { ExecResult } from "./types";
+import { ExecOptions, ExecResult } from "./types";
 import byline from "byline";
 import { ContainerClient } from "./container-client";
 import { log, execLog, streamToString } from "../../../common";
@@ -173,21 +174,31 @@ export class DockerContainerClient implements ContainerClient {
     }
   }
 
-  async exec(container: Container, command: string[], opts?: { log: boolean }): Promise<ExecResult> {
-    const chunks: string[] = [];
+  async exec(container: Container, command: string[], opts?: Partial<ExecOptions>): Promise<ExecResult> {
+    const execOptions: ExecCreateOptions = {
+      Cmd: command,
+      AttachStdout: true,
+      AttachStderr: true,
+    };
 
+    if (opts?.env !== undefined) {
+      execOptions.Env = Object.entries(opts.env).map(([key, value]) => `${key}=${value}`);
+    }
+    if (opts?.workingDir !== undefined) {
+      execOptions.WorkingDir = opts.workingDir;
+    }
+    if (opts?.user !== undefined) {
+      execOptions.User = opts.user;
+    }
+
+    const chunks: string[] = [];
     try {
       if (opts?.log) {
         log.debug(`Execing container with command "${command.join(" ")}"...`, { containerId: container.id });
       }
-      const exec = await container.exec({
-        Cmd: command,
-        AttachStdout: true,
-        AttachStderr: true,
-      });
 
+      const exec = await container.exec(execOptions);
       const stream = await exec.start({ stdin: true, Detach: false, Tty: true });
-
       if (opts?.log && execLog.enabled()) {
         byline(stream).on("data", (line) => execLog.trace(line, { containerId: container.id }));
       }
@@ -200,7 +211,7 @@ export class DockerContainerClient implements ContainerClient {
       stream.destroy();
 
       const inspectResult = await exec.inspect();
-      const exitCode = inspectResult.ExitCode === null ? -1 : inspectResult.ExitCode;
+      const exitCode = inspectResult.ExitCode ?? -1;
       const output = chunks.join("");
       if (opts?.log) {
         log.debug(`Execed container with command "${command.join(" ")}"...`, { containerId: container.id });
--- a/packages/testcontainers/src/container-runtime/clients/container/podman-container-client.ts
+++ b/packages/testcontainers/src/container-runtime/clients/container/podman-container-client.ts
@@ -1,20 +1,34 @@
-import { Container } from "dockerode";
-import { ExecResult } from "./types";
+import { Container, ExecCreateOptions } from "dockerode";
+import { ExecOptions, ExecResult } from "./types";
 import byline from "byline";
 import { DockerContainerClient } from "./docker-container-client";
 import { execLog, log } from "../../../common";
 
 export class PodmanContainerClient extends DockerContainerClient {
-  override async exec(container: Container, command: string[], opts?: { log: boolean }): Promise<ExecResult> {
+  override async exec(container: Container, command: string[], opts?: Partial<ExecOptions>): Promise<ExecResult> {
+    const execOptions: ExecCreateOptions = {
+      Cmd: command,
+      AttachStdout: true,
+      AttachStderr: true,
+    };
+
+    if (opts?.env !== undefined) {
+      execOptions.Env = Object.entries(opts.env).map(([key, value]) => `${key}=${value}`);
+    }
+    if (opts?.workingDir !== undefined) {
+      execOptions.WorkingDir = opts.workingDir;
+    }
+    if (opts?.user !== undefined) {
+      execOptions.User = opts.user;
+    }
+
     const chunks: string[] = [];
-
     try {
-      const exec = await container.exec({
-        Cmd: command,
-        AttachStdout: true,
-        AttachStderr: true,
-      });
+      if (opts?.log) {
+        log.debug(`Execing container with command "${command.join(" ")}"...`, { containerId: container.id });
+      }
 
+      const exec = await container.exec(execOptions);
       const stream = await this.demuxStream(container.id, await exec.start({ stdin: true, Detach: false, Tty: true }));
       if (opts?.log && execLog.enabled()) {
         byline(stream).on("data", (line) => execLog.trace(line, { containerId: container.id }));
@@ -28,7 +42,7 @@ export class PodmanContainerClient extends DockerContainerClient {
       stream.destroy();
 
       const inspectResult = await exec.inspect();
-      const exitCode = inspectResult.ExitCode === null ? -1 : inspectResult.ExitCode;
+      const exitCode = inspectResult.ExitCode ?? -1;
       const output = chunks.join("");
 
       return { output, exitCode };
--- a/packages/testcontainers/src/generic-container/generic-container.test.ts
+++ b/packages/testcontainers/src/generic-container/generic-container.test.ts
@@ -51,6 +51,40 @@ describe("GenericContainer", () => {
     await container.stop();
   });
 
+  it("should execute a command in a different working directory", async () => {
+    const container = await new GenericContainer("cristianrgreco/testcontainer:1.1.14").withExposedPorts(8080).start();
+
+    const { output, exitCode } = await container.exec(["pwd"], { workingDir: "/var/log" });
+
+    expect(exitCode).toBe(0);
+    expect(output).toEqual(expect.stringContaining("/var/log"));
+
+    await container.stop();
+  });
+
+  it("should execute a command with custom environment variables", async () => {
+    const container = await new GenericContainer("cristianrgreco/testcontainer:1.1.14").withExposedPorts(8080).start();
+
+    const { output, exitCode } = await container.exec(["env"], { env: { TEST_ENV: "test" } });
+
+    expect(exitCode).toBe(0);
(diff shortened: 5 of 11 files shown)


packages/testcontainers/src/container-runtime/clients/container/container-client.ts
```diff
@@ -22,7 +22,11 @@ export interface ContainerClient {
   stop(container: Container, opts?: { timeout: number }): Promise<void>;
   attach(container: Container): Promise<Readable>;
   logs(container: Container, opts?: ContainerLogsOptions): Promise<Readable>;
-  exec(container: Container, command: string[], opts?: { log: boolean }): Promise<ExecResult>;
+  exec(
+    container: Container,
+    command: string[],
+    opts?: Partial<ExecOptions>,
+  ): Promise<ExecResult>;
```
javierlopezdeancos, line 29: It's me, or probably you don't have configured prettier to format on save? If you run `npm run format` you can see how there are at least 4 files to be formatted.

We should add to our precommit hook the format script adapted to lint-stage? 🤔

javierlopezdeancos, line 29:
```suggestion
 exec(container: Container, command: string[], opts?: Partial<ExecOptions>): Promise<ExecResult>;
```

packages/testcontainers/src/container-runtime/clients/container/docker-container-client.ts
```diff
@@ -173,21 +174,35 @@ export class DockerContainerClient implements ContainerClient {
     }
   }
 
-  async exec(container: Container, command: string[], opts?: { log: boolean }): Promise<ExecResult> {
-    const chunks: string[] = [];
+  async exec(
+    container: Container,
+    command: string[],
+    opts?: Partial<ExecOptions>,
+  ): Promise<ExecResult> {
```
javierlopezdeancos, line 181:
```suggestion
  async exec(container: Container, command: string[], opts?:Partial<ExecOptions> ): Promise<ExecResult> {
```

packages/testcontainers/src/container-runtime/clients/container/podman-container-client.ts
```diff
@@ -1,20 +1,38 @@
...
 import { execLog, log } from "../../../common";
 
 export class PodmanContainerClient extends DockerContainerClient {
-  override async exec(container: Container, command: string[], opts?: { log: boolean }): Promise<ExecResult> {
-    const chunks: string[] = [];
+  override async exec(
+    container: Container,
+    command: string[],
+    opts?: Partial<ExecOptions>,
+  ): Promise<ExecResult> {
```
javierlopezdeancos, line 12:
```suggestion
  override async exec(container: Container, command: string[], opts?:Partial<ExecOptions>): Promise<ExecResult> {
```

Same prettier format problem

javierlopezdeancos (changes requested):
I think the addition of these options is very useful and necessary when executing commands within a container. 💯 

Just some nits relevant to the format that should give us prettier.

javierlopezdeancos:
What difference and advantage we have doing 

```typescript
const container = await new GenericContainer("alpine")
  .withCommand(["sleep", "infinity"])
  .start();

const { output, exitCode } = await container.exec(["echo", "hello", "world"], {
	workingDir: "/app/src/",
	user: "alpine",
	env: {
		"VAR1": "enabled",
		"VAR2": "full",
	}
});
```

instead or doing 🤔 ?

```typescript
const startedContainer = await new GenericContainer("alpine")
  .withWorkingDir("/app/src/")
  .withUser("alpine")
  .withEnvironment({
        "VAR1": "enabled",
         "VAR2": "full",
   })
  .withCommand(["sleep", "infinity"])
  .start();

const { output, exitCode } = await container.exec(["echo", "hello", "world"]);
```

apart from being able to redefine these options in each execution of a command?

docs/features/containers.md
```diff
@@ -503,6 +506,31 @@ const container = await new GenericContainer("alpine")
 const { output, exitCode } = await container.exec(["echo", "hello", "world"]);
 ```
 
+The following options can be provided to modify the command execution:
+
+1. **`user`:** The user, and optionally, group to run the exec process inside the container. Format is one of: `user`,
+               `user:group`, `uid`, or `uid:gid`.
```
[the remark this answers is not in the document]
LNSD: Done ✅

packages/testcontainers/src/container-runtime/clients/container/container-client.ts
LNSD: I just executed `npm run format` and pushed the changes 😁

----

> We should add to our precommit hook the format script adapted to lint-stage? 🤔

I would add the format task (before running the linter) in the pre-commit git hook.  But this is a personal decision. This is something someone who works frequently with this codebase should take. cc @cristianrgreco

LNSD:
The two things are different. 

> What difference and advantage we have doing
> 
> ```ts
> const container = await new GenericContainer("alpine")
>   .withCommand(["sleep", "infinity"])
>   .start();
> 
> const { output, exitCode } = await container.exec(["echo", "hello", "world"], {
> 	workingDir: "/app/src/",
> 	user: "alpine",
> 	env: {
> 		"VAR1": "enabled",
> 		"VAR2": "full",
> 	}
> });
> ```

The first, which I added support for, specifies the current working directory for the specific command you are running in **an already-running container**. So, you are only affecting the command you execute. You must specify the `workingDir` option for every new command you want to execute in the container.

The use case here is:

> I want to run a command in a directory different than the container running the main process.

For example, I want to modify on-the-fly a configuration file stored in `/etc/app/app.conf`, by executing a `sed` command, while the container's main process is running and serving requests.

> instead or doing 🤔 ?
> 
> ```ts
> const startedContainer = await new GenericContainer("alpine")
>   .withWorkingDir("/app/src/")
>   .withUser("alpine")
>   .withEnvironment({
>         "VAR1": "enabled",
>          "VAR2": "full",
>    })
>   .withCommand(["sleep", "infinity"])
>   .start();
> 
> const { output, exitCode } = await startedContainer.exec(["echo", "hello", "world"]);
> ```
> 
> apart from being able to redefine these options in each execution of a command?

This second case affects the current working directory of **the new container you will create** after calling `start()`. 

This _`ENV`/`WORKDIR`/`CMD` override_ is a "global" modification that will affect all commands running within the container (the ENTRYPOINT+CMD and all the processes executed using the `exec()` function). By using `withCommand(...)`, you are also changing the new container's start command.

For reference:

* https://docs.docker.com/engine/reference/builder/#cmd
* https://devtron.ai/blog/cmd-and-entrypoint-differences/

javierlopezdeancos (approved):

javierlopezdeancos:
The feature has really sense to me @LNSD thanks to clarify with all this great explanation

packages/testcontainers/src/container-runtime/clients/container/docker-container-client.ts
```diff
@@ -200,7 +211,7 @@ export class DockerContainerClient implements ContainerClient {
       stream.destroy();
 
       const inspectResult = await exec.inspect();
-      const exitCode = inspectResult.ExitCode === null ? -1 : inspectResult.ExitCode;
+      const exitCode = inspectResult.ExitCode ?? -1;
```
mdelapenya, line 214: I'm not familiar with Node idioms, but I've found the docs for [this operator](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Nullish_coalescing). IIUC, if the code is null or undefined, it will return -1, else, the code. And it makes sense to convert the ternary expression to the coalescence one. My question is about the usage across the library, is it used elsewhere so that we can accept the pattern? If not, I'd keep the PR without it and probably submit a separate one updating all the ternary usages if possible 🙏 Does it make sense to you?

packages/testcontainers/src/container-runtime/clients/container/container-client.ts
```diff
@@ -22,7 +22,7 @@ export interface ContainerClient {
   stop(container: Container, opts?: { timeout: number }): Promise<void>;
   attach(container: Container): Promise<Readable>;
   logs(container: Container, opts?: ContainerLogsOptions): Promise<Readable>;
-  exec(container: Container, command: string[], opts?: { log: boolean }): Promise<ExecResult>;
+  exec(container: Container, command: string[], opts?: Partial<ExecOptions>): Promise<ExecResult>;
```
mdelapenya, line 25: Asking without being a Typescript person: would this be a risk for the existing users, as a breaking change? If they were already passing an opts object including the log boolean, what will happen to them at compile and run time? Because we are introducing here a new type for the ExecOptions (which makes sense to me in any case). Do you think this can cause a BC?

packages/testcontainers/src/container-runtime/clients/container/podman-container-client.ts
```diff
@@ -28,7 +42,7 @@ export class PodmanContainerClient extends DockerContainerClient {
       stream.destroy();
 
       const inspectResult = await exec.inspect();
-      const exitCode = inspectResult.ExitCode === null ? -1 : inspectResult.ExitCode;
+      const exitCode = inspectResult.ExitCode ?? -1;
```
mdelapenya, line 45: Ditto: see my previous comment

packages/testcontainers/src/container-runtime/clients/container/docker-container-client.ts
```diff
@@ -173,21 +174,31 @@ export class DockerContainerClient implements ContainerClient {
...
+
+    if (opts?.env !== undefined) {
+      execOptions.Env = Object.entries(opts.env).map(([key, value]) => `${key}=${value}`);
+    }
+    if (opts?.workingDir !== undefined) {
+      execOptions.WorkingDir = opts.workingDir;
+    }
+    if (opts?.user !== undefined) {
+      execOptions.User = opts.user;
+    }
```
mdelapenya, line 192: Not a blocker: have you checked how this null/undefined checks are done elsewhere in the library?

Just to confirm the evaluations are done consistently in the project.

Other than that. The ifs are fine for me.

mdelapenya:
BTW it'd be important if you could add tests demonstrating the new behaviour 🙏

packages/testcontainers/src/types.ts
```diff
@@ -81,6 +81,8 @@ export type RegistryConfig = {
 
 export type BuildArgs = { [key in string]: string };
 
+export type ExecOptions = { workingDir: string; user: string; env: Environment };
```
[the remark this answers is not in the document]
LNSD: I wanted to respect the original public API as much as possible to avoid any undesired issues. So, following what already existed, I opted not to add a log option here in the "public" API.

packages/testcontainers/src/container-runtime/clients/container/docker-container-client.ts
LNSD: Honestly, this change solves a _SonarLint_ warning. It was an opportunistic change. As I was touching this file, I decided to solve this minor lint issue.

From my experience, performing small maintenance changes as you move along the codebase is better than making a big PR with multiple changes like this. This is mainly because, often, these changes are very _context-dependent_, and when you are working in a specific part of your codebase, you already have the context.

Said that, I don't have any strong opinion here; I would be okay reverting this line's changes 😁

packages/testcontainers/src/container-runtime/clients/container/container-client.ts
LNSD: Note that I added two different `ExecOptions` types in this PR. 

One, in the public API that does not include the `log` parameter, and another that supports the `log` option in the container client interface (this case above 👆🏼).

So, TL;DR, these changes are backwards compatible 🙂

packages/testcontainers/src/container-runtime/clients/container/docker-container-client.ts
LNSD: Sorry, I am not that familiar with the codebase style. 

But, for some context, the issue here boils down to what Javascript (and Typescript) understand as a "falsy" value (see [this](https://developer.mozilla.org/en-US/docs/Glossary/Falsy)).

packages/testcontainers/src/container-runtime/clients/container/docker-container-client.ts
mdelapenya: I'm fine adding it, not opposed to it at all. Just coming as a reviewer from another language makes it very difficult to adapt the mental model for new idioms very quickly, that's why I ask first to trust the author :)

For context, I'm not opposed to introduce anything new (as this one) given the tiny size of it. I just want to make sure we are clear and consistent for the sake of the project health, not introducing patterns "because yes".

In this case, I've double-checked the codebase, and it's already used 👏 So good to me! So thank you so much for your explanation about the rationale, as it makes me realise it'd be safe adding it even though not used before.

```
$ grep -R "??" .
./packages/testcontainers/src/container-runtime/clients/compose/default-compose-options.ts:  const log = options.logger ?? composeLog;
./packages/testcontainers/src/container-runtime/clients/image/docker-image-client.ts:      const authconfig = await getAuthConfig(imageName.registry ?? this.indexServerAddress);
./packages/testcontainers/src/container-runtime/clients/container/docker-container-client.ts:        since: opts?.since ?? 0,
./packages/testcontainers/src/container-runtime/auth/auths.ts:    const authEntries = dockerConfig.auths ?? {};
./packages/testcontainers/src/container-runtime/utils/pull-image.ts:    const authconfig = await getAuthConfig(options.imageName.registry ?? indexServerAddress);
./packages/testcontainers/src/utils/dockerfile-parser.ts:      .map((line) => line.replace(buildArgRegex, (_, arg) => buildArgs[arg] ?? ""))
./packages/testcontainers/src/generic-container/generic-container-builder.ts:        const authConfig = await getAuthConfig(imageName.registry ?? indexServerAddress);
./packages/testcontainers/src/generic-container/generic-container.ts:        ...(this.hostConfig.ExtraHosts ?? []),
./packages/testcontainers/src/generic-container/generic-container.ts:      ...(this.createOpts.Env ?? []),
./packages/testcontainers/src/generic-container/generic-container.ts:      ...(this.hostConfig.Ulimits ?? []),
./packages/testcontainers/src/generic-container/generic-container.ts:    this.hostConfig.CapAdd = [...(this.hostConfig.CapAdd ?? []), ...capabilities];
./packages/testcontainers/src/generic-container/generic-container.ts:    this.hostConfig.CapDrop = [...(this.hostConfig.CapDrop ?? []), ...capabilities];
./packages/testcontainers/src/generic-container/generic-container.ts:      ...(this.hostConfig.ExtraHosts ?? []),
./packages/testcontainers/src/reaper/reaper.ts:export const REAPER_IMAGE = process.env["RYUK_CONTAINER_IMAGE"] ?? "testcontainers/ryuk:0.5.1";
./packages/testcontainers/src/reaper/reaper.ts:    sessionId = reaperContainer?.Labels["org.testcontainers.session-id"] ?? new RandomUuid().nextUuid();
./packages/testcontainers/src/port-forwarder/port-forwarder.ts:export const SSHD_IMAGE = process.env["SSHD_CONTAINER_IMAGE"] ?? "testcontainers/sshd:1.1.0";
```

packages/testcontainers/src/container-runtime/clients/container/container-client.ts
javierlopezdeancos: I agree that  is following the same pattern that we have to other kind of types like `ExecResult`.  So in the internal API we keep log prop without breaking risk at we had without this PR.

I agree to add test and, at least to me, is a LGTM.

mdelapenya (approved):
This LGTM, added a comment regarding the added test, but other than that, thank you for this contribution. Looks a great improvement!

BTW I'll not merge it immediately, as I'd like to give @cristianrgreco a couple of days to have the chance to take a look if needed.

packages/testcontainers/src/generic-container/generic-container.test.ts
```diff
@@ -51,6 +51,17 @@ describe("GenericContainer", () => {
     await container.stop();
   });
 
+  it("should execute a command in a different working directory", async () => {
+    const container = await new GenericContainer("cristianrgreco/testcontainer:1.1.14").withExposedPorts(8080).start();
+
+    const { output, exitCode } = await container.exec(["pwd"], { workingDir: "/var/log" });
```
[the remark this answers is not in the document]
LNSD: Sure, added here: https://github.com/testcontainers/testcontainers-node/pull/668/commits/389e3495d2cd44aaea34afb9647da1ef4f096527

LNSD:
Rebased the PR on top of the CI fix 🔁

cristianrgreco:
Thanks for raising the PR @LNSD, very well tested and documented 🙂 

Because `ExecOptions` is now part of the public API: https://github.com/testcontainers/testcontainers-node/blob/bb25fe4677335eb8aaf54f1c1cd34e8373859e01/packages/testcontainers/src/test-container.ts#L77 We'll need to export the type here: https://github.com/testcontainers/testcontainers-node/blob/824cd8dbac0615ae00a83d82e8625a108b885ecb/packages/testcontainers/src/index.ts#L22.

Other than that it LGTM

packages/testcontainers/src/index.ts
```diff
@@ -19,7 +19,7 @@ export { Network, StartedNetwork, StoppedNetwork } from "./network/network";
 export { Wait } from "./wait-strategies/wait";
 export { StartupCheckStrategy, StartupStatus } from "./wait-strategies/startup-check-strategy";
 export { PullPolicy, ImagePullPolicy } from "./utils/pull-policy";
-export { InspectResult, Content, ExecResult } from "./types";
+export { InspectResult, Content, ExecOptions, ExecResult } from "./types";
```
LNSD, line 22: Done ✔️ 😁

PR #670, opened 2023-11-13
Add support for fetching container logs from a given time

NuShoSinkuPomogliTebeTvoiSankcii:
Added ability to pass options to dockerode logs API for StartedContainer.
I need it to pass [since](https://docs.docker.com/engine/reference/commandline/logs/#options) parameter for example.


NuShoSinkuPomogliTebeTvoiSankcii:
Hello! @javierlopezdeancos

Can u please review my PR?

NuShoSinkuPomogliTebeTvoiSankcii:
Hello! @kiview

Can u please review my PR? It's very simple

javierlopezdeancos (approved):

javierlopezdeancos:
Hey @NuShoSinkuPomogliTebeTvoiSankcii can you add some description or link the issue to get context and leave it track?

NuShoSinkuPomogliTebeTvoiSankcii:
@javierlopezdeancos 
My changes added ability to pass options to dockerode logs API for StartedContainer.
I need it to pass [since](https://docs.docker.com/engine/reference/commandline/logs/#options) parameter for example.

javierlopezdeancos (approved):

javierlopezdeancos:
Thanks to your clarification. is LGTM
I notice that I dont have permissions with my approval, probably @mdelapenya can take another look and approve to merge?

mdelapenya:
> @javierlopezdeancos My changes added ability to pass options to dockerode logs API for StartedContainer. I need it to pass [since](https://docs.docker.com/engine/reference/commandline/logs/#options) parameter for example.

Thanks for submitting this PR @NuShoSinkuPomogliTebeTvoiSankcii , could you please update the PR description with the motivation for it (the same in your comment). We need discoverability for the future us when going back to the code, and for eventual readers to check why some code was introduced in the project. It's on us to update the PR template including certain sections, like "What does this PR do?" and "Why is it important?" (PTAL to other testcontainers projects)

In the mean time, this PR looks good to me. Will merge it once the description is updated and the CI passes.

Thanks!

mdelapenya:
Oh I forgot! Could you please add a test demonstrating the new behaviour? 🙏

packages/testcontainers/src/container-runtime/clients/container/docker-container-client.ts
```diff
@@ -158,6 +158,7 @@ export class DockerContainerClient implements ContainerClient {
     try {
       log.debug(`Fetching container logs...`, { containerId: container.id });
       const stream = (await container.logs({
+        ...opts,
```
NuShoSinkuPomogliTebeTvoiSankcii, line 161: added ability to pass all params to logs

earlier we can pass only since option

NuShoSinkuPomogliTebeTvoiSankcii:
@mdelapenya thanks for your advice! Added tests and found out that I missed some logic (added it to docker-container-client).

Also added description to my PR.

NuShoSinkuPomogliTebeTvoiSankcii:
@mdelapenya Can you also publish new library version to npm after my PR will be merged, please?

packages/testcontainers/src/container-runtime/clients/container/docker-container-client.ts
mdelapenya: What will happen if the user of this method passes opts with follow, stdout and/or stderr with false value? Will those values be overridden by the defaults in https://github.com/testcontainers/testcontainers-node/pull/670/files#diff-cde64c872b844fe1a7e8949d7eee46cea0cb346b40d840597c1a4785d68998d2R162-R164?

If so, it will cause an unexpected behaviour on the user side, as the values won't be applied but overridden by the library.

I'm not a Typescript developer, so sorry if this is an incorrect observation from my side.

If that's the case I'd go with defining a fallback/default opts variable that will take the current default values if the opts passed by the users are empty/invalid/null. Wdyt?

packages/testcontainers/src/generic-container/generic-container-logs.test.ts
```diff
@@ -26,6 +26,16 @@ describe("GenericContainer logs", () => {
     await container.stop();
   });
 
+  it("should stream logs with ContainerLogsOptions from a started container", async () => {
+    const container = await new GenericContainer("cristianrgreco/testcontainer:1.1.14").withExposedPorts(8080).start();
+    const currentYear = new Date().getFullYear().toString();
+
+    const stream = await container.logs({timestamps: true});
```
mdelapenya, line 33: Here you can see what I meant in my previous comment: if users pass
```
    const stream = await container.logs({timestamps: true, follow: false});
```

Will the false value be overridden by https://github.com/testcontainers/testcontainers-node/pull/670/files#diff-cde64c872b844fe1a7e8949d7eee46cea0cb346b40d840597c1a4785d68998d2R162?

mdelapenya (commented):
I added a few comments regarding the new opts variable.

Thanks!

mdelapenya:
> @mdelapenya Can you also publish new library version to npm after my PR will be merged, please?

Hey @NuShoSinkuPomogliTebeTvoiSankcii I'm the maintainer of testcontainers-go, helping out with the maintenance of the NodeJS flavour while @cristianrgreco is on PTO. I can help with the big picture of the library, where @javierlopezdeancos can support me with the Typescript specifics.

Said that, I'm afraid that we are going to slow down a bit the release cadence until we get more familiar with the codebase. But hopefully soon enough to see your changes released, for sure. So please be patient with us 🙏

packages/testcontainers/src/generic-container/generic-container-logs.test.ts
NuShoSinkuPomogliTebeTvoiSankcii: Yes, it will be overridden inside library and will be true

NuShoSinkuPomogliTebeTvoiSankcii: Also it is already skipping opts passed to this method, because it's using only `since` parameter – [here](https://github.com/testcontainers/testcontainers-node/blob/main/packages/testcontainers/src/container-runtime/clients/container/docker-container-client.ts#L164)

It's not so easy as I thought firstly :)

@javierlopezdeancos I think easiest way will be to leave it as it now in [docker-container-client.ts](https://github.com/testcontainers/testcontainers-node/pull/670/files/992c3112c415b879cf151c29b2556478b9d8be1a#diff-cde64c872b844fe1a7e8949d7eee46cea0cb346b40d840597c1a4785d68998d2).  What do you think about that?

packages/testcontainers/src/container-runtime/clients/container/docker-container-client.ts
```diff
@@ -158,6 +158,7 @@ export class DockerContainerClient implements ContainerClient {
...
         follow: true,
```
javierlopezdeancos, line 162: But, if we are doing this now

```typescript
const stream = (await container.logs({
        ...opts,
        follow: true,
        stdout: true,
        stderr: true,
        since: opts?.since ?? 0,
})) as IncomingMessage;
```

at the end, like we are applying  this property setting: after destructuring opts:

````typescript
 follow: true,
 stdout: true,
 stderr: true,
 since: opts?.since ?? 0
 ````
 
 we can pass for example the follow as false and he will be set always to true so, probably you want implement this in a way that allow passing any option with any value the user wants?
 
 You can review the snippet https://jsbin.com/gofedomanu/edit?js,console to take a look the behavior that I'm talking about.

File: packages/testcontainers/src/generic-container/generic-container-logs.test.ts (hunk 2 above)
Comment (javierlopezdeancos) [reply packages/testcontainers/src/generic-container/generic-container-logs.test.ts:33]: To not breking changes I suggest to this

```javascript
const myOpts = {
  foo: "foo",
  bar: "bar"
}

const getOpts = (opts) => {
  return  {
    ...opts,
    foo: false
  }
}

console.log(getOpts(myOpts))

const getOptsOverrited = (opts) => {
  return  {
    ...opts,
    foo: opts?.foo ?? false
  }
}

console.log(getOptsOverrited(myOpts))
```

Review here the snippet https://jsbin.com/gofedomanu/edit?js,console

That in our code should be maintain the since pattern to each prop that we should set to the default state

```typescript
 try {
      log.debug(`Fetching container logs...`, { containerId: container.id });
      const stream = (await container.logs({
        ...opts,
        stdout: opts?.stdout ?? true,
        stderr: opts?.stderr ?? true,
        since: opts?.since ?? 0,
      })) as IncomingMessage;
      stream.socket.unref();
      const demuxedStream = this.demuxStream(container.id, stream);
      log.debug(`Fetched container logs`, { containerId: container.id });
      return demuxedStream;
    }
```

javierlopezdeancos (changes requested):

NuShoSinkuPomogliTebeTvoiSankcii:
@javierlopezdeancos I reverted my changes with overriding options. Just passing options for since parameter – actually it's the only thing that I need for now.

Also added test for this behaviour. Review it please. 🙏

packages/testcontainers/src/container-runtime/clients/container/docker-container-client.test.ts
```diff
@@ -0,0 +1,22 @@
+import Dockerode from "dockerode";
```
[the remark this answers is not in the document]
NuShoSinkuPomogliTebeTvoiSankcii: done

packages/testcontainers/src/generic-container/generic-container-logs.test.ts
NuShoSinkuPomogliTebeTvoiSankcii: reverted to previous version

packages/testcontainers/src/container-runtime/clients/container/docker-container-client.ts
NuShoSinkuPomogliTebeTvoiSankcii: reverted to previous version

packages/testcontainers/src/container-runtime/clients/container/docker-container-client.ts
NuShoSinkuPomogliTebeTvoiSankcii: reverted to previous version

NuShoSinkuPomogliTebeTvoiSankcii:
@javierlopezdeancos I'm sorry for disturbing, but I will also be on leave (as @cristianrgreco 😄) after Friday for 2 weeks, so it will be nice to merge my PR before that date.

cristianrgreco:
Hi @NuShoSinkuPomogliTebeTvoiSankcii, thanks for raising the PR. I'd prefer not to expose the underlying Docker API (dockerode) in testcontainers' public API. This way we do not proxy breaking changes from dockerode when they happen. I see from the test you want to add support for fetching logs since a certain time, so let's implement support just for that and update the docs to explain how to use it.

If it is low level access to the Docker client you need, this PR will give you what you're after: https://github.com/testcontainers/testcontainers-node/pull/644

NuShoSinkuPomogliTebeTvoiSankcii:
@cristianrgreco Hi! Done refactoring and docs.

About low level access – I understand the problem with breaking changes. 
But for me it would be nice to have both low level dockerode API and high level tc-node at the same time through the same JS object. In your example it's produce more boilerplate with storing many different instances of the same object.

For example this PR. In my project I had to copy-paste [this logic with dockerode interactions](https://github.com/testcontainers/testcontainers-node/blob/ec59ed862c7915ef5a6cdd5ce1a9d4105bb6de3f/packages/testcontainers/src/container-runtime/clients/container/docker-container-client.ts#L160) just to pass since parameter directly to dockerode container logs method.
If someone will need to pass tail/until args (from `ContainerLogsOptions`) to StartedContainer they will also need to extend options interface that I just created. Or create another container instance through `container-runtime-client` just for access logs with some extra options. It's not so convenient.

P.S. Glad to have you back!

javierlopezdeancos (approved):

cristianrgreco:
Thanks for the response @NuShoSinkuPomogliTebeTvoiSankcii. 

I think I understand your points, but I think this result is actually ideal: You identified a use case for fetching logs since a given time, you've raised a well documented PR with this behaviour, where a conflict when upgrading Dockerode results in a compile time error, which ensures we don't accidentally introduce breaking changes. IMO this is a win.

If you think supporting tail/until is a common use case then please feel free to raise more PRs, but to date this has not been asked for. If raising a PR is an inconvenience then adding a couple of lines of code to get the underlying Docker client I think is a compromise worthwhile for not having breaking changes.

The PR LGTM, than you for making the changes. I'll get this deployed shortly.

> P.S. Glad to have you back!

Thanks 🙂

PR #671, opened 2023-11-13
Add support for building a container from context archive

LNSD:
This PR adds support for building a container image from a tar archive stream (i.e., `NodeJs.ReadableStream`). 

> **Note**
> This functionality is already existing in _testcontainers-go_: [dynamic-build-context](https://golang.testcontainers.org/features/build_from_dockerfile/#dynamic-build-context)

This covers the case where one wants to programmatically create a dynamic `Dockerfile` and context (e.g., using a template engine) without the overhead of writing to disk.

```javascript
const tar = require('tar-stream');

const tarStream = tar.pack();
tarStream.entry({ name: 'alpine.Dockerfile' }, 
  `
    FROM alpine:latest
    CMD ["sleep", "infinity"]
  `
);
tarStream.finalize();

const container = await GenericContainer
  .fromContextArchive(tarStream, 'alpine.Dockerfile')
  .build();
```

### DIFF
--- a/packages/testcontainers/package.json
+++ b/packages/testcontainers/package.json
@@ -55,6 +55,8 @@
     "@types/proper-lockfile": "^4.1.3",
     "@types/properties-reader": "^2.1.2",
     "@types/tar-fs": "^2.0.3",
-    "@types/tmp": "^0.2.5"
+    "@types/tar-stream": "^3.1.3",
+    "@types/tmp": "^0.2.5",
+    "tar-stream": "^3.1.6"
   }
 }
--- a/packages/testcontainers/src/container-runtime/clients/image/docker-image-client.ts
+++ b/packages/testcontainers/src/container-runtime/clients/image/docker-image-client.ts
@@ -1,14 +1,11 @@
 import Dockerode, { ImageBuildOptions } from "dockerode";
 import byline from "byline";
-import tar from "tar-fs";
-import path from "path";
-import { existsSync, promises as fs } from "fs";
-import dockerIgnore from "@balena/dockerignore";
 import { getAuthConfig } from "../../auth/get-auth-config";
 import { ImageName } from "../../image-name";
 import { ImageClient } from "./image-client";
 import AsyncLock from "async-lock";
 import { log, buildLog, pullLog } from "../../../common";
+import stream from "stream";
 
 export class DockerImageClient implements ImageClient {
   private readonly existingImages = new Set<string>();
@@ -16,23 +13,12 @@ export class DockerImageClient implements ImageClient {
 
   constructor(protected readonly dockerode: Dockerode, protected readonly indexServerAddress: string) {}
 
-  async build(context: string, opts: ImageBuildOptions): Promise<void> {
+  async build(context: stream.Readable, opts: ImageBuildOptions): Promise<void> {
     try {
-      log.debug(`Building image "${opts.t}" with context "${context}"...`);
-      const isDockerIgnored = await this.createIsDockerIgnoredFunction(context);
-      const tarStream = tar.pack(context, {
-        ignore: (aPath) => {
-          const relativePath = path.relative(context, aPath);
-          if (relativePath === opts.dockerfile) {
-            return false;
-          } else {
-            return isDockerIgnored(relativePath);
-          }
-        },
-      });
+      log.debug(`Building image "${opts.t}"...`);
       await new Promise<void>((resolve) => {
         this.dockerode
-          .buildImage(tarStream, opts)
+          .buildImage(context, opts)
           .then((stream) => byline(stream))
           .then((stream) => {
             stream.setEncoding("utf-8");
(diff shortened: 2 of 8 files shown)


packages/testcontainers/src/generic-container/generic-container-builder.ts
```diff
@@ -105,3 +138,17 @@ export class GenericContainerBuilder {
       .reduce((prev, next) => ({ ...prev, ...next }), {} as RegistryConfig);
   }
 }
+
+async function newDockerignoreFilter(context: string): Promise<(path: string) => boolean> {
```
javierlopezdeancos, line 142: Only to curiosity in your approach, why `newDockerignoreFilter` function  is not part from `class GenericContainerBuilder` as a private method take into account that you are only using it to this class responsabilities?

javierlopezdeancos:
After reviewing the PR I think that for me it is LGTM and approved.

@LNSD Thanks for the explanations about **how the docker and dockerode api works to build an image with a dockerfile (_recipe_) and a files source (_ingredients_ aka context)**, either through a path on our disk or through an external data source, In both cases,  we have to package these files into a tar file and create a stream from it which is as the docker api requires in its context.

The feature itself is very powerful, being able to programmatically create a dynamic dockerfile and feed the stream with ingredients that do not exactly belong to our disk, a use case, an AWS stream.

I see the refactor that you include to allow the feature appropriate, promoting the creation of the source tar file which we will create the stream from.

packages/testcontainers/src/generic-container/generic-container-builder.ts
```diff
@@ -44,6 +48,12 @@ export class GenericContainerBuilder {
     return this;
   }
 
+  /**
```
javierlopezdeancos, line 51: I really really really really appreciate these comments for our methods and functions but, since they are not anywhere in the library, I don't know if it would be convenient to start doing it. 🤔

LNSD:
I am putting this PR on hold until #668 is merged.

packages/testcontainers/src/container-runtime/clients/image/docker-image-client.ts
```diff
@@ -1,38 +1,24 @@
...
 export class DockerImageClient implements ImageClient {
   private readonly existingImages = new Set<string>();
   private readonly imageExistsLock = new AsyncLock();
 
   constructor(protected readonly dockerode: Dockerode, protected readonly indexServerAddress: string) {}
 
-  async build(context: string, opts: ImageBuildOptions): Promise<void> {
+  async build(context: stream.Readable, opts: ImageBuildOptions): Promise<void> {
     try {
-      log.debug(`Building image "${opts.t}" with context "${context}"...`);
```
cristianrgreco, line 21: What's the reasoning behind moving this logic to the `GenericContainerBuilder`? I like the idea that the `DockerImageClient` can be used standalone and supports `.dockerignore` files out of the box, this is no longer the case

LNSD: I understand your point.

I decided to raise the filtering to a higher abstraction level, recognizing that the `.dockerignore` filtering is a specific instance of passing a tar stream to the container runtime client. This approach streamlines the code and renders the `DockerImageClient` agnostic to the underlying filesystem, enhancing its versatility.

At the time of implementation, I faced two options:

1. Introduce a method in `GenericContainerBuilder` and `DockerImageClient` for passing the stream directly to the container runtime client without `.dockerignore` filtering. This would create an additional flow in both classes.

2. Maintain the `DockerImageClient` agnostic to the filesystem (not utilizing `tar-fs`) and keep the API straightforward by passing a stream to the container runtime client. Since the filtered stream is a subset of the passed stream, this approach avoids needing two separate APIs in both the client and the `GenericContainerBuilder` class.

packages/testcontainers/src/container-runtime/clients/image/docker-image-client.ts
```diff
@@ -1,38 +1,24 @@
...
-import tar from "tar-fs";
-import path from "path";
-import { existsSync, promises as fs } from "fs";
-import dockerIgnore from "@balena/dockerignore";
 import { getAuthConfig } from "../../auth/get-auth-config";
 import { ImageName } from "../../image-name";
 import { ImageClient } from "./image-client";
 import AsyncLock from "async-lock";
 import { log, buildLog, pullLog } from "../../../common";
+import stream from "stream";
```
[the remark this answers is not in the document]
LNSD: Yes, it is. I can change the import name with `node:stream`, but I don't know the difference.

packages/testcontainers/src/generic-container/generic-container-builder.ts
LNSD: I believe class methods should access or modify the object instance. 

In this case, `newDockerignoreFilter` does not access nor modify any object fields. It can be seen as a factory method. It is a function that, given a context path, returns a `.dockerignore` based filter function.

packages/testcontainers/package.json
```diff
@@ -55,6 +55,8 @@
     "@types/proper-lockfile": "^4.1.3",
     "@types/properties-reader": "^2.1.2",
     "@types/tar-fs": "^2.0.3",
-    "@types/tmp": "^0.2.5"
+    "@types/tar-stream": "^3.1.3",
+    "@types/tmp": "^0.2.5",
+    "tar-stream": "^3.1.6"
```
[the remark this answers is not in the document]
LNSD: I only used it in the tests to generate a test tar stream. It is not used in the package's production code.

PR #672, opened 2023-11-13
Add DynamoDB module

tenlisboa:
Motivated by the need for e2e testing using DynamoDB outside a serverless infraestructure inside a LocalStack.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -53,6 +53,650 @@
         "node": ">=6.0.0"
       }
     },
+    "node_modules/@aws-crypto/crc32": {
+      "version": "3.0.0",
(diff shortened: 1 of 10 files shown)


cristianrgreco:
Thanks for raising @tenlisboa! Could you please also add a page in the docs and add the module to be tested here https://github.com/testcontainers/testcontainers-node/blob/2fa66156bc5f21e2754be87ac15aced66f729f1e/.github/workflows/test.yml#L83-L93

cristianrgreco:
Localstack module has been added which supports DynamoDB and more. Thanks again @tenlisboa for the contribution

PR #675, opened 2023-11-15
Retire Node 16 (End-of-life) from CI

LNSD:
### Context

Node 16 is End-of-life since September 11th, 2023 (see [EOL statement](https://nodejs.org/en/blog/announcements/nodejs16-eol)).

schedule
<sup>Source: https://nodejs.org/en/about/previous-releases#release-schedule</sup>

Based on the diagram above, the only LTS versions are **Node 18** and **Node 20**.

### Changes description

- [x] Remove Node 16.x from the GitHub actions CI's test matrix.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -25,7 +25,7 @@ jobs:
     strategy:
       matrix:
         runner: [ ubuntu-22.04 ]
-        node-version: [ 16.x, 18.x, 20.x ]
+        node-version: [ 18.x, 20.x ]
     steps:
       - name: Code checkout
         uses: actions/checkout@v3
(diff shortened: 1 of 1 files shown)


kiview (approved):

mdelapenya (approved):
LGTM!

I have one question: Is it convenient to add Node 21 to the matrix?

LNSD:
> LGTM!
> 
> I have one question: Is it convenient to add Node 21 to the matrix?

Typically, projects test only against LTS versions. Node 21 has a "short lifespan" (temptatively, EOL April 2024), for this reason most of the developers avoid using this intermediate versions in production.

So, IMHO, I wouldn't add it, given that Node 22 (LTS) will be out in few months.

javierlopezdeancos:
I have a doubt here, we should point our local dev stack to node 18x at least? we couldn't find unexpected catches in ci if we don't do this?

Issue #659, opened 2023-10-04
Default to compose v2


jalaziz:
It seems that the current version of `testcontainers` [first attempts to use compose V1, then tries V2](https://github.com/testcontainers/testcontainers-node/blob/73d3a2d73aad62b04d37b9c4e959895a467705a1/packages/testcontainers/src/container-runtime/clients/compose/compose-client.ts#L29-L45).

In environments where both V1 and V2 are installed (e.g. GitHub Actions), this will always default to V1 with no easy option to use the newer version.

Since V1 is deprecated, `testcontainers` should probably default to V2. Alternatively, it would be great if there was a way to override the default.

daniel-glazer:
hey, looks like a good first issue to me.. can I take it? 
can you share a link to the guide on how to contribute? @cristianrgreco

PR #681, opened 2023-11-26
Switch docker compose defaults from v1 to v2

danielglazer:
Initial work to fix [the issue mentioned here](https://github.com/testcontainers/testcontainers-node/issues/659).
Since V1 is deprecated, testcontainers should probably default to V2. could probably override the defaults if we pass it through the process env but not sure if its the right way. 
couldn't find related tests that failed because of this change or documentation related change needed.. @cristianrgreco can you help with the review / guide me on further changes?

### DIFF
--- a/packages/testcontainers/src/container-runtime/clients/compose/compose-client.ts
+++ b/packages/testcontainers/src/container-runtime/clients/compose/compose-client.ts
@@ -29,14 +29,14 @@ export async function getComposeClient(environment: NodeJS.ProcessEnv): Promise<
 async function getComposeInfo(): Promise<ComposeInfo | undefined> {
   try {
     return {
-      version: (await dockerComposeV1.version()).data.version,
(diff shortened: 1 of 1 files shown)


cristianrgreco:
Thanks for raising the PR @danielglazer 

---

> could probably override the defaults if we pass it through the process env but not sure if its the right way.

Agreed I don't think it's necessary for now

> couldn't find related tests that failed because of this change or documentation related change needed..

Indeed we don't test the docker-compose version because it'd be a pain to need both version installed for tests to pass. The GHA runners have both versions of compose installed. I've also checked locally and it's working as expected.

>  @cristianrgreco can you help with the review / guide me on further changes?

LGTM, I'm just thinking whether or not this warrants a major change. After that I'll let you know once it's released

cristianrgreco (approved):

PR #691, opened 2023-12-12
feat(testcontainers): get existing network by id or name

LNSD:
# Context

The current implementation only supports passing a testcontainers-generated network instance to the `GenericContainer` class. This is because the `withNetwork` method only accepts `StartedNetwork` instances. And the `StrtedNetwork` class is not instantiable.

```ts
const network = await new Network().start();

const container = await new GenericContainer("cristianrgreco/testcontainer:1.1.14").withNetwork(network).start();
```

# Use cases

* **Scenario:** _Retrieve and Attach Container to a Pre-existent Network_

  The pre-existent network could be a network created by docker-compose.

  ```gherkin
  Given a pre-existent container network identified by name or id
  When a user retrieves the container network
  Then the network should be available for attaching containers
  And the retrieved network instance should not be stoppable
  ```

# Implementation details

The following implementation extends the `Network` facade class and introduces the `StartedExistingNetwork` class covering this scenario.

Leveraging typescript interfaces and applying the _Interface segregation_ SOLID principle:

- The `StartedNetwork` class is renamed to `StartedBridgedNetwork` to avoid clashing with the new interface name.
- Two new interfaces are introduced: `StartedNetwork` and `StoppableNetwork`. 
- The `StartedBridgedNetwork` (previously `StartedNetwork`) implements the `StartedNetwork` and `StoppableNetwork` interfaces.
- The new `StartedExistingNetwork` class only implements the `StartedNetwork` interface since it is not desirable to stop and remove a pre-existing network.

With this approach, both classes can be passed to the generic container's `withNetwork()` method since both implement the `StartedNetwork` interface.

### DIFF
--- a/packages/modules/selenium/src/selenium-container.ts
+++ b/packages/modules/selenium/src/selenium-container.ts
@@ -8,7 +8,7 @@ import {
   GenericContainer,
   log,
   Network,
-  StartedNetwork,
+  StoppableNetwork,
   StartedTestContainer,
   StopOptions,
   StoppedTestContainer,
@@ -76,10 +76,6 @@ export class StoppedSeleniumContainer extends AbstractStoppedContainer {
 }
 
 export class SeleniumRecordingContainer extends SeleniumContainer {
-  constructor(image: string) {
-    super(image);
-  }
-
   public override async start(): Promise<StartedSeleniumRecordingContainer> {
     const network = await new Network().start();
     this.withNetwork(network);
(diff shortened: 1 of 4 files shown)


cristianrgreco:
Hi @LNSD, thanks for the PR!

> The new StartedExistingNetwork class only implements the StartedNetwork interface since it is not desirable to stop and remove a pre-existing network.

Because of this it seems this PR would need a breaking change. There are a couple of other breaking changes which I'm not sure are required:

1. Return type of `Network.start()` changed from `StartedNetwork` to `StartedBridgeNetwork`. The following code will error because `stop` is no longer a method on the `StartedNetwork`:

```ts
const network: StartedNetwork = await Network.start();
await network.stop();
```

2. Return type of `Network.stop()` has changed from `StoppedNetwork` to `void` . The following code will error:

```ts
const stoppedNetwork: StoppedNetwork = await network.stop();
```

LNSD:
> I know that the `withNetworkMode, ' when passed the network name, connects the container to the network at the time of creation. However, this does not work when one wants to attach a container to multiple networks (at this moment, as far as I know, this is not supported by the generic container class).

This is out of the scope of this PR.

However, it is a limitation of the current implementation that I would like to address in subsequent PR.

Issue #701, opened 2024-01-03
[Enhancement] LocalstackContainer API enhancement


lbroudoux:
Hello,

Thanks for the great work out there!

I've used the recently released `LocalstackContainer` in the [Microcks.io testcontainers module](github.com/microcks/microcks-testcontainers-node) and it works like a charm. However, as a user of the [Java version](https://github.com/testcontainers/testcontainers-java/blob/main/modules/localstack/src/main/java/org/testcontainers/containers/localstack/LocalStackContainer.java), I found the module API less easy to use if you're not an experienced user and know the internals of Localstack.

Typically, the below functions/methods are very convenient in the Java version:
* `withServices()` method to select AWS services,
* `getRegion()` method to get the default Localstack region,
* `getAccessKey()` /  `getSecretKey()` method to get the default credentials,
* `getEndpointOverride(service)` to get the service-specific endpoint URL.

It would be awesome to have a similar interface with those methods in the Node/Typescript world!
What do you think?

cristianrgreco:
Hi @lbroudoux, thanks for the kind words, and yes sounds great to me!

lbroudoux:
Hi! I reviewed the code this morning to see if I can submit a PR.

If my understanding is ok, it looks like we're missing a way to retrieve env vars from a `StartedTestContainer`. `GenericContainer.createOpts.Env` is not transferred to the started container, and `inspectResult` is not directly usable.

I think we'd need to add a `public getEnvironment(): Record<string, string>` method to the `StartedTestContainer` interface so that we'll be able to retrieve region, access key and secret key - for this localstack use-case. I've made the addition, and it works well.

As this change is not local to localstack use-case, would you like to have a specific issue + PR on this?

lbroudoux:
Reading the Localstack doc thoroughly, I can see that setting the list of services is no longer required, as services are now all available on the same exposed port. So you don't need to replicate the `withServices()` and `getEndpointOverride(service)` methods!

eddumelendez:
my 2 cents here, `withServices()` is not really needed because the `SERVICES` env var can be used directly. `SERVICES` env var was required until LocalStack 0.13, after that services are lazy loaded. Same with `getEndpointOverride(service)`, until LocalStack 0.11 returns a endpoint with different port by service but that's not the case anymore.

PR #640, opened 2023-08-16
Add Localstack module

ilopezluna:

PR #702, opened 2024-01-05
LocalstackContainer API enhancement

lbroudoux:
As discussed in #701

* Align Localstack module API on Java version
* Add environment variables retrieval after container startup

Thank you!

Signed-off-by: Laurent Broudoux <laurent.broudoux@gmail.com>

### DIFF
--- a/packages/modules/localstack/src/localstack-container.test.ts
+++ b/packages/modules/localstack/src/localstack-container.test.ts
@@ -69,4 +69,18 @@ describe("LocalStackContainer", () => {
     await container.stop();
     await awsCliInDockerNetwork.stop();
   });
+
+  it("should use access environment variables and default creds", async () => {
+    const container = await new LocalstackContainer()
+      .withEnvironment({
+        DEFAULT_REGION: 'eu-west-3'
+      })
+      .start();
+
+    expect(container.getRegion()).toBe('eu-west-3');
+    expect(container.getAccessKey()).toBe('test');
+    expect(container.getSecretKey()).toBe('test');
+
+    await container.stop();
+  });
 });
--- a/packages/modules/localstack/src/localstack-container.test.ts
+++ b/packages/modules/localstack/src/localstack-container.test.ts
@@ -73,13 +73,13 @@ describe("LocalStackContainer", () => {
   it("should use access environment variables and default creds", async () => {
     const container = await new LocalstackContainer()
       .withEnvironment({
-        DEFAULT_REGION: 'eu-west-3'
+        DEFAULT_REGION: "eu-west-3",
       })
       .start();
 
-    expect(container.getRegion()).toBe('eu-west-3');
-    expect(container.getAccessKey()).toBe('test');
-    expect(container.getSecretKey()).toBe('test');
+    expect(container.getRegion()).toBe("eu-west-3");
+    expect(container.getAccessKey()).toBe("test");
+    expect(container.getSecretKey()).toBe("test");
 
     await container.stop();
   });
--- a/packages/modules/localstack/src/localstack-container.ts
+++ b/packages/modules/localstack/src/localstack-container.ts
@@ -2,6 +2,10 @@ import { AbstractStartedContainer, GenericContainer, log, StartedTestContainer, 
 
 export const LOCALSTACK_PORT = 4566;
 
+const DEFAULT_REGION = 'us-east-1';
+const DEFAULT_AWS_ACCESS_KEY_ID = 'test';
+const DEFAULT_AWS_SECRET_ACCESS_KEY = 'test';
+
 export class LocalstackContainer extends GenericContainer {
   constructor(image = "localstack/localstack:2.2.0") {
     super(image);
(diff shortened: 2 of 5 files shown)


packages/modules/localstack/src/localstack-container.test.ts
```diff
@@ -69,4 +69,18 @@ describe("LocalStackContainer", () => {
     await container.stop();
     await awsCliInDockerNetwork.stop();
   });
+
+  it("should use access environment variables and default creds", async () => {
+    const container = await new LocalstackContainer()
+      .withEnvironment({
+        DEFAULT_REGION: "eu-west-3",
```
cristianrgreco, line 76: Where does the `DEFAULT_REGION` convention come from? I had a quick look at the docs and found https://docs.localstack.cloud/references/configuration/#legacy, which says that it's no longer supported.

packages/modules/localstack/src/localstack-container.ts
```diff
@@ -50,4 +54,30 @@ export class StartedLocalStackContainer extends AbstractStartedContainer {
...
+   * Provides a default region that is preconfigured to communicate with a given simulated service.
+   * @returns A default region
+   */
+  public getRegion(): string {
+    return this.startedTestContainer.getEnvironment()["DEFAULT_REGION"] || DEFAULT_REGION;
+  }
+
+  /**
+   * Provides a default access key that is preconfigured to communicate with a given simulated service.
+   * <a href="https://github.com/localstack/localstack/blob/master/doc/interaction/README.md?plain=1#L32">AWS Access Key</a>
```
cristianrgreco, line 68: I'm getting a 404 for these URLs. I also couldn't find reference to these environment variables in the LS config

packages/modules/localstack/src/localstack-container.test.ts
```diff
@@ -69,4 +69,18 @@ describe("LocalStackContainer", () => {
...
+      })
+      .start();
+
+    expect(container.getRegion()).toBe("eu-west-3");
+    expect(container.getAccessKey()).toBe("test");
+    expect(container.getSecretKey()).toBe("test");
```
cristianrgreco, line 82: The container is mapping the environment variables to these getter methods, but this test is not checking that LS is actually using them. From the deprecation notices in the docs I get the impression they're not being used.

packages/modules/localstack/src/localstack-container.ts
lbroudoux: Sorry, I blindly copy/paste here the comments coming from testcontainers-java/localstack

packages/modules/localstack/src/localstack-container.test.ts
lbroudoux: Yes, I think you're right. Many things have changed with 3.x and I should have refreshed my knowledge. The point is the default image is still `localstack/localstack:2.2.0`. As multi-region has been introduced incrementally, this may not affect the module, but it may be nice to upgrade to 3.x to definitely get rid of this question. What do you think?

packages/modules/localstack/src/localstack-container.test.ts
lbroudoux: Yes, I think you're right... I see nothing in current documentation referencing these... Here again, I tried to reproduce the behavior of the test containers-java/local stack module, but this may be inappropriate.

lbroudoux:
I've done some other client tests with the `amazon/aws-cli:2.7.27` image, and it seems that whatever the credentials - we should have at least `AWS_ACCESS_KEY_ID` and `AWS_SECRET_ACCESS_KEY` env var set but they could be totally random - things are working! This is because IAM enforcement is turned off by default. 

So I tend to think this PR (and related issue) is, in fact, no longer needed with the recent version of localstack.

cristianrgreco:
Sounds good to me @lbroudoux, shall we close this PR and raise another for upgrading the LS to container to v3?

djakielski:
Hey @lbroudoux could you also fix missing reference in [mkdocs.yml](https://github.com/testcontainers/testcontainers-node/blob/main/mkdocs.yml) to add localstack to the documentation? It was forgotten in PR https://github.com/testcontainers/testcontainers-node/pull/640. 

@cristianrgreco or should I create a separate PR with this small fix for you?

```yaml
 - Modules:
      - ....
      - Localstack: modules/localstack.md
```

cristianrgreco:
> Hey @lbroudoux could you also fix missing reference in [mkdocs.yml](https://github.com/testcontainers/testcontainers-node/blob/main/mkdocs.yml) to add localstack to the documentation? It was forgotten in PR #640.
> 
> @cristianrgreco or should I create a separate PR with this small fix for you?
> 
> ```yaml
>  - Modules:
>       - ....
>       - Localstack: modules/localstack.md
> ```

If you could that would be much appreciated!

djakielski:
> > Hey @lbroudoux could you also fix missing reference in [mkdocs.yml](https://github.com/testcontainers/testcontainers-node/blob/main/mkdocs.yml) to add localstack to the documentation? It was forgotten in PR #640.
> > @cristianrgreco or should I create a separate PR with this small fix for you?
> > ```yaml
> >  - Modules:
> >       - ....
> >       - Localstack: modules/localstack.md
> > ```
> 
> If you could that would be much appreciated!

@ cristianrgreco Done, see https://github.com/testcontainers/testcontainers-node/pull/710

lbroudoux:
Sorry I was away from keyboard these last days. I think we can close that PR as unsolved and I'll submit another one tomorrow to upgrade to localstack latest 3.x.

PR #712, opened 2024-02-01
Move default container configurations to constructor to allow for overriding

### LINKED ISSUE #711: Wait strategy is always overwritten on start
**Expected Behaviour**
A manually set wait strategy won't be overwritten on start.

**Actual Behaviour**
On start, a manually set the wait strategy will be overwritten with `withWaitStrategy(testcontainers_1.Wait.forLogMessage("Ready", 1))`.

cristianrgreco:
Fixes #711

### DIFF
--- a/packages/modules/localstack/src/localstack-container.ts
+++ b/packages/modules/localstack/src/localstack-container.ts
@@ -5,6 +5,8 @@ export const LOCALSTACK_PORT = 4566;
 export class LocalstackContainer extends GenericContainer {
   constructor(image = "localstack/localstack:2.2.0") {
     super(image);
+    this.resolveHostname();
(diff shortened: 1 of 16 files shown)


packages/modules/localstack/src/localstack-container.ts
```diff
@@ -5,6 +5,8 @@ export const LOCALSTACK_PORT = 4566;
 export class LocalstackContainer extends GenericContainer {
   constructor(image = "localstack/localstack:2.2.0") {
     super(image);
+    this.resolveHostname();
```
joebowbeer, line 8: @cristianrgreco This line should not have been moved. There is no way to provide an environment before the constructor is called so `resolveHostname` will never resolve `LOCALSTACK_HOST`

joebowbeer:
Observation: No tests were affected by these changes, indicating that this functionality is not currently tested.

PR #714, opened 2024-02-15
Fix custom image log line

omerlh:
A small fix for this log message:

```
Building Dockerfile "<>" as image "[object Object]"...
```


cristianrgreco:
`ImageName` is an object, it'd be better to log `imageName.string` instead of JSON.stringifying all of it

omerlh:
Sure, fixed!

PR #723, opened 2024-03-07
Support Kraft mode for Kafka container

silh:
Kraft is out of beta for some time now and and soon zookeeper support will be completely removed so it's useful to prepare for kraft support. It was also possible to use testcontainers-java without Zookeeper since 1.18.0.

This PR ports the solution to enable usage of Kafka without Zookeeper from testcontainers-java.

Unrelated changes:
- fixed `copyContentToContainer` not setting the correct mode for the file inside the container.

### DIFF
--- a/packages/modules/kafka/src/kafka-container.ts
+++ b/packages/modules/kafka/src/kafka-container.ts
@@ -11,6 +11,10 @@ import {
 const KAFKA_PORT = 9093;
 const KAFKA_BROKER_PORT = 9092;
 const DEFAULT_ZOOKEEPER_PORT = 2181;
+const DEFAULT_CLUSTER_ID = "4L6g3nShT-eMCtK--X86sw";
+
+// https://docs.confluent.io/platform/7.0.0/release-notes/index.html#ak-raft-kraft
+const MIN_KRAFT_VERSION = "7.0.0";
 
 export const KAFKA_IMAGE = "confluentinc/cp-kafka:7.2.2";
 
@@ -36,10 +40,16 @@ interface PKCS12CertificateStore {
   passphrase: string;
 }
 
+enum KafkaMode {
+  EMBEDDED_ZOOKEEPER,
+  PROVIDED_ZOOKEEPER,
+  KRAFT,
+}
+
 export class KafkaContainer extends GenericContainer {
   private readonly uuid: Uuid = new RandomUuid();
 
-  private isZooKeeperProvided = false;
+  private mode = KafkaMode.EMBEDDED_ZOOKEEPER;
   private zooKeeperHost?: string;
   private zooKeeperPort?: number;
   private saslSslConfig?: SaslSslListenerOptions;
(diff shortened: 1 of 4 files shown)


silh:
Hey @cristianrgreco , fixed all the problems. Now the PR is ready for review.

cristianrgreco:
Thanks @silh, could you please also add an example to the Kafka docs showing how to use Kraft with the KafkaContainer?

silh:
@cristianrgreco done 👍

packages/modules/kafka/src/kafka-container.ts
```diff
@@ -83,35 +112,83 @@ export class KafkaContainer extends GenericContainer {
...
+    // Run the original command
+    command += "/etc/confluent/docker/run \n";
+    container.copyContentToContainer([{ content: command, target: STARTER_SCRIPT, mode: 0o777 }]);
+
+    const client = await getContainerRuntimeClient();
+    const dockerContainer = client.container.getById(container.getId());
+    const boundPorts = BoundPorts.fromInspectResult(client.info.containerRuntime.hostIps, inspectResult).filter(
+      this.exposedPorts
+    );
+    await waitForContainer(client, dockerContainer, this.originalWaitinStrategy, boundPorts);
```
cristianrgreco, line 189: This is different, to manually invoke a wait strategy after a container is said to have started. Ideally we should get this working just by setting `this.waitStrategy` and nothing else. E.g would it work if on line 139 we do 
```ts
this.waitStrategy = Wait.forAll([
  this.waitStrategy, 
  Wait.forLogMessage(WAIT_FOR_SCRIPT_MESSAGE)
]);
```

packages/modules/kafka/src/kafka-container.ts
```diff
@@ -83,35 +112,83 @@ export class KafkaContainer extends GenericContainer {
...
+      if (this.isLessThanCP(7, 4)) {
+        command += this.commandKraft();
+      }
+    } else if (this.mode === KafkaMode.EMBEDDED_ZOOKEEPER) {
+      command += this.commandZookeeper();
+    }
+
+    // Run the original command
+    command += "/etc/confluent/docker/run \n";
+    container.copyContentToContainer([{ content: command, target: STARTER_SCRIPT, mode: 0o777 }]);
```
cristianrgreco, line 182: This probably needs to be awaited

silh: Fixed.

packages/modules/kafka/src/kafka-container.ts
silh: This won't work unfortunately.

The main problem is we need to gather some info which is only available in when container is started and copy it to the container. Unlike the java implementation, some of this methods are not available on not-started container. It can be achieved, but the code will be messier and will have some duplicated functionality. I'll try to prepare a commit with that implementation.

cristianrgreco: OK, we can look at this some other time. For now I think it's more important to add support for Kraft

PR #728, opened 2024-03-14
Fix node 20 pipeline 

### LINKED ISSUE #727: Install node dependencies action are not taking right node js version
**Expected Behaviour**
Install node js dependency should install the right node js version per action description

**Actual Behaviour**
Install node js dependency to Node 20.x is installing  Node 18.x version

**Testcontainer Logs**
What is happening 
Captura de pantalla 2024-03-14 a las 17 47 12
 
What should be
Captura de pantalla 2024-03-14 a las 17 43 25

**Steps to Reproduce**
1. Run any test job

**Environment Information**
- Operating System: ci
- Docker Version: ci
- Node version: 20.x
- Testcontainers version: latest

javierlopezdeancos:
Install Node js dependency action should install the right Node js version per action description

We can see it in each job that execute test with different Node js versions
imagen

I think that there is an error in the job github template, fixing it we have the expected behavior, installing the right node version
imagen

Close #727

### DIFF
--- a/.github/workflows/test-template.yml
+++ b/.github/workflows/test-template.yml
@@ -133,7 +133,7 @@ jobs:
       - name: Install NodeJS ${{ inputs.node-version }}
         uses: actions/setup-node@v3
         with:
-          node-version: ${{ matrix.node-version }}
+          node-version: ${{ inputs.node-version }}
 
       - name: Install dependencies
         run: npm ci --omit=optional
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -74,8 +74,7 @@ jobs:
     steps:
       - uses: actions/checkout@v2
       - id: list_modules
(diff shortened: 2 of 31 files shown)


eddumelendez (approved):
nice catch! LGTM

cristianrgreco (approved):
Thanks @javierlopezdeancos, will merge when the build passes

cristianrgreco:
The module was merged recently and looks like it's tests are flaking.

cristianrgreco:
I've looked into the test failures, seems it's specific to node 20

javierlopezdeancos:
I found that tc-java have the same issue 
https://www.couchbase.com/forums/t/test-container-error-with-image-couchbase-enterprise-7-1-4/36200
There are an open issue to resolve it but, is not clear the approach to fix it
https://github.com/testcontainers/testcontainers-java/pull/7049

Issue #706, opened 2024-01-16
Add support for image name substitution


cristianrgreco:
Add support for image name substitution: https://java.testcontainers.org/features/image_name_substitution/

silh:
I would like to work on this one if possible.

cristianrgreco:
@silh that would be appreciated

PR #729, opened 2024-03-16
Add support for image registry substitution via environment variable

silh:
Added support for dockerhub registry substitution: when a registry is not provided for an image and the env variable `TESTCONTAINERS_HUB_IMAGE_NAME_PREFIX` is set, then the registry specified in `TESTCONTAINERS_HUB_IMAGE_NAME_PREFIX` will be added to an image.

Adds part of the functionality mentioned in https://github.com/testcontainers/testcontainers-node/issues/706.

What's left out is:
- ability to specify the substitution via properties file - not sure if it should be added, properties files are not that common in JS world.
- a method `asCompatibleSubstituteFor` - no need for it in the current state as there are no checks if the image is compatible with the one the library expects.

### DIFF
--- a/docs/features/image-name-substitution.md
+++ b/docs/features/image-name-substitution.md
@@ -1,0 +1,9 @@
+# Image name substitution
+
+Testcontainers supports automatic substitution of Docker image names.
+
+This allows replacement of an image name specified in test code with an alternative name - for example, to replace the name of a Docker Hub image dependency with an alternative hosted on a private image registry.
+
+This is advisable to avoid Docker Hub rate limiting, and some companies will prefer this for policy reasons.
+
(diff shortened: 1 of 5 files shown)


docs/features/image-name-substitution.md
```diff
@@ -0,0 +1,9 @@
+# Image name substitution
```
cristianrgreco, line 1: To add a new page you'll need to add it here: https://github.com/testcontainers/testcontainers-node/blob/cce5bccd7f4dee04de3948bdf7dae52e21969883/mkdocs.yml#L37-L43. However I wonder if this would be better off in the `Images` feature page?

silh: Sounds good, I'll add it to `Images`.

jonesbusy:
Possible cause of https://github.com/joyrex2001/kubedock/issues/84

ilia-beliaev-miro:
Hey @jonesbusy, I've prepared a fix for this issue here - Mentioned [here](https://github.com/joyrex2001/kubedock/issues/84#issue-2222355855) and [here](https://github.com/testcontainers/testcontainers-node/pull/729#issuecomment-2034003392). Meanwhile removing the trailing slash should help with the issue.

PR #730, opened 2024-03-17
Add `OneShotWaitStrategy`

omerlh:


cristianrgreco:
I think the issue is that the container runtime strategy is injected into wait strategy before it is initialised. There is a complexity in that the runtime initialisation is lazy, and a user defines their wait strategy before any container is started. I reckon you want to get the container runtime strategy in the wait strategy itself instead of passing it in.

omerlh:
I wasn't sure - I tried following the docs on creating a wait strategy that needs the container client, but seemed weird to me I need to pass it on the ctor. I guess you are saying I should not use `StartupCheckStrategy` here? Use `AbstractWaitStrategy` like `log-wait-startegy?

omerlh:
I let myself change StartupCheckStrategy to get the container runtime only when checking but it is still failing with the same error...

cristianrgreco:
> I guess you are saying I should not use StartupCheckStrategy here? Use AbstractWaitStrategy like log-wait-startegy?

Yes that's right

omerlh:
So look at my latest changes - I think I did that, but it still failing...

packages/testcontainers/src/wait-strategies/one-shot-startup-startegy.test.ts
```diff
@@ -0,0 +1,31 @@
+import { GenericContainer } from "../generic-container/generic-container";
+import { Wait } from "./wait";
+import { checkContainerIsHealthy } from "../utils/test-helper";
+
+jest.setTimeout(180_000);
+
+const mockImageInspect = jest.fn();
```
cristianrgreco, line 7: The reason why the tests are failing is because you are mocking the entirety of Dockerode, so Testcontainers ends up failing to create a valid container runtime strategy. I guess you only want to mock the image inspect function

cristianrgreco: Even better, is it really necessary to mock anything here? I think we can just start a container that exits immediately

omerlh: Ah yes this was it! Thank you :) I added basic test, but I am still thinking how I test failed cases

omerlh: fixed, do you want me to squash my commits?

cristianrgreco: Thanks. No need I'll squash the commits on merge

cristianrgreco:
Could you please add an example in the wait strategies doc?

cristianrgreco:
Tests are failing:

```
FAIL packages/testcontainers/src/wait-strategies/startup-check-strategy.test.ts
  ● Test suite failed to run

    packages/testcontainers/src/wait-strategies/startup-check-strategy.test.ts:26:8 - error TS2554: Expected 0 arguments, but got 1.

    26     })(client);
              ~~~~~~
    packages/testcontainers/src/wait-strategies/startup-check-strategy.test.ts:40:8 - error TS2554: Expected 0 arguments, but got 1.

    40     })(client);
              ~~~~~~
    packages/testcontainers/src/wait-strategies/startup-check-strategy.test.ts:58:8 - error TS2554: Expected 0 arguments, but got 1.

    58     })(client);
```

omerlh:
Sure, I think I fixed both!

docs/features/wait-strategies.md
```diff
@@ -184,6 +178,18 @@ const container = await new GenericContainer("alpine")
   .start();
 ```
 
+## One shot startup strategy example
```
cristianrgreco, line 181:
```suggestion
## One shot
```

omerlh:
Thanks for merging!

PR #736, opened 2024-03-22
Add Weaviate module

antas-marcin:
This PR adds a Weaviate module.

### DIFF
--- a/packages/modules/weaviate/src/weaviate-container.ts
+++ b/packages/modules/weaviate/src/weaviate-container.ts
@@ -1,0 +1,40 @@
+import { AbstractStartedContainer, GenericContainer, StartedTestContainer, Wait } from "testcontainers";
+
+const WEAVIATE_HTTP_PORT = 8080;
+const WEAVIATE_GRPC_PORT = 50051;
+
+export class WeaviateContainer extends GenericContainer {
+  constructor(image = "semitechnologies/weaviate:1.24.5") {
+    super(image);
+    this.withCommand(["--host", "0.0.0.0", "--scheme", "http", "--port", `${WEAVIATE_HTTP_PORT}`]);
+    this.withExposedPorts(WEAVIATE_HTTP_PORT, WEAVIATE_GRPC_PORT);
+    this.withEnvironment({
+      AUTHENTICATION_ANONYMOUS_ACCESS_ENABLED: "true",
+      PERSISTENCE_DATA_PATH: "/var/lib/weaviate",
+    });
+    this.withWaitStrategy(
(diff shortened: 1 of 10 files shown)


packages/modules/weaviate/src/weaviate-container.ts
```diff
@@ -0,0 +1,43 @@
...
+  constructor(image = "semitechnologies/weaviate:1.24.5") {
+    super(image);
+    this.withCommand(["--host", "0.0.0.0", "--scheme", "http", "--port", "8080"])
+    this.withExposedPorts(WEAVIATE_REST_PORT, WEAVIATE_GRPC_PORT);
+    this.withEnvironment({ 
+      QUERY_DEFAULTS_LIMIT: "25",
+      AUTHENTICATION_ANONYMOUS_ACCESS_ENABLED: "true",
+      PERSISTENCE_DATA_PATH: "/var/lib/weaviate",
+      DEFAULT_VECTORIZER_MODULE: "none",
+      CLUSTER_HOSTNAME: "node1",
```
eddumelendez, line 16: Just curious about `QUERY_DEFAULTS_LIMIT`, `DEFAULT_VECTORIZER_MODULE`, `CLUSTER_HOSTNAME` because those have not been set in java and go implementations.

* `QUERY_DEFAULTS_LIMIT` is a good default? According to the [docs](https://weaviate.io/developers/weaviate/config-refs/env-vars) it changed in version 1.24. Unless performance is improved we use to respect the default.
* `DEFAULT_VECTORIZER_MODULE`, any improvement by switching from the default to none?
* `CLUSTER_HOSTNAME`, I think this one can be let to the user when a cluster is created.

eddumelendez (commented):
Thanks for your contribution! Just leave some questions in order to understand better about the settings and if those should be applied to java and go as well.

packages/modules/weaviate/src/weaviate-container.ts
antas-marcin: hmm, actually you are right that this can be simplified, we can actually remove those settings. I was used to our development settings, that why I copied them mechanically here. I will remove them to expose the same settings as we do in java and go modules.

cristianrgreco:
Thanks for the contribution @antas-marcin. Looks like the dependencies aren't being included in the package-lock file

antas-marcin:
Let me take a look at those errors

antas-marcin:
I have pushed updated `package-lock.json` should be fine now

javierlopezdeancos (approved):

eddumelendez (approved):

PR #746, opened 2024-04-02
Add ChromaDB module 

AlabasterAxe:
ChromaDB already has testcontainers for Java and Go: https://testcontainers.com/modules/chroma/ This just adds it for Node as well.


javierlopezdeancos (approved):

packages/modules/chromadb/src/chromadb-container.ts
```diff
@@ -0,0 +1,31 @@
...
+}
+
+export class StartedChromaDBContainer extends AbstractStartedContainer {
+  private readonly host: string;
+  private readonly port: number;
+
+  constructor(startedTestContainer: StartedTestContainer) {
+    super(startedTestContainer);
+    this.host = this.startedTestContainer.getHost();
+    this.port = this.startedTestContainer.getMappedPort(CHROMADB_PORT);
```
cristianrgreco, line 25: The port may change if the container is restarted, I would get the mapped port on demand each time instead of once on construction

cristianrgreco (commented):
Thanks for the contribution! Could you please also add a page in the docs?

AlabasterAxe:
> Thanks for the contribution! Could you please also add a page in the docs?

done! although it doesn't look like the netlify built again so I'm not 100% sure that I got the markdown right for the example.

javierlopezdeancos:
> > Thanks for the contribution! Could you please also add a page in the docs?
> 
> done! although it doesn't look like the netlify built again so I'm not 100% sure that I got the markdown right for the example.

I just re run the jobs again

cristianrgreco:
> > Thanks for the contribution! Could you please also add a page in the docs?
> 
> done! although it doesn't look like the netlify built again so I'm not 100% sure that I got the markdown right for the example.

Thanks, you'll need to add a link in the mkdocs.yml in the project root

AlabasterAxe:
Okay, I've updated the readme to add a section about updating the docs, too in case that's useful :)

cristianrgreco:
> Okay, I've updated the readme to add a section about updating the docs, too in case that's useful :)

Thanks, though that would be better placed in the CONTRIBUTING.md

AlabasterAxe:
Okay, I've moved the bit about updating the docs into contributing and added a link to contributing.md from the readme

Issue #747, opened 2024-04-03
Multiple "regression" on 10.7.0 (wait strategy) and 10.8.0 (image substitution) when running using kubedock environment


jonesbusy:
Hi,

I already posted some details on https://github.com/joyrex2001/kubedock/issues/84 because I'm not sure where is located the issue.

We noticed behavior/regression changes when upgrading from 10.6 to 10.8.1 when running using kubedock on Kubernetes

We have the same setup with Testcontainer java (version 1.19.7) and testcontainer .NET (version 3.7.0) and all works fine

But for some reason node module is not working post 10.6

I noticied 2 PR that could have cause such changes

https://github.com/testcontainers/testcontainers-node/pull/712 (released on 10.7)
https://github.com/testcontainers/testcontainers-node/pull/729 (release on 10.8)

The issue is not producible using a "real" docker socket with Rancher Deskop

**Expected Behaviour**

Should work

**Actual Behaviour**

Container is started like this

```javascript
    wireMockContainer = await new GenericContainer("wiremock/wiremock:3.3.1-2")
      .withExposedPorts(8080)
      .withCommand(["--verbose"])
      .withWaitStrategy(Wait.forLogMessage(/port:.*8080/))
      .withStartupTimeout(120000)
      .start();
```

### On 10.7

```
[11:33:21.886+02:00] - 2024-04-03T09:33:21.059Z testcontainers [ERROR] [283b40bf8ce7] Container failed to be ready: Error: Log stream ended and message "/.+ Started!/" was not received
````

```
[11:33:21.885+02:00] - 2024-04-03T09:33:20.978Z testcontainers [DEBUG] [283b40bf8ce7] Waiting for container to be ready...
[11:33:21.885+02:00] - 2024-04-03T09:33:20.979Z testcontainers [DEBUG] [283b40bf8ce7] Waiting for log message "/.+ Started!/"...
[11:33:21.885+02:00] - 2024-04-03T09:33:20.979Z testcontainers [DEBUG] [283b40bf8ce7] Fetching container logs...
[11:33:21.885+02:00] - 2024-04-03T09:33:20.982Z testcontainers [DEBUG] [283b40bf8ce7] Demuxing stream...
[11:33:21.885+02:00] - 2024-04-03T09:33:20.982Z testcontainers [DEBUG] [283b40bf8ce7] Demuxed stream
[...]

PR #748, opened 2024-04-03
Fix parsing registry from env variable to work with trailing slash

ilia-beliaev-miro:
Fixed the trailing slash in registry in env variable parsing behavior to work as specified in documentation.

### DIFF
--- a/packages/testcontainers/src/container-runtime/image-name.test.ts
+++ b/packages/testcontainers/src/container-runtime/image-name.test.ts
@@ -54,20 +54,23 @@ describe("ContainerImage", () => {
       expect(imageName.string).toBe("aa285b773a2c042056883845aea893a743d358a5d40f61734fa228fde93dae6f:1");
     });
 
-    it("should substitute no registry with the one provided via TESTCONTAINERS_HUB_IMAGE_NAME_PREFIX", () => {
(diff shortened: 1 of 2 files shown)


cristianrgreco:
@ilia-beliaev-miro do you know if this PR resolves https://github.com/testcontainers/testcontainers-node/issues/747 in its entirety?

ilia-beliaev-miro:
> @ilia-beliaev-miro do you know if this PR resolves #747 in its entirety?

Hey @cristianrgreco, I can't test it with their setup. Maybe @ jonesbusy could check it?

jonesbusy:
I don't think it resolve it. I was trying to remove the trailing slash from the env vars and the status was the same. Was hanging on the lock. Same if I remove the env vars (for the 10.8)

While this PR will avoid creating double slash, it doesn't fix the issues I have.

Right now I cannot explain further those regression with 10.7 and 10.8. Also why using same setup it work with testcontainer java et .net.

Using 10.7 which doesn't use the env vars, error is that the "Started" log message is not received. Which made me think it was related to changes on the  wait conditions.

cristianrgreco:
@ilia-beliaev-miro is this PR ok to merge/release?

ilia-beliaev-miro:
> @ilia-beliaev-miro is this PR ok to merge/release?

@cristianrgreco I am working on fixing the problem with the authentication with registry. Hope to push it in a couple of hours.

Issue #571, opened 2023-05-12
support building docker images using BuildKit


helmlover:
The claim ["Docker [...] Works out of the box."](https://node.testcontainers.org/supported-container-runtimes/) is no longer true.
That is because docker now uses BuildKit per default, while the BuildKit Dockerfile syntax is not supported when building docker images with testcontainers-node (e.g. with `GenericContainer.fromDockerfile(buildContext).build()`)

When building a Dockerfile containing BuildKit-features e.g.
```
RUN--mount=type=cache,id=maven,target=/root/.m2/repository mvn --batch-mode --no-transfer-progress dependency:resolve dependency:resolve-plugins
```
through testcontains-node (with `export 'DEBUG=testcontainers*'`), the output currently looks like:
```
2023-05-12T08:52:02.265Z testcontainers:build [localhost/43420fa7dc4a:2c9e53cede2c] {"errorDetail":{"message":"the --mount option requires BuildKit. Refer to https://docs.docker.com/go/buildkit/ to learn how to build images with BuildKit enabled"},"error":"the --mount option requires BuildKit. Refer to https://docs.docker.com/go/buildkit/ to learn how to build images with BuildKit enabled"}
```

Exporting `export DOCKER_BUILDKIT=1` does not change the problem/output.

PS: Sibling issue in testcontainers-java: https://github.com/testcontainers/testcontainers-java/issues/2857

praveensvsrk:
+1, we need support for this....curious if there a workaround available to achieve the RUN cache like functionality without buildkit features?

silh:
It looks like what's necessary is a support for creating a session and running it. Go implementation for that in terraform was added here - https://github.com/kreuzwerker/terraform-provider-docker/pull/387/files#diff-4596d40531ae2e21f6074d104e6dc7317537946b56d95df847c9209dfbe30fceR329 The session run code is [here](https://github.com/moby/buildkit/blob/master/session/session.go#L104)

mikeseese:
Note that the API does support using BuildKit with the `version` option: this issue has fixed in the correct repo: https://github.com/moby/moby/blob/master/api/swagger.yaml#L8722-L8731

The linked upstream issue (https://github.com/docker/for-linux/issues/1136) is on a deprecated/seemingly abandoned repo.

silh:
The main problem is with setting up a websocket connect for the session which is required for version 2. IIRC, `dockerode` (or its underling library `docker-modem`) didn't support that.

[...]

PR #761, opened 2024-04-27
Allow building with BuildKit

schummar:
Adds a `withBuildKit` option. With the option enabled, a build will be done using BuildKit - within a docker container.
This should be relatively independent on the host system's BuildKit support.

Questions:
- Should the env var `DOCKER_BUILDKIT=1` enabled the setting by default?
- Should it be enabled by default in general? For a "just works" for newer dockerfile features like mounts?

Solves #571

### DIFF
--- a/docs/features/images.md
+++ b/docs/features/images.md
@@ -91,6 +91,15 @@ const container = await GenericContainer
   .build();
 ```
 
+### Using BuildKit
+
+```javascript
+const container = await GenericContainer
+  .fromDockerfile("/path/to/build-context")
+  .withBuildKit()
+  .build();
+```
+
 ## Image name substitution
 
 Testcontainers supports automatic substitution of Docker image names.
--- a/packages/testcontainers/fixtures/docker/docker-with-buildkit-features/Dockerfile
+++ b/packages/testcontainers/fixtures/docker/docker-with-buildkit-features/Dockerfile
@@ -1,0 +1,21 @@
+FROM node:10-alpine
+
+MAINTAINER Cristian Greco
+
+EXPOSE 8080
+
+RUN --mount=type=cache,target=/var/cache/apk apk add --no-cache curl dumb-init libcap openssl
+
+RUN openssl req -x509 -nodes -days 36500 \
+    -subj  "/C=CA/ST=QC/O=Company Inc/CN=localhost" \
+    -newkey rsa:2048 -keyout /etc/ssl/private/cert.key \
+    -out /etc/ssl/certs/cert.crt \
+    && chmod 666 /etc/ssl/private/cert.key
+
+RUN npm init -y \
+    && npm install express@4.16.4
+
+COPY index.js .
+
+ENTRYPOINT ["/usr/bin/dumb-init", "--"]
+CMD ["node", "index.js"]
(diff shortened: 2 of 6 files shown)


cristianrgreco:
Hi @schummar, thanks for the contribution! 

Because build kit is the default image builder I'd say this should be the default behaviour. 

However because this implementation requires a separate container for the build, as well as copying potentially large files between images, I'm concerned about performance, which I assume is significantly worse when compared to a non-buildkit build. If we make this functionality opt-in (keeping withBuildKit), then I'm concerned about maintenance as it's a large and complex change. I haven't run the tests yet but I'm assuming issues with other container runtimes like podman too.  

My plan was to wait for build kit to be natively supported by Docker engine and Dockerode, which should result in no or minor changes and no performance impact.

schummar:
Fair enough. I think the performance might not too bad in comparison. BuildKit builds are usually run in a container anyway, as far as I know - but my knowledge there is admittedly limited. I totally get the reluctance to add the complexity though.

Unfortunately, it's unclear when this might be supported (the referenced issues are 3,5 years old at this point). Usage of BuildKit features is getting more common - my very first try with testcontainers immediately failed because of this 😉. So, it would be good to have some kind of solution. What do you think about these options:
1. I can create a separate npm package that provides the functionality. To load the image some access to container runtime would be great. Like keeping the proposed `ImageClient.load` method. Or exposing the underlying `Dockerode` instance?
2. Would you be interested in having that package as a testcontainers module?

schummar:
@cristianrgreco when you have a minute, could you give feedback?

schummar:
Ok, completely new development:

As @mikeseese pointed out (https://github.com/testcontainers/testcontainers-node/issues/571#issuecomment-2159657555), the version flag is in fact working.

Upon investigation I found that I can make it work from testcontainers as well with one caveat: When passing `pull: "true"` and `version: "2"`, the build always fails with "node:10-alpine: failed to resolve source metadata for docker.io/library/node:10-alpine: no active sessions". And because of this issue that is handled in PR #771, `pull: "true" is currently always passed and therefore `vesion: "2" never seemed to work at all!

That make this PR so much more concise! More open questions:
- I suppose the mentioned error is a docker bug. Until it's resolved, should testcontainers emit a warning when using both options together?
- Currently dockerode types don't contain the `version` flag. Either we wait for https://github.com/DefinitelyTyped/DefinitelyTyped/pull/69780 or use a cast `as ImageBuildOptions` to cover it up for now

cristianrgreco:
Hi @schummar, thanks for staying on top of this issue! I've checked out this branch and run the test `should work with buildKit` and it passes. Are you saying that it doesn't for you? If it does could you please add a test for your valid example which is failing so I can also have a look? I'm on Docker version 24.0.6 on an M1 Mac

schummar:
A simple manual way to reproduce it is to run `docker image rm node:10-alpine`, them execute just that test case: `npm run test -- packages/testcontainers/src/generic-container/generic-container-dockerfile.test.ts -t "buildKit"`. It will fail.

Then run `docker pull node:10-alpine` and execute the test again and it will succeed.

I can try and come up with a test case later. Removing images is not covered by the ContainerRuntimeClient, so I have to find a way around that.

cristianrgreco:
Closing for now. We're waiting for upstream support.

Issue #739, opened 2024-03-27
`An error occurred listing credentials` with `docker-credential-gcr`


glebbash:
**Expected Behaviour**
Should work fine :)

**Actual Behaviour**
Having [docker-credential-gcr](https://github.com/GoogleCloudPlatform/docker-credential-gcr) installed makes it not possible to start testcontainers.

Problem is that running `docker-credential-gcr list` command returns exit code 1, which triggers an error here: https://github.com/testcontainers/testcontainers-node/blob/0488293b7b9cb5883a1f0c46fa4df74c53d465e6/packages/testcontainers/src/container-runtime/auth/credential-provider.ts#L46

This is caused by `list` command being not implemented:
```
Usage: docker-credential-gcr <flags> <subcommand> <subcommand args>
...
Subcommands for Docker credential store API:
        erase            (UNIMPLEMENTED) erase any stored credentials for the server specified via stdin
        get              for the server specified via stdin, return the stored credentials via stdout
        list             (UNIMPLEMENTED) list all stored credentials
...
```

> P.S: not really sure what commands are required to be implemented by docker credential helpers spec (if there is one) so it might be an issue for [docker-credential-gcr](https://github.com/GoogleCloudPlatform/docker-credential-gcr) repo.

**Testcontainer Logs**
```
Error: An error occurred listing credentials
    at /workspace/***/node_modules/testcontainers/build/container-runtime/auth/credential-provider.js:32:35
    at ChildProcess.exithandler (node:child_process:430:5)
    at ChildProcess.emit (node:events:518:28)
    at maybeClose (node:internal/child_process:1105:16)
    at ChildProcess._handle.onexit (node:internal/child_process:305:5)
```

**Steps to Reproduce**
1. Install [docker-credential-gcr](https://github.com/GoogleCloudPlatform/docker-credential-gcr)
2. Run the following (or any other container):
```js
import { GenericContainer, Wait } from "testcontainers";

const spannerEmulator = await new GenericContainer(
  "gcr.io/cloud-spanner-emulator/emulator"
)
  .withExposedPorts(9010)
  .withWaitStrategy(Wait.forLogMessage("gRPC server listening at"))
  .start();
```

**Environment Information**
- Operating System: Ubuntu 22.04
- Docker Version: 25.0.3
- Node version: v20.11.1
- Testcontainers version: 10.7.2

ghtaylor:
If anyone has come to this page from a quick Google, the below fixed it for me:

[...]

PR #764, opened 2024-05-02
Fix fetching registry credentials when using Gcloud cred helper

HiranmayaGundu:
When creating a container using a docker image from a private repository, testcontainers would fail to auth correctly, and would never fetch the image. i.e., `FROM gcr.io/<image>` would fail. This is not a problem when running the image directly. After a bit of digging, it seems like `serverURL` was being set as undefined, and it seems like the fix is to use the credentials object that has the server list. 

I'm not sure how to go about adding a test for this, it needs a Dockerfile to a private repository that the project has access to.


javierlopezdeancos:
Hi @HiranmayaGundu, looking the problem of testing the functionality, could we test this functionality with a private access image in Docker Hub? 🤔

javierlopezdeancos:
hi @HiranmayaGundu yep, I think that should be an image created to someone from the docker organization to this proposes cc @eddumelendez @cristianrgreco

kiview:
I don't think using Docker Hub for such an integration test makes sense, since community contributors would not have access to such a private image.

Can we replicate the scenario by starting a registry ourselves, or would the auth mechanism behave very differently?

If a maintainer of the repo (e.g. @cristianrgreco) can manually test that this works, I think it is also fine and we should not block the merging if the can't find a way to conveniently test it.

HiranmayaGundu:
@javierlopezdeancos @cristianrgreco wanted to bump this for review 🙏🏽

cristianrgreco:
Hi @HiranmayaGundu, thanks for raising this PR, I'll verify this works this weekend.

cristianrgreco:
@HiranmayaGundu I can't reproduce the issue an image from AWS ECR. I'm creating a container as follows:

```ts
new GenericContainer("[ID].dkr.ecr.eu-west-2.amazonaws.com/[IMAGE-NAME]:latest")
```

I am running this after I have done a `docker login`. After logging in I can see the registry key under `auths` in `~/.docker/config.json`.

Could you share some more info:
- How to reproduce
- What you `~/.docker/config.json` looks like for the registry you're failing to auth with via Testcontainers

HiranmayaGundu:
@cristianrgreco This issue only occured for me when I tried to build an image, so when doing 
```js
const container =  await GenericContainer.fromDockerfile("./dockerfile").build();
``` 
where the docker file is 
```
FROM <private repo>/image
```
The issue was consistently reproducible for me. I was using a private GCR repository for the base image.
my docker config looks like this 
```
{
        "credsStore": "desktop",
        "credHelpers": {
                "asia.gcr.io": "gcloud",
                "eu.gcr.io": "gcloud",
                "gcr.io": "gcloud",
                "marketplace.gcr.io": "gcloud",
                "staging-k8s.gcr.io": "gcloud",
                "us.gcr.io": "gcloud"
        },
        "currentContext": "desktop-linux"
}          
```

cristianrgreco:
I've tried as you said from a Dockerfile and it also works.

The difference between your setup and mine is that you have a cred helper setup for gcloud. This _is_ supported so we'll need to further debug what's going on.

When you run:

```bash
docker-credential-gcloud list
```

Could you confirm that any of the keys in the response matches the gcr registry from which you're trying to pull the image? A match is considered it they're equal with or without protocol. If unsure please share the contents here, omitting any sensitive data.

If there's a match, run:

```bash
echo '<registry>' | docker-credential-gcloud get
```

Where `registry` is the exact same registry as you have in your Dockerfile. Do you get a response and does the `ServerURL` match? If not please share details as to what's mismatched.

HiranmayaGundu:
> When you run:
docker-credential-gcloud list
Could you confirm that any of the keys in the response matches the gcr registry from which you're trying to pull the image? A match is considered if they're equal with or without protocol. If unsure please share the contents here, omitting any sensitive data.

Yes, I have an exact match for the registry I am pulling from. 

> run:
echo '<registry>' | docker-credential-gcloud get
Where registry is the exact same registry as you have in your Dockerfile. Do you get a response and does the ServerURL match? If not please share details as to what's mismatched.

There is no `ServerURL` in the response, just a `Secret` and  `Username` field

cristianrgreco:
Looks like there's a lot of inconsistency around the Google credential helpers, see here as well: https://github.com/testcontainers/testcontainers-node/issues/739. I guess there's no harm in defaulting to the registry URL if the `response.ServerURL` is undefined. @HiranmayaGundu are you OK to update the PR for this? I'm thinking `registryAddress: credentialForRegistry => registryAddress: response.ServerURL ?? credentialForRegistry` and a test in `credential-provider.test.ts` for it

HiranmayaGundu:
Yeah, I can push the changes later today!

HiranmayaGundu:
@cristianrgreco made the changes!

packages/testcontainers/src/container-runtime/auth/credential-provider.ts
```diff
@@ -34,7 +36,7 @@ export abstract class CredentialProvider implements RegistryAuthLocator {
     return {
       username: response.Username,
       password: response.Secret,
-      registryAddress: response.ServerURL,
+      registryAddress: response.ServerURL || credentialForRegistry,
```
cristianrgreco, line 39:
```suggestion
      registryAddress: response.ServerURL ?? credentialForRegistry,
```

HiranmayaGundu: adopted the suggestion

cristianrgreco:
Thanks for the PR and for your patience in getting it reviewed and merged @HiranmayaGundu!

PR #766, opened 2024-05-08
feat: Improved ChromaDB docs and tests

tazarov:


packages/modules/chromadb/src/chromadb-container.ts
```diff
@@ -3,7 +3,7 @@ import { AbstractStartedContainer, GenericContainer, StartedTestContainer, Wait
 const CHROMADB_PORT = 8000;
 
 export class ChromaDBContainer extends GenericContainer {
-  constructor(image = "chromadb/chroma:0.4.22") {
+  constructor(image = "chromadb/chroma:0.5.0") {
```
eddumelendez, line 6: IIRC, this could bring a breaking change with the env var for auth and also for those relying on version 0.4.22.

tazarov: Thanks @eddumelendez, indeed Auth is a breaking change in 0.5.0

cristianrgreco:
Hi @tazarov, thanks for raising. The smoke test is failing because the lock file doesn't match the updated dependency. There's also some lint issues.

packages/modules/chromadb/src/chromadb-container.ts
eddumelendez: what I meant is that existing users using default image will break test just by upgrading testcontainers, this should not happen in a patch release. So, my suggestion is to keep 0.4.22, users who want to use a different version can do it by overriding the default image.

tazarov:
@cristianrgreco, thanks for pinging me on this. I'll rebase and fix this tomorrow.

tazarov:
@cristianrgreco, rebased from latest main.

eddumelendez (approved):

cristianrgreco:
@tazarov This PR doesn't allow edits from maintainers so I can't fix it myself. Could you merge main? Should be enough to re-run `npm install` after merge.

cristianrgreco:
Replaced by https://github.com/testcontainers/testcontainers-node/pull/940

PR #770, opened 2024-05-31
Add RabbitMQ module

gundersong:
Add [RabbitMQ](https://www.rabbitmq.com) Module

### DIFF
--- a/packages/modules/rabbitmq/src/rabbitmq-container.test.ts
+++ b/packages/modules/rabbitmq/src/rabbitmq-container.test.ts
@@ -1,0 +1,68 @@
+import amqp, { ConsumeMessage } from "amqplib";
+import { RabbitMQContainer } from "./rabbitmq-container";
+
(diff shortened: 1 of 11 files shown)


packages/modules/rabbitmq/src/rabbitmq-container.test.ts
```diff
@@ -0,0 +1,68 @@
...
+
+    const connection = await amqp.connect(rabbitMQContainer.getAmqpUrl());
+    await connection.close();
+
+    await rabbitMQContainer.stop();
+  });
+  // }
+
+  // credentials {
+  it("different username and passord", async () => {
```
cristianrgreco, line 19:
```suggestion
  it("different username and password", async () => {
```

gundersong: Fixed! Good catch

PR #771, opened 2024-06-03
`GenericContainerBuilder` false pull policy should not pull image

schummar:
Small issue I found by accident. Instead of `this.pullPolicy` the check should be `this.pullPolicy.shouldPull()`.
The `cleanOpts` part makes sure that configs with `undefined` value are not included in the request - it caused issues for me.

### DIFF
--- a/packages/testcontainers/src/container-runtime/clients/image/docker-image-client.ts
+++ b/packages/testcontainers/src/container-runtime/clients/image/docker-image-client.ts
@@ -17,17 +17,13 @@ export class DockerImageClient implements ImageClient {
   constructor(protected readonly dockerode: Dockerode, protected readonly indexServerAddress: string) {}
 
   async build(context: string, opts: ImageBuildOptions): Promise<void> {
-    const cleanOpts = Object.fromEntries(
-      Object.entries(opts).filter(([_, value]) => value !== undefined)
-    ) as ImageBuildOptions;
-
     try {
-      log.debug(`Building image "${cleanOpts.t}" with context "${context}"...`);
+      log.debug(`Building image "${opts.t}" with context "${context}"...`);
       const isDockerIgnored = await this.createIsDockerIgnoredFunction(context);
       const tarStream = tar.pack(context, {
         ignore: (aPath) => {
           const relativePath = path.relative(context, aPath);
-          if (relativePath === cleanOpts.dockerfile) {
+          if (relativePath === opts.dockerfile) {
             return false;
           } else {
             return isDockerIgnored(relativePath);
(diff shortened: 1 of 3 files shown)


cristianrgreco:
Thanks for raising @schummar. Could you give an example where providing an undefined value caused issues?

schummar:
Hmm, I am pretty sure I saw some kind of error when the parameter was passed as undefined. But I cannot reproduce it now. Maybe I mixed something up. I have reverted that change now.

schummar:
Ok I guess that was the reason - it results in an error when running the build via Podman.

image

So I guess I bring the change back? I could also extract the logic to a util function to keep the code in docker-image-client.ts cleaner?

cristianrgreco:
It is interesting that `undefined` is interpreted an empty string and fails validation with Podman. 

What about if we use `null` instead of `undefined`? 
I'd rather explicitly add the `pull` field to the `opts` if `shouldPull()` returns `true`. I think that code is easier to reason about than something which dynamically alters the options.

Thanks again for raising this PR

schummar:
That's how the `querystring` lib handles it. Both `undefined` and `null` result in an empty string being sent. See https://runkit.com/embed/8n76jv2sc2lc

I changed it like you suggested. And I added a test case that fails without the fix.

cristianrgreco:
I think the test added is a duplicate of this one, https://github.com/testcontainers/testcontainers-node/blob/824cd8dbac0615ae00a83d82e8625a108b885ecb/packages/testcontainers/src/generic-container/generic-container-dockerfile.test.ts#L59-L73, which is unfortunately disabled for Podman

schummar:
That one test that the image is pulled when the pull policy says so. The new test tests that it doesn't pull when no pull policy is supplied.

jonesbusy:
Without any other changes, upgrading from 10.9 to 10.10 break my build on my environment 

```
[05:23:12.330+02:00] -     Could not find a working container runtime strategy
[05:23:12.330+02:00] - 
[05:23:12.330+02:00] -       18 |
[05:23:12.330+02:00] -       19 | export async function buildWireMockContainer(): Promise<StartedTestContainer> {
[05:23:12.330+02:00] -     > 20 |   return await new GenericContainer(`wiremock/wiremock:${WIREMOCK_VERSION}`)
```

Is suspect this PR cause my issue beucase it cause only GenericContainer to fail with `Could not find a working container runtime strategy`

I'm running testcontainer using kubedock (https://github.com/joyrex2001/kubedock)

cristianrgreco:
Hi @jonesbusy, the code change in this PR only takes effect after a working container runtime is established. If you want help please raise an issue and fill out the template, providing logs, etc.

jonesbusy:
> Hi @jonesbusy, the code change in this PR only takes effect after a working container runtime is established. If you want help please raise an issue and fill out the template, providing logs, etc. 

Hi, sure I will provide all the details like I did for https://github.com/testcontainers/testcontainers-node/issues/747

It's just that I didn't saw any other difference between 10.9 and 10.10 that might have cause the new issue I'm facing. 

Thanks

PR #778, opened 2024-06-14
Add ollama module

ilopezluna:
adds Ollama Module based on module of TC java


packages/modules/ollama/src/ollama-container.ts
```diff
@@ -0,0 +1,55 @@
...
+export const OLLAMA_PORT = 11434;
+
+export class OllamaContainer extends GenericContainer {
+  constructor(image = "ollama/ollama") {
+    super(image);
+    this.withExposedPorts(OLLAMA_PORT)
+      .withWaitStrategy(Wait.forLogMessage("Listening on "))
+      .withStartupTimeout(120_000);
+
+    getContainerRuntimeClient().then((client) => {
```
cristianrgreco, line 18: Although this should be fine because the container runtime client will be initialised before this constructor is invoked, it is still async and not guaranteed to resolve before `start` is called. I'd move this into the `start` method or one of the other lifecycle callbacks

ilopezluna: Thanks Cristian :)

cristianrgreco: Thank you Ignasi!

Issue #785, opened 2024-06-24
GenericContainer broken after upgrade from 10.9 to 10.10


jonesbusy:
Hi,

I've facing an issue after upgrade testcontainer node from 10.9 to 10.10 which any other changes

**Expected Behaviour**

Same as 10.9.0, it should work

**Actual Behaviour**

Doesn't work, tested with different framework (Backstage and one JS vite application). It's the same

**Testcontainer Logs**

```
[DEBUG] Testing container runtime strategy "TestcontainersHostStrategy"...
[DEBUG] Found custom configuration: dockerHost: "tcp://localhost:2475"
[DEBUG] Testing container runtime strategy "ConfigurationStrategy"...
[DEBUG] Container runtime strategy "ConfigurationStrategy" does not work
[DEBUG] Testing container runtime strategy "UnixSocketStrategy"...
[DEBUG] Testing container runtime strategy "RootlessUnixSocketStrategy"...
[DEBUG] Testing container runtime strategy "NpipeSocketStrategy"...
```

This line is different from

```
[DEBUG] Container runtime strategy "ConfigurationStrategy" does not work
```

On 10.9 is

```
[DEBUG] Container runtime strategy "ConfigurationStrategy" works
```

**Steps to Reproduce**

On code

```typescript
    wireMockContainer = await new GenericContainer("wiremock/wiremock:3.5.4-1")
      .withExposedPorts(8080)
      .withCommand(["--verbose"])
      .withWaitStrategy(Wait.forLogMessage(/port:.*8080/))
      .withStartupTimeout(120000)
      .start();
```

**Environment Information**

- Running node 20 LTS
- I'm running testcontainer with kubedock and following env set

```
DOCKER_HOST = 'tcp://localhost:2475'
TESTCONTAINERS_HUB_IMAGE_NAME_PREFIX  = '*********-docker-io.***************.ch/'
```

I can provide more details if needed

Thanks

cristianrgreco:
Hi @jonesbusy,

I gotta say I'm stumped by this one 😄 

These are the changes between 10.9 and 10.10 (https://github.com/testcontainers/testcontainers-node/releases/tag/v10.10.0):

```
🚀 Features
- Add ollama module @ilopezluna (https://github.com/testcontainers/testcontainers-node/pull/778)
- Add RabbitMQ module @gundersong (https://github.com/testcontainers/testcontainers-node/pull/770)

🐛 Bug Fixes
[...]

PR #787, opened 2024-06-25
Add logs around resolving a container runtime

cristianrgreco:
To help debug #785

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -83,8 +83,6 @@ jobs:
         include:
           - container-runtime: docker
             runner: ubuntu-22.04
-          - container-runtime: docker-rootless
-            runner: ubuntu-22.04
           - container-runtime: podman
             runner: ubuntu-22.04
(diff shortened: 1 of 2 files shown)


cristianrgreco:
Unable to get the project to build after repeated attempts. Though the only change is the addition of logs, the GH action agents are failing to install dependencies. Either an issue with the agents or NPM 🤷

jonesbusy:
Thanks, if you are able to release it I'm more that happy to test it and provide those logs

jonesbusy:
Thanks but sadly I will not be able to test before 2 weeks. In any case I think those logs are useful.

So if anyone else want to test the future version and provide logs I'm very curious to know the cause of the issue

cristianrgreco:
Thanks @jonesbusy. I think the issue may be specific to kubedock as I haven't heard anything else, so we may well have to wait for your return. Apologies this took so long to merge it turns out that https://github.com/marketplace/actions/rootless-docker has suddenly stopped working on the GH actions agents 🤷

PR #794, opened 2024-07-08
Log stacktrace when container runtime strategy throws

cristianrgreco:
To help further debug #785

### DIFF
--- a/packages/testcontainers/src/container-runtime/clients/client.ts
+++ b/packages/testcontainers/src/container-runtime/clients/client.ts
@@ -55,7 +55,10 @@ export async function getContainerRuntimeClient(): Promise<ContainerRuntimeClien
         return client;
       }
     } catch (err) {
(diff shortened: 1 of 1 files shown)


packages/testcontainers/src/container-runtime/clients/client.ts
```diff
@@ -55,7 +55,10 @@ export async function getContainerRuntimeClient(): Promise<ContainerRuntimeClien
         return client;
       }
     } catch (err) {
-      log.debug(`Container runtime strategy "${strategy.getName()}" does not work: ${err}`);
+      log.debug(`Container runtime strategy "${strategy.getName()}" does not work: "${err}"`);
+      if (err !== null && typeof err === "object" && "stack" in err && typeof err.stack === "string") {
```
cristianrgreco, line 59: Turns out `err` is not an instance of `Error` 🤷

PR #797, opened 2024-07-17
chore: export containers default images

idbenami:
- export each module default image as `<NAME>_IMAGE`


cristianrgreco:
Why?

cristianrgreco:
See image name substitution section here https://node.testcontainers.org/features/images/

PR #800, opened 2024-07-25
Support custom compose project names

jalaziz:
Allow custom project names for docker compose environments. This can be useful when running tests in CI environment where images may be built separately and loaded into the local docker image cache and consistent image names are needed (without having to specify local image names for each service).


cristianrgreco:
Thanks @jalaziz. Could you pls document this feature in the compose docs?

jalaziz:
@cristianrgreco added some basic docs based on the docs that are already there, but let me know if you would like me to add more comprehensive ones 🙏🏼

PR #801, opened 2024-07-25
MSSQL Docker image, 2022-latest -> 2022-CU13-ubuntu-22.04

cristianrgreco:
Looks like the test for express edition has suddenly started failing. Looks like there was a new release for 2022-latest (2022-CU14-ubuntu-22.04). Let's explicitly set the tag to the previous version 2022-CU13-ubuntu-22.04 in case the issue is coming from the new version.


oskarwidmark (approved):
LGTM!

PR #806, opened 2024-07-28
Add support for specifying image platform

### LINKED ISSUE #773: Add `withPlatform()` method to `GenericContainer`
**Expected Behaviour**
I want to be able to specify platform so I can run `linux/amd64` images on apple mac without issues.

Most images have both `amd64` & `arm64` flavours, but i'd like to be able to specify platform for the times where only x86 is available.

**Actual Behaviour**
I get errors because an image can't be pulled for the current host platform architecture. This is analagous with running `docker pull moonsonglabs/storage-hub` on an Apple Silicon device.

**Testcontainer Logs**
```sh
/Users/timbo/workspace/personal/test-repro-testcontainers/node_modules/.pnpm/docker-modem@3.0.8/node_modules/docker-modem/lib/modem.js:343
      var msg = new Error(
                ^

Error: (HTTP code 404) unexpected - no match for platform in manifest: not found
    at /Users/timbo/workspace/personal/test-repro-testcontainers/node_modules/.pnpm/docker-modem@3.0.8/node_modules/docker-modem/lib/modem.js:343:17
    at IncomingMessage.<anonymous> (/Users/timbo/workspace/personal/test-repro-testcontainers/node_modules/.pnpm/docker-modem@3.0.8/node_modules/docker-modem/lib/modem.js:370:9)
    at IncomingMessage.emit (node:events:531:35)
    at IncomingMessage.emit (node:domain:488:12)
    at endReadableNT (node:internal/streams/readable:1696:12)
    at process.processTicksAndRejections (node:internal/process/task_queues:82:21) {
  reason: undefined,
  statusCode: 404,
  json: null
}
```

**Steps to Reproduce**

Have a go with the snippet on an apple mac:
```typescript
import { GenericContainer } from "testcontainers";

async function main() { const container =await  new GenericContainer("moonsonglabs/storage-hub")
.start();
console.log("Started")

await new Promise(resolve=> setTimeout(resolve, 60000));
container.stop();
}

main()
```

> [!CAUTION]
> Error thrown: `Error: (HTTP code 404) unexpected - no match for platform in manifest: not found`

> [!NOTE]  
> I have tried to workaround this with `DOCKER_DEFAULT_PLATFORM` but this doesn't make a difference.

**Environment Information**
- Operating System: macOS 14.4.1
- Docker Version: 26.1.1
- Node version: 21.x
- Testcontainers version: 10.9.0

weyert:
Adds the ability to the pass `--platform`-argument for the docker container

fixes #773

### DIFF
--- a/packages/testcontainers/src/generic-container/generic-container.test.ts
+++ b/packages/testcontainers/src/generic-container/generic-container.test.ts
@@ -123,6 +123,15 @@ describe("GenericContainer", () => {
     expect(output).toEqual(expect.stringContaining("/tmp"));
   });
 
+  it("should set platform", async () => {
+    const container = await new GenericContainer("amd64/alpine")
+      .withPlatform("linux/amd64")
+      .withExposedPorts(8080)
+      .start();
+
+    await checkContainerIsHealthy(container);
+  });
+
   it("should set entrypoint", async () => {
     const container = await new GenericContainer("cristianrgreco/testcontainer:1.1.14")
       .withEntrypoint(["node"])
--- a/packages/testcontainers/src/generic-container/generic-container.test.ts
+++ b/packages/testcontainers/src/generic-container/generic-container.test.ts
@@ -124,9 +124,8 @@ describe("GenericContainer", () => {
   });
 
   it("should set platform", async () => {
-    // This test uses an older version of cristianrgreco/testcontainer as all the other tests are using the latest version and the code doesn't pull the image again when the platform differs at this time
-    const container = await new GenericContainer("cristianrgreco/testcontainer:1.1.13")
-      .withWorkingDir("/tmp")
+    const container = await new GenericContainer("cristianrgreco/testcontainer:1.1.14")
+      .withPullPolicy(PullPolicy.alwaysPull())
       .withCommand(["node", "../index.js"])
       .withPlatform("linux/amd64")
       .withExposedPorts(8080)
--- a/packages/testcontainers/src/generic-container/generic-container.test.ts
+++ b/packages/testcontainers/src/generic-container/generic-container.test.ts
@@ -124,7 +124,8 @@ describe("GenericContainer", () => {
   });
 
   it("should set platform", async () => {
-    const container = await new GenericContainer("cristianrgreco/testcontainer:1.1.14")
+    // This test uses an older version of cristianrgreco/testcontainer as all the other tests are using the latest version and the code doesn't pull the image again when the platform differs at this time
+    const container = await new GenericContainer("cristianrgreco/testcontainer:1.1.13")
       .withWorkingDir("/tmp")
       .withCommand(["node", "../index.js"])
       .withPlatform("linux/amd64")
--- a/packages/testcontainers/src/generic-container/generic-container.test.ts
+++ b/packages/testcontainers/src/generic-container/generic-container.test.ts
@@ -124,12 +124,15 @@ describe("GenericContainer", () => {
   });
 
   it("should set platform", async () => {
-    const container = await new GenericContainer("amd64/alpine")
+    const container = await new GenericContainer("cristianrgreco/testcontainer:1.1.14")
+      .withWorkingDir("/tmp")
+      .withCommand(["node", "../index.js"])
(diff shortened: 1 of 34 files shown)


packages/testcontainers/src/generic-container/generic-container.test.ts
```diff
@@ -123,6 +123,15 @@ describe("GenericContainer", () => {
     expect(output).toEqual(expect.stringContaining("/tmp"));
   });
 
+  it("should set platform", async () => {
+    const container = await new GenericContainer("amd64/alpine")
```
weyert, line 127: I try to do something like the code below to get the architecture of the image only that keeps failing and I don't understand. Some help would be appreciated.

```typescript
 const { output } = await container.exec(["uname", "-a"]);
expect(output).toEqual(expect.stringContaining("x86_64"));
```

cristianrgreco: A few things:
1. Starting a container and running an exec command isn't equivalent to `docker run`. You would need to keep the container running. Something like:
    ```js
    .withEntrypoint(["bash"])
    .withCommand(["-c", "echo 'START'; sleep infinity"])
    ```
2. The image `amd64/alpine` doesn't expose ports `8080`, so you'll want to remove `withExposedPorts(8080)` or else Testcontainers will assume the container has failed to start as it never exposes this port
3. This test won't work on ARM machines so I'm not sure it's possible to have a test for it

packages/testcontainers/src/generic-container/generic-container.test.ts
```diff
@@ -123,6 +123,19 @@ describe("GenericContainer", () => {
     expect(output).toEqual(expect.stringContaining("/tmp"));
   });
 
+  it("should set platform", async () => {
+    // This test uses an older version of cristianrgreco/testcontainer as all the other tests are using the latest version and the code doesn't pull the image again when the platform differs at this time
```
cristianrgreco, line 127: https://node.testcontainers.org/features/containers/#with-a-pull-policy

packages/testcontainers/src/generic-container/generic-container.test.ts
```diff
@@ -123,6 +123,19 @@ describe("GenericContainer", () => {
...
+    const container = await new GenericContainer("cristianrgreco/testcontainer:1.1.13")
+      .withWorkingDir("/tmp")
```
cristianrgreco, line 129: Setting the work dir and command should have no effect on your test

packages/testcontainers/src/generic-container/generic-container.test.ts
weyert: Thanks. I have updated it.

packages/testcontainers/src/generic-container/generic-container.test.ts
weyert: I have removed the `.withWorkingDir`-argument from the test case. I left the rest as that is an easy way to keep the container running so I can execute the `arch`-command to verify it pull the right arch version of the image.

I have locally confirmed it pulling the `linux/amd64`-version of the image on my Apple Silicon machine before making the pull policy change. See screenshot the below.

CleanShot 2024-08-11 at 14 55 33@2x

weyert:
@cristianrgreco I updated the documentation to add a callout to the new `withPlatform`-option

weyert:
@cristianrgreco Any more changes needed? Love to stop using my fork of this package :)

packages/testcontainers/src/generic-container/generic-container.test.ts
cristianrgreco: It'll work on your mac because you have rosetta emulation enabled, it won't work on ARM based linux/windows machines. Let's merge it anyway and if it's a big deal we'll think of something else

cristianrgreco (approved):

packages/testcontainers/src/generic-container/generic-container.test.ts
weyert: Thank you. Right, didn't think of that. You would other ARM based environments Linux or Windows would have a similar solution or maybe QEmu works.

PR #811, opened 2024-08-03
Set Node engine requirement to 18

### LINKED ISSUE #810: Breaking Change with Node 16 in Minor Release
**Expected Behaviour**
In the most recent release of testcontainers-node(https://github.com/testcontainers/testcontainers-node/releases/tag/v10.11.0), there is a change to stop using `node-fetch` in favor of the the native `fetch` first introduced in Node 18. This means this release will break anything running Node 16. 

Overall, I would have expected this change to come in a major release or for there to be clear documentation saying that `testcontainers-node` needs >= version 18 or node to run. I couldn't find any references to that on this GitHub or in the documentation site(https://node.testcontainers.org/configuration/). 

I will admit that Node 16 should not be used given it already reached EOL, but my team works across many machines with various versions of Node and it is hard for us to keep track of which versions have been updated. This request is for better warnings/release strategy to help us prevent seemingly random changes. 

Please let me know if my understanding is incorrect here 

**Actual Behaviour**
Using a careted version(^10.9.0) in the package.json means we automatically pulled in the latest version of `testcontainers-go` and were surprised to see our HTTP wait strategies all stop working. We were able to debug and see it all traced back to the fetch call. 

**Steps to Reproduce**
1. Using Node 16
2. Write a program with testcontainers
3. Use a HTTP wait strategy in the most recent version of testcontainers. 
4. See failure and timeout of wait strategy

**Environment Information**
- Operating System: Debian 11 
- Docker Version: Docker version 20.10.5
- Node version: 16
- Testcontainers version: 10.11.0

cristianrgreco:
Fixes #810


cristianrgreco:
Will be done as part of https://github.com/testcontainers/testcontainers-node/pull/938, just easier to make the change together with other stuff like updating deps

Issue #697, opened 2023-12-20
Possible documentation refinement for the MongoDBContainer module


csprocket777:
**Expected Behaviour**
I was following the instructions on the [MongoDB Module](https://node.testcontainers.org/modules/mongodb/) page and trying to implement in an ES6 environment that outputs CommonJS. The examples don't include the `require()/import` statement which would have helped a ton.

I also expected that once getting the container running and returning the value from `container.getConnectionString()` that it would simply connect using that string. I found that I had to add `?directConnection=true` to the connection string for it to actually connect.

Also, the `checkMongo` function seems to be from the test suite instead of from the library itself.

**Proposed changes:**
1. Make the generic similar to the below:
```typescript
  // ES6
  import { MongoDBContainer, StartedMongoDBContainer }  from '@testcontainers/mongodb';
  // --or--
  // CommonJS
  const { MongoDBContainer, StartedMongoDBContainer } = require('@testcontainers/mongodb');

  const mongodbContainer:StartedMongoDBContainer = await new MongoDBContainer().start();

  // ...your code

  await mongodbContainer.stop();
```
2. It would be super helpful to have some API docs for the modules. I ended up using `container.getConnectionString()` to get the connection string back to my testing suite.
3. Include a note about the usage of `?directConnection=true` in conjunction with the connection string returned. I ended up concatenating it to the end elsewhere: `mongodb://localhost:[PORT]/test-db?directConnection=true`

**Environment Information**
- Operating System: `MacOS 14.2 (Apple M1 Max)`
- Docker Version: `24.0.7, build afdd53b`
- Node version: `v18.12.1`
- Testcontainers version: `^10.4.0`

robvanderleek:
I bumped into the same issues.

FWIW, I've created a minimal example repository that shows the issue when not using `directConnection: true`: https://github.com/robvanderleek/testcontainers-mongodb-issue

**Environment Information**

- Operating System: MacOS 14.3.1 (Apple M1 Max)
- Docker Version: 24.0.6
- Node version: v20.11.0
- Testcontainers version: ^10.7.1

cristianrgreco:
I'm no Mongo expert.. should `directConnection=true` be included by default in the connection string?

robvanderleek:
[...]

PR #821, opened 2024-08-19
Update MongoDB docs about `directConnection`

BradLewis:
When playing around with the mongodb testcontainers today, we ran into an issue where the tests were unable to connect to the mongodb database. After a bit of debugging, we found that this was due to `directConnection` not being set to `true`. 

This PR is simply to update the documentation to make it explicit that you will need to set `directionConnection: true` either as an option when connecting, or directly as part of the connection string.

Addresses https://github.com/testcontainers/testcontainers-node/issues/697


cristianrgreco:
Love it, thanks @BradLewis

PR #805, opened 2024-07-28
Add `CloudStorageEmulatorContainer` and `PubSubEmulatorContainer` to gcloud module

weyert:
Adds two new emulator containers to the `gcloud`-package:

 - `CloudStorageEmulatorContainer`: Cloud Storage
 - `PubSubEmulatorContainer`: Cloud PubSub

The Cloud PubSub emalator container uses the same approach as used by the existing containers of using the `gcloud cli` tos tart the PubSub emulator.

The Cloud Storage emulator container uses an existing fake GCS emulator named `fsouza/fake-gcs-server`.

PR #825, opened 2024-08-27
Add `autoUpdateExternalUrl` to `CloudStorageEmulatorContainer`

alecmev:
Previously discussed here: https://github.com/testcontainers/testcontainers-node/pull/805#issuecomment-2308886117

Defaulting to `true` because it's very likely that the container will be used locally, and that it's desired for the emulator to return valid URLs. If an `externalUrl` is provided then the update is skipped.

Can't get tests to work with Colima (I added `forListeningPorts` but it hangs in `InternalPortCheck.isBound`, [here](https://github.com/testcontainers/testcontainers-node/blob/9f7fd4eff5e6f05440016010964a78aa80557cb1/packages/testcontainers/src/wait-strategies/utils/port-check.ts#L48-L50)), so need a CI run to check.

### DIFF
--- a/packages/modules/gcloud/src/cloudstorage-emulator-container.ts
+++ b/packages/modules/gcloud/src/cloudstorage-emulator-container.ts
@@ -7,13 +7,12 @@ const DEFAULT_IMAGE = "fsouza/fake-gcs-server";
 export class CloudStorageEmulatorContainer extends GenericContainer {
   private _externalURL?: string;
   private _publicHost?: string;
+  private autoUpdateExternalUrl = true;
 
   constructor(image = DEFAULT_IMAGE) {
     super(image);
 
-    this.withExposedPorts(PORT)
-      .withWaitStrategy(Wait.forLogMessage(/server started/g, 1))
-      .withStartupTimeout(120_000);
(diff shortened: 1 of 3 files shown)


packages/modules/gcloud/src/cloudstorage-emulator-container.ts
```diff
@@ -100,11 +111,7 @@ export class StartedCloudStorageEmulatorContainer extends AbstractStartedContain
    * @return a <code>host:port</code> pair corresponding to the address on which the emulator is
    * reachable from the test host machine.
    */
-  public getExternalUrl(): string {
-    if (this._externalURL) {
-      return this._externalURL;
-    } else {
-      return this.getEmulatorEndpoint();
-    }
```
alecmev, line 108: Removed this fallback because `undefined` is a valid state (i.e., "not set"), and `_externalURL` will usually be up-to-date, thanks to `updateExternalUrl`.

alecmev:
Both errors appear to be unrelated to this PR, do I need to do anything?

alecmev:
Thank you 👍

cristianrgreco:
Thank you, released in https://github.com/testcontainers/testcontainers-node/releases/tag/v10.13.0

PR #834, opened 2024-09-10
Fix LocalStack container hostname resolution

### LINKED ISSUE #833: localstack: constructor ignores LOCALSTACK_HOST
The fix to allow wait strategy overrides broke hostname resolution. @cristianrgreco 

_Originally posted by @joebowbeer in https://github.com/testcontainers/testcontainers-node/pull/712#discussion_r1751033935_

The call to `resolveHostname` was moved into the constructor but it should be made after the environment has been assigned. There is no way to provide an environment before the constructor is called so `resolveHostname` will never resolve `LOCALSTACK_HOST`

joebowbeer:
Fixes #833 

Moves the `resolveHostname` invocation back to the `beforeContainerCreated` method.

Fixes the networkAliases override in `resolveHostname` - to now call `withEnvironment`.

Also removes trailing `"` from the `resolveHostname` log message and encloses the value assigned to LOCALSTACK_HOST in quotes. Sample messages:

> testcontainers [INFO] LOCALSTACK_HOST environment variable set to "myhost" (explicitly as environment variable)

> testcontainers [INFO] LOCALSTACK_HOST environment variable set to "myalias" (to match last network alias on container with non-default network)

> testcontainers [INFO] LOCALSTACK_HOST environment variable set to "localhost" (to match host-routable address for container)

### DIFF
--- a/packages/modules/localstack/src/localstack-container.test.ts
+++ b/packages/modules/localstack/src/localstack-container.test.ts
@@ -68,5 +68,39 @@ describe("LocalStackContainer", () => {
     expect(response).toContain(`http://localstack:${LOCALSTACK_PORT}`);
     await container.stop();
     await awsCliInDockerNetwork.stop();
+    await network.stop();
+  });
+
+  it("should not override LOCALSTACK_HOST assignment", async () => {
+    const container = await new LocalstackContainer()
+      .withEnvironment({ LOCALSTACK_HOST: "myhost" })
+      .withNetworkAliases("myalias")
+      .start();
+
+    const { output, exitCode } = await container.exec(["printenv", "LOCALSTACK_HOST"]);
+    expect(exitCode).toBe(0);
+    expect(output).toContain("myhost");
+
+    await container.stop();
+  });
+
+  it("should override LOCALSTACK_HOST with last network alias", async () => {
+    const container = await new LocalstackContainer().withNetworkAliases("other", "myalias").start();
+
+    const { output, exitCode } = await container.exec(["printenv", "LOCALSTACK_HOST"]);
+    expect(exitCode).toBe(0);
+    expect(output).toContain("myalias");
+
+    await container.stop();
+  });
+
+  it("should assign LOCALSTACK_HOST to localhost", async () => {
+    const container = await new LocalstackContainer().start();
+
+    const { output, exitCode } = await container.exec(["printenv", "LOCALSTACK_HOST"]);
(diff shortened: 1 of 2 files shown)


cristianrgreco:
```
FAIL packages/modules/localstack/src/localstack-container.test.ts (47.116 s)
  LocalStackContainer
    ✓ should create a S3 bucket (26507 ms)
    ✓ should use custom network (14459 ms)
    ✕ should assign last network alias to LOCALSTACK_HOST by default (1365 ms)

  ● LocalStackContainer › should assign last network alias to LOCALSTACK_HOST by default

    expect(received).toContain(expected) // indexOf

    Expected substring: "myhost"
    Received string:    ""

      77 |
      78 |     const { output } = await container.exec(["printenv", "LOCALSTACK_HOST"]);
    > 79 |     expect(output).toContain("myhost");
         |                    ^
      80 |
      81 |     await container.stop();
      82 |   });

      at Object.<anonymous> (src/localstack-container.test.ts:79:20)
```

packages/modules/localstack/src/localstack-container.test.ts
```diff
@@ -68,5 +68,19 @@ describe("LocalStackContainer", () => {
...
+
+  it("should not override LOCALSTACK_HOST assignment", async () => {
+    const container = await new LocalstackContainer()
+      .withEnvironment({ LOCALSTACK_HOST: "myhost" })
+      .withNetworkAliases("myalias") // the last alias is assigned to LOCALSTACK_HOST by default
+      .start();
+
+    const { output, exitCode } = await container.exec(["printenv", "LOCALSTACK_HOST"]);
+    expect(exitCode).toBe(0);
+    expect(output).toContain("myhost");
```
cristianrgreco, line 82: This is now just testing that the environment variables are passed through. Is there something about this bug that we can test?

joebowbeer: This pass-through is apparently not happening, which may be another bug.

joebowbeer: @cristianrgreco I changed the test to test the original bug, which is that LOCALHOST was being "overridden". 

But now I wonder if the overrides are working...

joebowbeer: Should `resolveHostname` be calling `withEnvironment` to make the assignment?

cristianrgreco (approved):

PR #839, opened 2024-10-02
Fix for Docker registry passwords that contain a colon

### LINKED ISSUE #838: A docker auth password that contains a colon truncates the password
**Expected Behaviour**
The docker auth password should not be truncated if it contains a `:`

**Actual Behaviour**
The docker auth password is truncated if it contains a `:`

eg. If the password is `password1:1` it will be reduced to `password1`.

**Testcontainer Logs**
Since the password is truncated, it ends up producing the `unauthorized` error:

```
 testcontainers [ERROR] Failed to pull image "testcontainers/ryuk:0.5.1": Error: (HTTP code 500) server error - Head "https://registry-1.docker.io/v2/testcontainers/ryuk/manifests/0.5.1": unauthorized: incorrect username or password
```

**Steps to Reproduce**
1. Create a docker auth password that contains a colon (`:`)
2. Set the credentials in the `~/.docker/config.json`
3. Run your tests with testcontainer

**Code**
You can clearly see in the code it is currently doing a naive string split on `:`

https://github.com/testcontainers/testcontainers-node/blob/main/packages/testcontainers/src/container-runtime/auth/auths.ts#L24

A simple fix could be something like this:

```typescript
const [username, ...passwordParts] = decodedAuth.split(":");
const password = passwordParts.join(":");
```

I will raise a PR.

joshystuart:
Fixes #838

### DIFF
--- a/packages/testcontainers/src/container-runtime/auth/auths.test.ts
+++ b/packages/testcontainers/src/container-runtime/auth/auths.test.ts
@@ -64,5 +64,23 @@ describe("Auths", () => {
       };
       expect(await locator.getAuthConfig("https://registry.example.com", containerRuntimeConfig)).toEqual(authConfig);
     });
+
+    it("should return credentials from encoded auth when the password contains a colon", async () => {
+      const containerRuntimeConfig: ContainerRuntimeConfig = {
+        auths: {
+          "https://registry.example.com": {
+            email: "user@example.com",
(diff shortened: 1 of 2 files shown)


joshystuart:
This was a problem in the java version too, but was resolved some time ago: https://github.com/testcontainers/testcontainers-java/pull/1492

cristianrgreco (approved):

PR #843, opened 2024-10-11
Bump default postgres version

nikeee:


cristianrgreco:
Updating the default image version for any modules requires a major release, which is quite unsustainable. I will likely remove image defaults in the module constructors, instead of updating them. That way it's up to users to manage and update their own versions.

nikeee:
Removing it entirely would be even better, yes.

PR #845, opened 2024-10-15
pinpoint issue with running testcontainers-node in deno

jarlah:
This PR is not a feature or bugfix, its only purpose serves to debug running tc-node in deno. Since this PR still uses npm libraries for calling out to docker, like dockero, its still doing its job as good as any.
tests fails due to 
https://github.com/denoland/deno/issues/25661#issuecomment-2414232856
and i expect it to fail horribly with tsconfig due to non relaxed imports with extensions 
in addition to the huge amount of changes in code where it prepends "node:" to imports ;) 
its alot easier to debug why it fails if the code is ported versus blackbox testing with the npm library

EDIT: made an exampler repo here which bypasses tc-node  https://github.com/jarlah/dockerode-in-deno

### DIFF
--- a/packages/testcontainers/deno.json
+++ b/packages/testcontainers/deno.json
@@ -1,0 +1,12 @@
+{
+  "compilerOptions": {
+    "checkJs": false
+  },
+  "tasks": {
+    "dev": "deno run --allow-env --allow-read --allow-sys --allow-net --watch src/index.ts",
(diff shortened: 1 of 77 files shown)


jarlah:
made a simpler example here proving its dockerode/docker-modem issue
https://github.com/jarlah/dockerode-in-deno
the test fails with same error

PR #846, opened 2024-10-22
Docs - make info about Wait Strategies easier to find

osher:
When I was searching for it - I almost missed it, and found it by accident a moment before I gave up.

I believe it's a too important feature to be left only in the advanced-features page.

I tried my best to understand the order by which items appear in the `containers` page and tried to fit in.

I hope you accept this :)

### DIFF
--- a/docs/features/containers.md
+++ b/docs/features/containers.md
@@ -5,7 +5,6 @@
 Create and start any container using a Generic Container:
 
 ```javascript
-
 const { GenericContainer } = require("testcontainers");
 
 const container = await new GenericContainer("alpine").start();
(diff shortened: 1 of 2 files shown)


kiview (approved):
I think it is a useful improvement. Also highlights, how we would benefit from language agnostic docs on the conepts.

osher:
If you point me to the language-agnostic docs - I'll see if I can PR the same thing there :)

cristianrgreco:
Hi, 

> I believe it's a too important feature to be left only in the advanced-features page.

The docs for wait strategies aren't in the advanced page. They have their own top-level page

image

I'd rather avoid duplicating the docs for the time being. If I hear again that someone's missed the page then I'll revisit this. 

Thanks for raising 🙂

PR #849, opened 2024-10-24
Add support for reusing stopped containers

### LINKED ISSUE #783: testcontainer not reusing container if stopped
Hi, thanks for the great libraries. 
Why does Testcontainers not reuse containers if they are stopped? Is this a bug or some limitation? My expectation is that Testcontainers just start existing stopped containers instead of creating new ones

Thank you

cbrevik:
Closes #783 

I looked into this a bit and as suggested removing the status-filter in `fetchByLabel` will make it find the exited container. And then I had to check if it is running or not, and start it if not.

But I hit on a problem when using `@testcontainers/postgresql` that it could not find bound ports on an exited container that is being restarted:
```
node_modules/testcontainers/build/utils/bound-ports.js:14  throw new Error(`No port binding found for :${port}`)
```

The bound ports seems to be created from a an [inspection result when trying to reuse the container](https://github.com/testcontainers/testcontainers-node/blob/main/packages/testcontainers/src/generic-container/generic-container.ts#L133-L135).

And the ports in the inspection results seem to be mapped from the [`NetworkSettings.Ports` section when inspecting the container](https://github.com/testcontainers/testcontainers-node/blob/main/packages/testcontainers/src/utils/map-inspect-result.ts#L24).

What seems to be the underlying issue here though is that `NetworkSettings` seems to be reflecting what a running container has as current network settings. Which seems to be empty values when doing `docker inspect` on an exited container:

```
"NetworkSettings": {
    "Ports": {},
}
```

Compared to a running container:

```
"NetworkSettings": {
  "Ports": {
      "5432/tcp": [
          {
              "HostIp": "0.0.0.0",
              "HostPort": "54301"
          },
          {
              "HostIp": "::",
              "HostPort": "54301"
          }
      ]
  },
}
```

So basically I had to inspect the container again after re-starting it to get the correct bound ports and their IPs.

### DIFF
--- a/packages/testcontainers/src/container-runtime/clients/container/docker-container-client.ts
+++ b/packages/testcontainers/src/container-runtime/clients/container/docker-container-client.ts
@@ -35,7 +35,6 @@ export class DockerContainerClient implements ContainerClient {
       const containers = await this.dockerode.listContainers({
         limit: 1,
         filters: {
-          status: ["running"],
           label: [`${labelName}=${labelValue}`],
         },
       });
--- a/packages/testcontainers/src/generic-container/generic-container.ts
+++ b/packages/testcontainers/src/generic-container/generic-container.ts
@@ -130,6 +130,11 @@ export class GenericContainer implements TestContainer {
   private async reuseContainer(client: ContainerRuntimeClient, container: Container) {
     const inspectResult = await client.container.inspect(container);
     const mappedInspectResult = mapInspectResult(inspectResult);
+    if (!mappedInspectResult.state.running) {
+      log.debug("Reused container is not running, attempting to start it");
+      await client.container.start(container);
+    }
+
     const boundPorts = BoundPorts.fromInspectResult(client.info.containerRuntime.hostIps, mappedInspectResult).filter(
       this.exposedPorts
     );
--- a/packages/testcontainers/src/generic-container/generic-container.ts
+++ b/packages/testcontainers/src/generic-container/generic-container.ts
@@ -128,13 +128,15 @@ export class GenericContainer implements TestContainer {
   }
 
   private async reuseContainer(client: ContainerRuntimeClient, container: Container) {
-    const inspectResult = await client.container.inspect(container);
-    const mappedInspectResult = mapInspectResult(inspectResult);
-    if (!mappedInspectResult.state.running) {
+    let inspectResult = await client.container.inspect(container);
+    if (!inspectResult.State.Running) {
       log.debug("Reused container is not running, attempting to start it");
       await client.container.start(container);
+      // Refetch the inspect result to get the updated state
+      inspectResult = await client.container.inspect(container);
     }
 
+    const mappedInspectResult = mapInspectResult(inspectResult);
     const boundPorts = BoundPorts.fromInspectResult(client.info.containerRuntime.hostIps, mappedInspectResult).filter(
       this.exposedPorts
     );
--- a/packages/testcontainers/src/utils/map-inspect-result.ts
+++ b/packages/testcontainers/src/utils/map-inspect-result.ts
@@ -7,7 +7,10 @@ export function mapInspectResult(inspectResult: ContainerInspectInfo): InspectRe
   return {
     name: inspectResult.Name,
     hostname: inspectResult.Config.Hostname,
-    ports: mapPorts(inspectResult),
+    ports: {
+      ...mapHostConfigPortBindings(inspectResult),
+      ...mapPorts(inspectResult),
+    },
     healthCheckStatus: mapHealthCheckStatus(inspectResult),
     networkSettings: mapNetworkSettings(inspectResult),
     state: {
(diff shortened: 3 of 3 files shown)


cristianrgreco:
Nice, thanks for raising. Could you please add a test for the new functionality?

cbrevik:
Sure thing @cristianrgreco! Added a new test for reusing stopped (but not removed) containers 👍🏻 

Another thing I thought about which I'm unsure if should be handled explicitly or not, but the `status` you can filter for can be one of `created|restarting|running|removing|paused|exited|dead`. But I'm guessing if it is `dead` or `removing` while trying to start it, then `container.start` will just throw and that is fine?

cbrevik:
Also, I'm sure you wouldn't want to necessarily break existing API with re-use. But coming from using TestContainers in C# I was a bit surprised to find that enabling re-use [does not also implicitly disable cleanup of the container](https://github.com/testcontainers/testcontainers-dotnet/blob/dfb9f41fbfc62d2bc93cdcfb3dd098f4d8d9d954/src/Testcontainers/Builders/AbstractBuilder%604.cs#L63). Unsure what is default behaviour in other languages, like Java?

Maybe `WithCleanUp(bool)` is something that could be implemented on `GenericContainer`? Even if isn't something that is implicitly called.

cristianrgreco:
@cbrevik You can specify not to remove the container when stopping it with `await container.stop({ remove: false });`. Is this what you mean? I see you used it in the tests.

> Another thing I thought about which I'm unsure if should be handled explicitly or not, but the status you can filter for can be one of created|restarting|running|removing|paused|exited|dead. But I'm guessing if it is dead or removing while trying to start it, then container.start will just throw and that is fine?

This is a good point. It is probably best not to try to restart the old container if it is dead/removing. If we add this filter, then testcontainers will start a new container as expected. In fact not adding this filter could result in breaking changes. Would you be able to update the PR with a test for this?

cbrevik:
Yeah I've used the that way to disable removing it with `container.stop({ remove: false })`. What I'm suggesting is possibly making the default behaviour configurable when creating the container. I've made an issue #850 with a suggested implementation. Just a suggestion though, not something critical for me 😅 

The reason I thought it might be better to just let it throw is because either way one would have to manually clean it up probably (for `dead` at least). Else it'll either way hit an error with colliding reuse-hash. But I'll update with the filter for dead/removing + tests for that. 👍🏻

cbrevik:
@cristianrgreco I've added a filter for `dead` and `removing` when trying to reuse. But I am a bit unsure how to test for it. It seems `dead` is something that happens in the extreme, and `removing` will require hitting a race condition.

packages/testcontainers/src/generic-container/generic-container-reuse.test.ts
```diff
@@ -102,6 +110,40 @@ describe("GenericContainer reuse", () => {
...
+          .getById(container1.getId())
+          .inspect()
+          .then((i) => i.State.Status)
+          .catch(() => "failed"),
+      (status) => status === "exited",
+      () => "timeout",
+      500
+    );
+
+    expect(status).toBe("exited");
```
cbrevik, line 134: Just to explain this one. I'm unsure if I'm hitting on a race condition here, but without this check it seems like the `container1`'s `Status` is still `running` while trying to start `container2`. In other words, just awaiting `container1.stop({ remove: false })` seems to not be enough. 

If I set the status filter in `fetchByLabel` to just filter `running` containers, then the test is still green without this `IntervalRetry` status-check. But the test becomes red if I do this check (which it should do, because I'm testing for stopped containers).

cristianrgreco: > In other words, just awaiting container1.stop({ remove: false }) seems to not be enough.

You'll need to give a timeout to let Testcontainers wait for the container to be stop. By default the timeout is `0`, meaning Testcontainers will issue the stop command and move on (which is why immediately after the container is still in a `running` state

cbrevik: Oh, I've read the code wrong then, sorry 🙈 It seemed to just pipe the timeout value to `dockerode` which I thought was just to give the container a chance to gracefully shut down 😅

cristianrgreco: That's waiting for 10ms

cbrevik: Oh sorry, I just followed the code to here: https://github.com/testcontainers/testcontainers-node/blob/main/packages/testcontainers/src/container-runtime/clients/container/docker-container-client.ts#L128

Which lead to `dockerode`, and there `t` said "number of seconds" 😅 :

```
    interface ContainerStopOptions {
        signal?: string;
        /** Number of seconds to wait before killing the container */
        t?: number;
        abortSignal?: AbortSignal;
    }
```

Upping it to `10000` fixes it, thanks 👍🏻

cristianrgreco:
Looks like Podman doesn't support the `restarting` container state:

```
 ● GenericContainer reuse › should reuse the container

    (HTTP code 500) server error - unknown container state: restarting: invalid argument
```

We can either add a Podman specific implementation, or just forget about the filters

cbrevik:
@cristianrgreco how about just removing `restarting` from included filter? It makes sense to not attempt to start a container that is already being restarted? 🤔

cristianrgreco:
> @cristianrgreco how about just removing `restarting` from included filter? It makes sense to not attempt to start a container that is already being restarted? 🤔

Sure

packages/testcontainers/src/container-runtime/clients/container/types.ts
```diff
@@ -3,3 +3,7 @@ export type Environment = { [key in string]: string };
 export type ExecOptions = { workingDir: string; user: string; env: Environment; log: boolean };
 
 export type ExecResult = { output: string; exitCode: number };
+
+export const ContainerStatus = ["created", "restarting", "running", "removing", "paused", "exited", "dead"] as const;
```
cristianrgreco, line 7: I'd rename this to either `CONTAINER_STATUSES` or `containerStatuses`. Pascal case makes me think it's a type or a class

cbrevik: Done 👍🏻 Liked `CONTAINER_STATUSES`, good point

cristianrgreco (approved):

PR #873, opened 2024-11-24
feat(nats): added js and support arg without value

Naymi:


packages/modules/nats/src/nats-container.test.ts
```diff
@@ -67,6 +67,43 @@ describe("NatsContainer", () => {
   });
   // }
 
+  // JetStream {
+  it("should start with JetStream ", async () => {
+    // enable JS
+    const container = await new NatsContainer().withJS().start();
```
cristianrgreco, line 73: Is `JS` a well known acronym for Jetstream for Nats users?

jonjomckay:
This would be helpful for us - we use JetStream, but we're unable to use `NatsContainer` at all at the moment, as we can't use `withArg` with a single parameter.

cristianrgreco:
Hi @jonjomckay, I'm not familiar with NATS. If you were using the `NatsContainer` and came across a method `withJS`, would you understand that it is for enabling jetstream?

jonjomckay:
@cristianrgreco `withJetStream` would be clearer to me. `JS` is probably a bit confusing, especially with this being a JavaScript library!

cristianrgreco:
IDK if @Naymi is going to address the issues with the PR, but @jonjomckay if you want this feature then I'd happily accept a PR. A copy of this one is fine with `withJS` changed to `withJetStream`, with lint issues addressed.

jonjomckay:
@cristianrgreco @Naymi I've opened up another PR with the linting and naming fixed: https://github.com/testcontainers/testcontainers-node/pull/877

Issue #842, opened 2024-10-07
stderr output


jeronimo-planet:
When we do an exec, we get an { ouput, exitCode }.
It would be nice if we could get a { stdout, stderr, exitCode }.
Do you guys think it's possible?

kiview:
This is the result type in tc-java, so that would be a quite alinged implementation:
```java
class ExecResult {
    int exitCode;
    String stdout;
    String stderr;
}
```

cristianrgreco:
PR welcome. I'd suggest keeping the existing `output` field, and adding the `stdout`/`stderr` fields. Otherwise we lose the ability to log the output in time order. Removing the `output` would also be a breaking change.

cristianrgreco:
This will be released shortly

PR #874, opened 2024-11-24
Add stderr output for exec function

chungjung-d:
Add stderr output for exec function on issue #842 

I chose not to add an outputMode option to the existing exec function because it would compromise the integrity of the current ExecResult type. Instead, I implemented a new execVerbose function to provide separate handling and output for stdout and stderr.

### DIFF
--- a/docs/features/containers.md
+++ b/docs/features/containers.md
@@ -537,6 +537,26 @@ const { output, exitCode } = await container.exec(["echo", "hello", "world"], {
 	}
 });
 ```
+To handle cases where you need separate outputs for `stdout` and `stderr`, you can use the `execVerbose` method. 
+It functions similarly to `exec`, but provides detailed output including `stdout`, `stderr`, and `exitCode`.
+
+```javascript
+const container = await new GenericContainer("alpine")
+  .withCommand(["sleep", "infinity"])
+  .start();
+
+const { stdout, stderr, exitCode } = await container.execVerbose(["echo", "hello", "world"], {
+	workingDir: "/app/src/",
+	user: "1000:1000",
+	env: {
+		"VAR1": "enabled",
+		"VAR2": "/app/debug.log",
+	}
+});
+```
+Use `execVerbose` when you require more granular control over command outputs,
+while retaining similar options and functionality as `exec`.
+
 
 ## Streaming logs
 
--- a/packages/testcontainers/src/container-runtime/clients/container/container-client.ts
+++ b/packages/testcontainers/src/container-runtime/clients/container/container-client.ts
@@ -8,27 +8,46 @@ import Dockerode, {
 } from "dockerode";
 import { Readable } from "stream";
 import { ContainerStatus, ExecOptions, ExecResult } from "./types";
+import { ExecVerboseResult } from "../../../types";
 
 export interface ContainerClient {
   dockerode: Dockerode;
+
   getById(id: string): Container;
+
   fetchByLabel(
     labelName: string,
     labelValue: string,
     opts?: { status?: ContainerStatus[] }
   ): Promise<Container | undefined>;
+
   fetchArchive(container: Container, path: string): Promise<NodeJS.ReadableStream>;
+
   putArchive(container: Dockerode.Container, stream: Readable, path: string): Promise<void>;
+
   list(): Promise<ContainerInfo[]>;
+
   create(opts: ContainerCreateOptions): Promise<Container>;
+
   start(container: Container): Promise<void>;
+
   inspect(container: Container): Promise<ContainerInspectInfo>;
+
   stop(container: Container, opts?: { timeout: number }): Promise<void>;
+
   attach(container: Container): Promise<Readable>;
+
   logs(container: Container, opts?: ContainerLogsOptions): Promise<Readable>;
+
   exec(container: Container, command: string[], opts?: Partial<ExecOptions>): Promise<ExecResult>;
+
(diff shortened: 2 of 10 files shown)


packages/testcontainers/src/container-runtime/clients/container/container-client.ts
```diff
@@ -8,27 +8,46 @@ import Dockerode, {
 } from "dockerode";
 import { Readable } from "stream";
 import { ContainerStatus, ExecOptions, ExecResult } from "./types";
+import { ExecVerboseResult } from "../../../types";
 
 export interface ContainerClient {
   dockerode: Dockerode;
+
```
cristianrgreco, line 15: Could you pls remove the newline changes to `container-client.ts` and `test-container.ts` to reduce the size of the PR?

docs/features/containers.md
```diff
@@ -537,6 +537,26 @@ const { output, exitCode } = await container.exec(["echo", "hello", "world"], {
 	}
 });
 ```
+To handle cases where you need separate outputs for `stdout` and `stderr`, you can use the `execVerbose` method.
```
cristianrgreco, line 540: Instead of adding a 2nd `exec` command, would it be possible to add the `stdout` and `stderr` fields to the existing `exec` response? It's a lot of code being added for a slight change

chungjung-d: I initially attempted the approach you suggested. However, adopting this method would result in the ExecResult type being defined as either:

```typescript
type ExecResult = { output: string; exitCode: number } | { stdout: string; stderr: string; exitCode: number };
```
or:

```typescript
type ExecResult = { output?: string; stdout?: string; stderr?: string; exitCode: number };
```
Both options are technically feasible, but they introduce potential confusion and require modifications to the existing test code that uses the exec functionality. This is because the current test cases assume that the output field is always present. Changing the type as proposed makes it challenging for the TypeScript compiler to validate whether output exists or not.

If you're okay with proceeding despite these issues, I can move forward with your suggestion and update the relevant test cases accordingly. Please let me know your preference.

cristianrgreco: Understood, I also prefer the separate `exec` method instead of that return type. Is it not possible to have them all?

```ts
type ExecResult = { output: string; stdout: string; stderr: string; exitCode: number };
```

chungjung-d: Understood. I have rewritten the exec code to always return an ExecResult with the following type:

```typescript
type ExecResult = { output: string; stdout: string; stderr: string; exitCode: number };
```
Additionally, I have added a few test cases for this implementation. Please review them and let me know your thoughts.

packages/testcontainers/src/container-runtime/clients/container/podman-container-client.ts
```diff
@@ -22,32 +23,52 @@ export class PodmanContainerClient extends DockerContainerClient {
...
+    const stdoutChunks: string[] = [];
+    const stderrChunks: string[] = [];
+
     try {
       if (opts?.log) {
         log.debug(`Execing container with command "${command.join(" ")}"...`, { containerId: container.id });
       }
 
       const exec = await container.exec(execOptions);
       const stream = await this.demuxStream(container.id, await exec.start({ stdin: true, Detach: false, Tty: true }));
```
cristianrgreco, line 36: Perhaps it is no longer necessary to demux the stream here, if you doing it on line 41. If this line can be removed, then I don't think there is any difference between the Docker and Podman implementations? If so we can remove the whole `podman-container-client` class 🚀

packages/testcontainers/src/container-runtime/clients/container/docker-container-client.ts
```diff
@@ -201,34 +201,55 @@ export class DockerContainerClient implements ContainerClient {
...
+      const stderr = stderrChunks.join("");
+
       if (opts?.log) {
         log.debug(`Execed container with command "${command.join(" ")}"...`, { containerId: container.id });
       }
-      return { output, exitCode };
+      return { output, stdout, stderr, exitCode };
     } catch (err) {
-      log.error(`Failed to exec container with command "${command.join(" ")}": ${err}: ${chunks.join("")}`, {
+      log.error(`Failed to exec container with command "${command.join(" ")}": ${err}: ${stderrChunks.join("")}`, {
```
cristianrgreco, line 252: Perhaps this should still be `chunks`, I guess that an exec could fail without error output, in which case it would be useful to see all the output?

packages/testcontainers/src/generic-container/abstract-started-container.ts
```diff
@@ -83,6 +91,10 @@ export class AbstractStartedContainer implements StartedTestContainer {
     return this.startedTestContainer.exec(command, opts);
   }
 
+  public execVerbose(command: string | string[], opts?: Partial<ExecOptions>): Promise<ExecVerboseResult> {
```
cristianrgreco, line 94: This can now be removed

packages/testcontainers/src/container-runtime/clients/container/podman-container-client.ts
chungjung-d: I agree as well. I just disabled the exec method in the Podman client implementation and replaced it with the Docker client implementation's exec method. I confirmed that the test cases pass without any issues.

Should I proceed with modifying the code to remove the Podman client implementation and commit the changes?

cristianrgreco: > Should I proceed with modifying the code to remove the Podman client implementation and commit the changes?

That would be great

cristianrgreco:
Thanks @chungjung-d !

PR #882, opened 2024-12-11
MySQL add `isRoot` parameter to `executeQuery`

jeengbe:
It's convenient to be able to perform root operations directly on the container. The parameter is optional and defaults to the current behaviour, so it's no breaking change.

### DIFF
--- a/packages/modules/mysql/src/mysql-container.test.ts
+++ b/packages/modules/mysql/src/mysql-container.test.ts
@@ -105,11 +105,11 @@ describe("MySqlContainer", () => {
 
     // Test non-root user
     const queryResult = await container.executeQuery("SELECT CURRENT_USER() as user");
-    expect(queryResult).toEqual(expect.stringContaining("user\ncustomUsername@%\n"));
+    expect(queryResult).toEqual(expect.stringContaining("user\ncustomUsername"));
 
     // Test root user
     const rootQueryResult = await container.executeQuery("SELECT CURRENT_USER() as user", [], true);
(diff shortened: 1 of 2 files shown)


packages/modules/mysql/src/mysql-container.test.ts
```diff
@@ -94,9 +94,14 @@ describe("MySqlContainer", () => {
   it("should execute a query and return the result", async () => {
     const container = await new MySqlContainer().start();
 
+    // Test non-root user
     const queryResult = await container.executeQuery("SELECT 1 as res");
     expect(queryResult).toEqual(expect.stringContaining("res\n1\n"));
 
+    // Test root user
+    const rootQueryResult = await container.executeQuery("SELECT CURRENT_USER() as user");
```
cristianrgreco, line 102: How does this test test the new `isRoot` parameter?

jeengbe: Here we go, fixed. I apologise for the incomplete PR, I don't know what happened.

I kept test root and non-root in the same test as is done in the "should work with database URI" test

cristianrgreco (approved):

cristianrgreco:
@jeengbe The new test is failing, could you take a look?

jeengbe:
It's weird that this test is failing; this implies that `SELECT CURRENT_USER()` returns `customUsername@%` for the user, but `root@localhost` for root. On my machine, both return `@%`, and I don't see why they wouldn't. Perhaps some difference in the environment, the mysql version being run, or something. Nonetheless, the host is not the essence of this test, so I have just removed it from the assertion.

PR #886, opened 2025-01-06
feat(file-lock): Added configurable retries for file-lock

GiDW:
When test processes with isolated testcontainers are running in parallel they become very slow to start because of the file-lock retry strategy in testcontainers. 

`proper-filelock` uses `node-retry` ([repo](https://github.com/tim-kos/node-retry/tree/v0.12.0))  and exposes options to configure `node-retry` behavior. The current option is `{ retries: { forever: true } }`. This leaves the rest on defaults (`factor` `2`, `minTimeout` `1000`, `maxTimeout` `infinity`).

This PR allows the user to configure some parts of the file lock retry strategy.

Example

Multiple tests are started in parallel with jest as different processes (default behavior). They start at the same time (default jest behavior). 
Each test uses a `GenericContainer` and tries to start it. Because the process are started at the same time and use the same exponential back-off algorithm (node-retry) most of the time the processes check again on the same moment. This causes exponentially longer tests depending on how many are running in parallel. 

Tweaking the retry strategy to lower the `minTimeout` and `factor` can result in serious speed-up.

### DIFF
--- a/packages/testcontainers/package.json
+++ b/packages/testcontainers/package.json
@@ -32,6 +32,7 @@
   "dependencies": {
     "@balena/dockerignore": "^1.0.2",
     "@types/dockerode": "^3.3.29",
+    "@types/retry": "^0.12.5",
     "archiver": "^7.0.1",
     "async-lock": "^1.4.1",
     "byline": "^5.0.0",
--- a/packages/testcontainers/src/common/file-lock.ts
+++ b/packages/testcontainers/src/common/file-lock.ts
@@ -2,14 +2,19 @@ import path from "path";
 import { writeFile } from "fs/promises";
 import lockFile from "proper-lockfile";
 import { log } from "./logger";
+import type { WrapOptions } from "retry";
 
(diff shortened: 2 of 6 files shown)


cristianrgreco:
Thanks for the PR @GiDW. I wonder instead of implementing the ability for users to configure the retry mechanism, if it would be better to just improve the defaults?

GiDW:
My recommendation would be something like

```typescript
{
  factor: 1.1,
  minTimeout: 500,
  maxTimeout: 2500
}
```

Based on arbitrary testing. In my case, the operation where the lock is held took around 1 to 2 seconds.

My current setup is macOs with podman, so I have the `TESTCONTAINERS_RYUK_DISABLED=true` environment variable set. 
I have no idea what the impact would be on other setups.

cristianrgreco:
The CICD test suites are pretty comprehensive, they run podman/docker and 2 node versions. It may be worth raising a PR with the improved defaults and we can assess. WDYT?

cristianrgreco:
Replaced by https://github.com/testcontainers/testcontainers-node/pull/890

PR #890, opened 2025-01-08
Optimise file-lock retry options

GiDW:
When processes with testcontainers are running in parallel they become very slow to start because of the file-lock retry strategy in testcontainers.

`proper-filelock` uses `node-retry` ([repo](https://github.com/tim-kos/node-retry/tree/v0.12.0)) and exposes options to configure `node-retry` behavior. The current option is `{ retries: { forever: true } }`. This leaves the rest on defaults (factor 2, minTimeout 1000, maxTimeout infinity).

Set `file-lock` retry options to retry faster and more frequently, avoiding timeouts that become too long. 

```typescript
{
  retries: {
    forever: true,
    factor: 1,
    minTimeout: 500,
    maxTimeout: 3000,
    randomize: true
  }
}
```

### DIFF
--- a/packages/testcontainers/src/common/file-lock.ts
+++ b/packages/testcontainers/src/common/file-lock.ts
@@ -9,7 +9,9 @@ export async function withFileLock<T>(fileName: string, fn: () => T): Promise<T>
   let releaseLockFn;
   try {
     log.debug(`Acquiring lock file "${file}"...`);
-    releaseLockFn = await lockFile.lock(file, { retries: { forever: true } });
+    releaseLockFn = await lockFile.lock(file, {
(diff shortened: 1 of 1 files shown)


GiDW:
@cristianrgreco Could you retry this run? 🙏  
Test failed on installing dependencies due to a network error. Seems unrelated to the actual change.

cristianrgreco:
@GiDW I don't see any difference in test times from my side, so if for your case this is a big help them I'm happy to merge it. Adding some randomness makes sense IMO because otherwise the processes will check for locks at the same time. Happy to merge if you are.

GiDW:
@cristianrgreco Yes, this helps my case. If you are willing to merge it, please do 🙂

PR #893, opened 2025-01-10
Add MinIO module

stscoundrel:
Add MinIO module. Already available for other TestContainers language versions, like Java, Go & .NET.

For successful CI run, see [PR in this fork](https://github.com/stscoundrel/testcontainers-node/pull/6)

### DIFF
--- a/packages/modules/minio/src/minio-container.ts
+++ b/packages/modules/minio/src/minio-container.ts
@@ -1,0 +1,65 @@
+import { AbstractStartedContainer, GenericContainer, Wait, type StartedTestContainer } from "testcontainers";
+
+const MINIO_PORT = 9000;
+const MINIO_UI_PORT = 9001;
+
+export class MinioContainer extends GenericContainer {
+  private username = "minioadmin";
+  private password = "minioadmin";
+
+  constructor(image = "minio/minio:RELEASE.2024-12-13T22-19-12Z") {
+    super(image);
+    this.withExposedPorts(MINIO_PORT, MINIO_UI_PORT);
+    this.withWaitStrategy(Wait.forAll([Wait.forHttp("/minio/health/live", MINIO_PORT)]));
+    this.withCommand(["server", "--console-address", `:${MINIO_UI_PORT}`, "/data"]);
+  }
+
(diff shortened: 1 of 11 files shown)


packages/modules/minio/src/minio-container.ts
```diff
@@ -0,0 +1,70 @@
...
+  private readonly port: number;
+  private readonly uiPort: number;
+
+  constructor(
+    startedTestContainer: StartedTestContainer,
+    private readonly username: string,
+    private readonly password: string
+  ) {
+    super(startedTestContainer);
+    this.port = startedTestContainer.getMappedPort(MINIO_PORT);
```
cristianrgreco, line 47: These will need to be computed on demand to support the container restarting, as it restarts with different ports

packages/modules/minio/tsconfig.build.json
```diff
@@ -0,0 +1,13 @@
+{
+  "extends": "./tsconfig.json",
+  "exclude": [
+    "build",
+    "jest.config.ts",
+    "src/**/*.test.ts"
```
cristianrgreco, line 6: We should exclude the dummy file here too

packages/modules/minio/src/minio-container.ts
stscoundrel: Is this something that has changed recently? I'm pretty sure there are existing modules with this pattern.

packages/modules/minio/tsconfig.build.json
stscoundrel: Good point, added.

packages/modules/minio/src/minio-container.ts
stscoundrel: Moved these to be resolved in `getPort`/`getUiPort` whenever they're called.

cristianrgreco: Not all modules support restarting the container. You could see for yourself by adding a test for restarting the container if you'd like

stscoundrel:
Looks like traditional flaky run in Weaviate. Maybe give the pipeline a retry, @cristianrgreco ?

PR #894, opened 2025-01-12
Add EventStoreDB module

botflux:
This PR adds an EventStoreDB module, the implementation is similar to the [dotnet implementation](https://github.com/testcontainers/testcontainers-dotnet/tree/develop/src/Testcontainers.EventStoreDb).

PR #901, opened 2025-01-22
Rename EventStoreDB to KurrentDB

botflux:
I've changed EventStoreDB to Kurrent to match their re-branding as pointed by https://github.com/testcontainers/testcontainers-node/pull/894#issuecomment-2605937410 .


cristianrgreco:
Build is failing because the kurrent lib doesn't support node < 20, which is fine as we'll match that soon with #938.

@joebowbeer / @botflux - should the published package be @testcontainers/kurrent or @testcontainers/kurrentdb? From what I'm reading it seems it should be the latter.

PR #902, opened 2025-01-23
Add support for Docker credential providers which do not implement list command

glebbash:
Fixes https://github.com/testcontainers/testcontainers-node/issues/739

### DIFF
--- a/packages/testcontainers/src/container-runtime/auth/credential-provider.test.ts
+++ b/packages/testcontainers/src/container-runtime/auth/credential-provider.test.ts
@@ -104,6 +104,16 @@ describe("CredentialProvider", () => {
     );
   });
 
+  it("should not throw when list credentials command is not implemented", async () => {
+    mockExec.mockImplementationOnce((command, callback) => {
+      return callback(new Error(), null, "list is unimplemented\n");
+    });
+
+    const credentials = await credentialProvider.getAuthConfig("registry", containerRuntimeConfig);
+
+    expect(credentials).toBeUndefined();
+  });
+
   it("should throw when get credentials fails", async () => {
     mockExecReturns(JSON.stringify({ registry: "username" }));
     mockSpawnReturns(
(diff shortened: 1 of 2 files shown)


cristianrgreco:
Hi @glebbash, you've added a test `should not throw when list credentials command is not implemented`, but it's not clear why not throwing when listing credentials is useful or required. Is it possible to instead add a test verifying an actual output? E.g a file containing GCR credentials now works?

glebbash:
I am not sure what you mean about a file with GCR credentials and what the expected output should be. As for reasons why this is required I think it's pretty clear - if command is not implemented it should just use defaults.

Adding a link to the issue in a comment before the condition could make it clearer but I just wanted to make the minimum amount of changes to support `docker-credential-gcr`.

cristianrgreco:
I mean, this PR supposedly fixes the resolution of GCR credentials, could you show that? I would look at the other tests for examples. A test which shows that one part of the resolution logic returns an empty array doesn't mean anything

glebbash:
This PR does not fix resolution of GCR credentials. What it does is make testcontainers usable for whoever happens to be using `docker-credential-gcr`. I really don't care about credentials and I don't want to dig into this repo to figure out all the details.

I was patching the lib's compiled JS source to make it not throw the `An error occurred listing credentials` error. And here I copied the changes to the TypeScript source.

I don't know if this supports private GCR registries or whatever, the only thing I know is that it fixes the issue described in #739.

glebbash:
@cristianrgreco if you have a better test in mind feel free to update the PR or fix this issue in different PR.

cristianrgreco:
I think it's my misunderstanding TBH. The PR said it added support for docker-credential-gcr so I assumed it would support resolving GCR credentials. I didn't realise that supporting docker-credential-gcr actually involves just not throwing when running the list command 😄

>  I really don't care about credentials and I don't want to dig into this repo to figure out all the details.

Thought this was a bit rude TBH. I work on this project in my free time. If I propose a suggestion and the response is effectively "I don't care and I can't be asked to find out", it doesn't particularly motivate me to work on it.

> the only thing I know is that it fixes the issue described in #739 which has 5 +1s so I don't think that it's just my setup that is affected

This project has a lot of moving parts and 2M downloads/month. Five +1s on a year old issue isn't exactly concerning, but thanks for the contribution. 

> if you have a better test in mind feel free to update the PR or fix this issue in different PR.

Thanks for the advice 😅

cristianrgreco (approved):

glebbash:
Definitely did not intend to be rude. Was just trying to describe the situation in multiple ways :)

And thanks for your work on the lib 👍

PR #906, opened 2025-01-29
Add Valkey module

mdodsworth:
…ntainer

Adding support for [Valkey](https://valkey.io/). Container and tests are based largely on the existing redis setup.

### DIFF
--- a/packages/modules/valkey/src/valkey-container.ts
+++ b/packages/modules/valkey/src/valkey-container.ts
@@ -31,6 +31,12 @@ export class ValkeyContainer extends GenericContainer {
     return this;
   }
 
+  protected override async containerStarted(container: StartedTestContainer): Promise<void> {
+    if (this.initialImportScriptFile) {
+      await this.importInitialData(container);
+    }
+  }
+
   public override async start(): Promise<StartedValkeyContainer> {
     this.withCommand([
       "valkey-server",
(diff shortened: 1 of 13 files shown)


packages/modules/valkey/src/valkey-container.ts
```diff
@@ -0,0 +1,103 @@
...
+        },
+        {
+          mode: 777,
+          source: path.join(__dirname, "import.sh"),
+          target: "/tmp/import.sh",
+        },
+      ]);
+    }
+    const startedContainer = new StartedValkeyContainer(await super.start(), this.password);
+    if (this.initialImportScriptFile) await this.importInitialData(startedContainer);
```
cristianrgreco, line 58: I know that this is copied from the Redis container code, but this can be refactored by moving this line into the `containerStarted` callback

mdodsworth: Will work for ❤️s.
Sure thing, I'll fix them both up

mdodsworth:
@cristianrgreco changes are ready for review. Thanks fro the input

cristianrgreco:
Thanks @mdodsworth. I had a thought. If Valkey is a drop-in replacement for Redis then could we get away with this?

```ts
class ValkeyContainer extends RedisContainer {
  constructor(image = "valkey/valkey:8.0") {
    super(image);
  }

  // maybe need to override the start method to return a `StartedValkeyContainer`, 
  // but even this just calls `super(...)` ?
}
```

cristianrgreco (approved):

mdodsworth:
> Thanks @mdodsworth. I had a thought.
> 
> If Valkey is a drop-in replacement for Redis then could we get away with this?
> 
> ```ts
> class ValkeyContainer extends RedisContainer {
>   constructor(image = "valkey/valkey:8.0") {
>     super(image);
>   }
> 
>   // maybe need to override the start method to return a `StartedValkeyContainer`, 
>   // but even this just calls `super(...)` ?
> }
> ```
> 
> EDIT: I think it may complicate things in terms of how the modules depend on each other.

Yeah, my feeling was that the functionality may _eventually_ diverge, so having it be separate container would be worthwhile

PR #910, opened 2025-02-13
Add CockroachDB module

leourbina:
# Adds CockroachDB Support

This is mainly the same as the postgres container, with a few minor details:

* It uses CRDB's `--insecure` flag to avoid needing to deal with certs for local testing/connections
* Due to the above, no password is set

### DIFF
--- a/mkdocs.yml
+++ b/mkdocs.yml
@@ -47,6 +47,7 @@ nav:
       - Cassandra: modules/cassandra.md
       - ChromaDB: modules/chromadb.md
       - Couchbase: modules/couchbase.md
+      - CockroachDB: modules/cockroachdb.md
       - Elasticsearch: modules/elasticsearch.md
       - EventStoreDB: modules/eventstoredb.md
       - GCloud: modules/gcloud.md
--- a/packages/modules/cockroachdb/src/cockroachdb-container.test.ts
+++ b/packages/modules/cockroachdb/src/cockroachdb-container.test.ts
@@ -8,8 +8,6 @@ describe("CockroachDbContainer", () => {
   it("should connect and return a query result", async () => {
     const container = await new CockroachDbContainer().start();
 
-    console.log(container.getDatabase(), container.getHost(), container.getPort());
-
     const client = new Client({
       host: container.getHost(),
       port: container.getPort(),
--- a/packages/modules/cockroachdb/src/cockroachdb-container.test.ts
+++ b/packages/modules/cockroachdb/src/cockroachdb-container.test.ts
@@ -101,4 +101,15 @@ describe("CockroachDbContainer", () => {
     await client.end();
     await container.stop();
   });
+
+  it("should allow custom healthcheck", async () => {
+    const container = new CockroachDbContainer().withHealthCheck({
+      test: ["CMD-SHELL", "exit 1"],
+      interval: 100,
+      retries: 0,
+      timeout: 0,
+    });
+
+    await expect(() => container.start()).rejects.toThrow();
+  });
 });
--- a/packages/modules/cockroachdb/src/cockroachdb-container.test.ts
+++ b/packages/modules/cockroachdb/src/cockroachdb-container.test.ts
@@ -1,0 +1,108 @@
+import { Client } from "pg";
+import { CockroachDbContainer } from "./cockroachdb-container";
+
+describe("CockroachDbContainer", () => {
+  jest.setTimeout(180_000);
+
+  // connect {
+  it("should connect and return a query result", async () => {
+    const container = await new CockroachDbContainer().start();
+
+    console.log(container.getDatabase(), container.getHost(), container.getPort());
+
+    const client = new Client({
+      host: container.getHost(),
+      port: container.getPort(),
+      database: container.getDatabase(),
+      user: container.getUsername(),
+      ssl: false,
+    });
+
+    await client.connect();
+
+    const result = await client.query("SELECT 1");
+    expect(result.rows[0]).toEqual({ "?column?": "1" });
+
+    await client.end();
+    await container.stop();
+  });
+  // }
+
+  // uriConnect {
+  it("should work with database URI", async () => {
+    const container = await new CockroachDbContainer().start();
+
+    const client = new Client({
+      connectionString: container.getConnectionUri(),
+    });
+    await client.connect();
+
+    const result = await client.query("SELECT 1");
+    expect(result.rows[0]).toEqual({ "?column?": "1" });
+
+    await client.end();
+    await container.stop();
+  });
+  // }
+
+  // setDatabase {
+  it("should set database", async () => {
+    const container = await new CockroachDbContainer().withDatabase("custom_database").start();
+
+    const client = new Client({
+      host: container.getHost(),
+      port: container.getPort(),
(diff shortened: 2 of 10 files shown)


mkdocs.yml
```diff
@@ -47,6 +47,7 @@ nav:
       - Cassandra: modules/cassandra.md
       - ChromaDB: modules/chromadb.md
       - Couchbase: modules/couchbase.md
+      - CockroachDB: modules/cockroachdb.md
```
cristianrgreco, line 50: Let's keep this list alphabetical 🙂

leourbina: Haha - I could swear this was alphabetical 🙃

packages/modules/cockroachdb/src/cockroachdb-container.ts
```diff
@@ -0,0 +1,71 @@
+import { AbstractStartedContainer, GenericContainer, StartedTestContainer, Wait } from "testcontainers";
+
+const COCKROACH_PORT = 26257;
+
+export class CockroachDbContainer extends GenericContainer {
+  private database = "test";
+  private username = "test";
```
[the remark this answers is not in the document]
leourbina: @cristianrgreco do these need to be addressed? I'm following exactly the same pattern as the postgres module.

mkdocs.yml
cristianrgreco: Haha I had to double check cos you clearly made effort to keep it alphabetical 😆 Thanks for sorting it

leourbina:
@cristianrgreco Let me know what, if anything, you need from me here to get this merged.

packages/modules/cockroachdb/src/cockroachdb-container.ts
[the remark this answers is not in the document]
cristianrgreco: They don't need to be addressed

packages/modules/cockroachdb/src/cockroachdb-container.ts
```diff
@@ -0,0 +1,71 @@
...
+export class StartedCockroachDbContainer extends AbstractStartedContainer {
+  private readonly port: number;
+
+  constructor(
+    startedTestContainer: StartedTestContainer,
+    private readonly database: string,
+    private readonly username: string
+  ) {
+    super(startedTestContainer);
+    this.port = startedTestContainer.getMappedPort(COCKROACH_PORT);
```
cristianrgreco, line 45: This needs to be calculated every time `getPort()` is called because the container can be restarted and get given a different port. E.g:

```js
const container = await new CockroachDbContainer().start();
const port = container.getFirstMappedPort();
await container.restart();
port !== container.getFirstMappedPort()
```

leourbina: Also, worth pointing out that currently this is not working for the postgres container. I'll make a follow up PR to fix it.

leourbina:
NP! Just addressed the latest comment and added the test. I had tweak the wait condition to be more robust and ensure that it would work across restarts. Thanks for the thorough review.

packages/modules/cockroachdb/src/cockroachdb-container.ts
```diff
@@ -0,0 +1,79 @@
...
+  }
+
+  public getPort(): number {
+    this.port = this.startedTestContainer.getMappedPort(COCKROACH_PORT);
```
cristianrgreco, line 56: Could just `return this.startedTestContainer.getMappedPort(COCKROACH_PORT);` and remove `this.port`

packages/modules/cockroachdb/src/cockroachdb-container.test.ts
```diff
@@ -0,0 +1,110 @@
...
+
+    const result = await client.query("SELECT current_user");
+    expect(result.rows[0]).toEqual({ current_user: "custom_username" });
+
+    await client.end();
+    await container.stop();
+  });
+  // }
+
+  // workWithRestartedContainer {
```
cristianrgreco, line 88: This is for documentation examples only, we don't need it here

leourbina: Removing

leourbina:
@cristianrgreco ok, I think it should be good now.

packages/modules/cockroachdb/src/cockroachdb-container.ts
leourbina: haha good point 🤦

cristianrgreco (approved):

leourbina:
Urgh... ok I think the healthcheck is not as robust as I actually thought. I'll look into it.

leourbina:
Hey @cristianrgreco I'm running into the following issue:

Whenever the container starts for the first time, it will run the init scripts. These need to finish running as they setup the user/db and need to be waited for before running anything against it. The only reliable way that I've figured out to detect this is to wait for the log message as so:

```typescript
    this.withExposedPorts(COCKROACH_PORT)
      .withCommand(["start-single-node", "--insecure"])
      .withWaitStrategy(Wait.forLogMessage(/.*end running init files.*/))
      .withStartupTimeout(120_000);
```

This is effectively the same thing that the postgres container is doing [here](https://github.com/testcontainers/testcontainers-node/blob/main/packages/modules/postgresql/src/postgresql-container.ts#L13). 

However, on restarts the init scripts don't run again. I tried fixing it using a healthcheck (as seen on this PR), but I think it is working on my machine as the timing lines up, but based on the test results it is not reliably waiting for the init scripts.

As it is currently written the restart issue we're seeing on the Cockroachdb container exists too on the postgres container. I'm not sure if there are other better ways to ensure that the container is ready both on the first start as well as after a restart. Any ideas here?

leourbina:
@cristianrgreco currently the postgres container does not work across restarts either as it suffers from the same issue I pointed above. Namely, the wait condition does not work across restarts. Any chance we can merge this as is? Or do you think we need to fix both CRDB and Postgres as part of this PR?

cristianrgreco:
Hi @leourbina, I've raised a PR for fixing the PG container across restarts: https://github.com/testcontainers/testcontainers-node/pull/914

For the Cockroach container I see in the testcontainers-java implementation that they're calling an endpoint ending with `init_success`, which I guess returns 200 once the init is finished - this sounds like what you need:

```java
waitStrategy.withStrategy(
    Wait.forHttp("/health").forPort(REST_API_PORT).forStatusCode(200).withStartupTimeout(Duration.ofMinutes(1))
);
if (this.isVersionGreaterThanOrEqualTo221) {
    waitStrategy.withStrategy(Wait.forSuccessfulCommand("[ -f ./init_success ] || { exit 1; }"));
}
```

cristianrgreco:
I've committed a change similar to the PG container healthcheck. Seems to work locally and it allows us to make the interval more frequent for faster startups. Will see what the pipeline sayd

cristianrgreco:
Latest changes should be final. @leourbina is there anything more or is this PR good to go?

cristianrgreco (approved):

cristianrgreco:
Thanks @leourbina for your contribution!

leourbina:
Thank you @cristianrgreco!

PR #914, opened 2025-03-01
Fix PostgreSQL container restart

cristianrgreco:


leourbina:
@cristianrgreco thanks for fixing this!

Issue #867, opened 2024-11-15
Remove default images and versions from modules


cristianrgreco:
We currently have default images and versions defined for all modules. This means users can start a container effortlessly like `new PostgreSqlContainer().start()`. This works well when the module is first created and is using the latest versions, but as these versions get out of date, it is not possible for Testcontainers to push a change increasing the default module versions, as it could potentially break the builds of everyone using the default values. It is not feasible to continuously update and push out breaking changes.

The change will simply make the image name and version mandatory, e.g `new PostgreSqlContainer("postgres:X.Y.Z").start()`.

joebowbeer:
NOTE to check/update the installation examples, too:

https://github.com/testcontainers/community-module-registry

cristianrgreco:
Will be completed as part of https://github.com/testcontainers/testcontainers-node/pull/938

PR #917, opened 2025-03-05
Update docs to demonstrate specifying Qdrant version

Anush008:
Hey 👋. Greetings from Qdrant.

`QdrantContainer` defaults to a specific version of Qdrant.

We had users coming from https://node.testcontainers.org/modules/qdrant/ not knowing how to specify a custom version. So this PR helps with demonstrating that.

### DIFF
--- a/packages/modules/qdrant/src/qdrant-container.test.ts
+++ b/packages/modules/qdrant/src/qdrant-container.test.ts
@@ -42,7 +42,7 @@ describe("QdrantContainer", () => {
       apiKey: "INVALID_KEY_" + crypto.randomUUID(),
     });
 
-    expect(client.getCollections()).rejects.toThrow("Forbidden");
+    expect(client.getCollections()).rejects.toThrow("Unauthorized");
 
     await container.stop();
   });
@@ -71,7 +71,7 @@ describe("QdrantContainer", () => {
       apiKey: "INVALID_KEY_" + crypto.randomUUID(),
     });
 
-    expect(client.getCollections()).rejects.toThrow("Forbidden");
+    expect(client.getCollections()).rejects.toThrow("Unauthorized");
 
     await container.stop();
   });
--- a/packages/modules/qdrant/src/qdrant-container.test.ts
+++ b/packages/modules/qdrant/src/qdrant-container.test.ts
@@ -49,7 +49,9 @@ describe("QdrantContainer", () => {
 
   // connectQdrantWithConfig {
   it("should work with config files - valid API key", async () => {
-    const container = await new QdrantContainer("qdrant/qdrant:v1.13.4").withConfigFile(path.resolve(__dirname, "test_config.yaml")).start();
+    const container = await new QdrantContainer("qdrant/qdrant:v1.13.4")
+      .withConfigFile(path.resolve(__dirname, "test_config.yaml"))
+      .start();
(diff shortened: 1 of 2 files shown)


packages/modules/qdrant/src/qdrant-container.test.ts
```diff
@@ -8,7 +8,7 @@ describe("QdrantContainer", () => {
 
   // connectQdrantSimple {
   it("should connect to the client", async () => {
-    const container = await new QdrantContainer().start();
+    const container = await new QdrantContainer("qdrant/qdrant:latest").start();
```
Anush008, line 11: This is displayed at
https://node.testcontainers.org/modules/qdrant/.

cristianrgreco: We should pin the version. If the container implementation uses `v1.13.4`, then we should use that. Otherwise these tests will break in the future when a breaking change is released.

Anush008: Done with https://github.com/testcontainers/testcontainers-node/pull/917/commits/f13cc864d62a2f497032f2e4576e99d818ff4144.

cristianrgreco:
@Anush008 Tests are failing:

```
FAIL packages/modules/qdrant/src/qdrant-container.test.ts (17.112 s)
  QdrantContainer
    ✓ should connect to the client (7699 ms)
    ✓ should work with valid API keys (1360 ms)
    ✕ should fail for invalid API keys (1369 ms)
    ✓ should work with config files - valid API key (1352 ms)
    ✕ should work with config files - invalid API key (1332 ms)

  ● QdrantContainer › should fail for invalid API keys

    expect(received).rejects.toThrow(expected)

    Expected substring: "Forbidden"
    Received message:   "Unauthorized"

      at Object.fun [as getCollections] (../../../node_modules/@qdrant/openapi-typescript-fetch/dist/cjs/fetcher.js:172:23)
      at async QdrantClient.getCollections (../../../node_modules/@qdrant/js-client-rest/dist/cjs/qdrant-client.js:835:26)
      at Object.toThrow (../../../node_modules/expect/build/index.js:218:22)
      at Object.<anonymous> (src/qdrant-container.test.ts:45:45)

  ● QdrantContainer › should work with config files - invalid API key

    expect(received).rejects.toThrow(expected)

    Expected substring: "Forbidden"
    Received message:   "Unauthorized"

      at Object.fun [as getCollections] (../../../node_modules/@qdrant/openapi-typescript-fetch/dist/cjs/fetcher.js:172:23)
      at async QdrantClient.getCollections (../../../node_modules/@qdrant/js-client-rest/dist/cjs/qdrant-client.js:835:26)
      at Object.toThrow (../../../node_modules/expect/build/index.js:218:22)
      at Object.<anonymous> (src/qdrant-container.test.ts:74:45)
```

We should probably keep the version the same as in main, `qdrant/qdrant:v1.8.1`, otherwise it'll need to be a breaking change. This will be resolved by https://github.com/testcontainers/testcontainers-node/issues/867

Anush008:
> otherwise it'll need to be a breaking change
In this case, it was actually an update in the REST status message.
Qdrant is backwards compatible.

cristianrgreco (approved):

PR #918, opened 2025-03-05
chore: cache NPM dependencies in the smoke-test phase

mdelapenya:
It caches the NPM dependencies using the official GH action for caching.

See https://github.com/actions/cache/blob/main/examples.md#node---npm for more info.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -35,6 +35,17 @@ jobs:
           node-version: ${{ matrix.node-version }}
       - name: Install dependencies
         run: npm ci --omit=optional
+      - name: Get npm cache directory
+        id: npm-cache-dir
+        shell: bash
+        run: echo "dir=$(npm config get cache)" >> ${GITHUB_OUTPUT}
+      - uses: actions/cache@v4
+        id: npm-cache # use this to check for `cache-hit` ==> if: steps.npm-cache.outputs.cache-hit != 'true'
+        with:
+          path: ${{ steps.npm-cache-dir.outputs.dir }}
+          key: ${{ runner.os }}-node-{{ matrix.node-version }}-${{ hashFiles('**/package-lock.json') }}
+          restore-keys: |
+            ${{ runner.os }}-node-{{ matrix.node-version }}-
       - name: Build workspaces
         run: npm run build -ws
       - name: Remove dev dependencies


.github/workflows/test-template.yml
```diff
@@ -130,13 +130,12 @@ jobs:
...
-        uses: actions/setup-node@v4
+      - name: Install Node ${{ inputs.node-version }} and Dependencies
+        uses: ./.github/actions/npm-setup
         with:
+          runner: ${{ inputs.runner }}
           node-version: ${{ inputs.node-version }}
-
-      - name: Install dependencies
-        run: npm ci --omit=optional
+          workspace: "${{ inputs.workspace }}"
```
mdelapenya, line 138: @cristianrgreco I have one question here, probably caused by my lack of knowledge in the build system used here.

IIUC each module will be built in a worker, but should they reuse any cached node_modules dir from a previous stage? I'm adding a cache key that uses the workspace dir, and it could not be that performant if each module needs to build their own node_modules for first time. Of course, if the package-json.lock file does not change, it will be reused across builds, but in any case, I'd like to to tune this in the right manner.

mdelapenya: See this error in chromadb: https://github.com/testcontainers/testcontainers-node/actions/runs/13686730534/job/38272880849

cristianrgreco: @mdelapenya The project is structured using [NPM workspaces](https://docs.npmjs.com/cli/v8/using-npm/workspaces). The `testcontainers` package is built standalone, and each module has a dependency on the `testcontainers` package **plus** its own dependencies. So chromaDB for example depends on `testcontainers` + `chromadb`, for example. I'm not aware of a way to merge dependencies between packages. I think the best we'll get is a cache for `testcontainers` and a cache for each module.

IDK if it's worth also looking at switching from NPM to PNPM, I hear it's faster at resolving/installing packages.

mdelapenya: I think I'm adding a cache for testcontainers and a cache per module 🤔 If you check out https://github.com/testcontainers/testcontainers-node/actions/runs/13686730534/job/38272872166, for chroma with NodeJS 18.x, the cache key is `refactor-pipeline-ubuntu-22.04-node-18.x-packages/modules/chromadb-cf7a257f906333ec16e5ea58a4b07b39c48331b08ee6d1250a6b4f61f56ffff9`, which is the same used by the failing pipeline: `refactor-pipeline-ubuntu-22.04-node-18.x-packages/modules/chromadb-cf7a257f906333ec16e5ea58a4b07b39c48331b08ee6d1250a6b4f61f56ffff9`.

I wonder if we should cache more paths, and not just node_modules. Maybe the `npm cache dir` too?

cristianrgreco: Yep the cache keys are correct. 

> I wonder if we should cache more paths, and not just node_modules

Yes I think that's the problem. There's `node_modules` in:
- `/node_modules`
- `/packages/testcontainers/node_modules`
- `packages/modules/*/node_modules`

.github/workflows/lint.yml
```diff
@@ -14,7 +14,11 @@ jobs:
...
       with:
         node-version: 22.x
-    - name: Install dependencies
-      run: npm ci --omit=optional
+    - name: Install Node and Dependencies
+      uses: ./.github/actions/npm-setup
+      with:
+        runner: ubuntu-22.04
+        node-version: 22.x
+        workspace: "packages/testcontainers"
```
cristianrgreco, line 22: Does this mean that the linter won't pick up issues from modules?

.github/scripts/changed-modules.sh
```diff
@@ -0,0 +1,151 @@
...
+# as it is not a module.
+for packageJSONFile in $(find "${ROOT_DIR}" -name "package.json" -not -path "*/node_modules/*"); do
+    name=$(basename "$(dirname "${packageJSONFile}")")
+    if [[ "${name}" != "testcontainers-node" ]]; then
+        modules+=("\"${name}\"")
+    fi
+done
+# sort modules array
+IFS=$'\n' modules=($(sort <<<"${modules[*]}"))
+unset IFS
```
cristianrgreco, line 79: Duplicate

mdelapenya: Good catch, removed in [39b904d](https://github.com/testcontainers/testcontainers-node/pull/918/commits/39b904dc66b051a47aefd520091f78f40c82de87)

PR #922, opened 2025-03-11
Fix lambda container is dropped by ryuk (#848)

### LINKED ISSUE #848: Lambda container not being removed automatically
Hi there!

I was doing some testing with Testcontainers and LocalStack. I created a container as the docs say, but it turned out that when I stopped it, the lambda container was left started. I found [a similar issue](https://github.com/localstack/localstack/issues/8616) in the Java implementation so I tried porting it to my Node solution, ending up with something like:
```
before(async () => {
    const resourceReaper = await getContainerRuntimeClient().then(crc => getReaper(crc))
    container = await new LocalstackContainer('localstack/localstack:latest')
        .withEnvironment(
            {
                SERVICES: 's3,lambda',
                EAGER_SERVICE_LOADING: '1'
            }
        )
        // below is needed so the lambda service could create a docker image for itself
        .withBindMounts([{
            source: "/var/run/docker.sock",
            target: "/var/run/docker.sock"
        }])
        .withEnvironment(
            {
                /**
                 * Localstack lambda spins a new Docker container up but the container is not removed by
                 * Testcontainers resource reaper unless we pass a label indicating the session id. 
                 */
                LAMBDA_DOCKER_FLAGS: `-l org.testcontainers.session-id=${resourceReaper.sessionId}`
            }
        )
        .start();

[...]
})
```

Can you make this out-of-the-box solution, please? I found those _orphan_ containers only because I started having issues with resources and it seems the LocalStack has no way to integrate with the resource reaper. If not, then maybe you could mention this in the docs? WDYT?

julienboulay:
Fixes #848 

Configure the LocalStack container to add sessionId when starting lambda container.
This allows the lambda container to be identified by the reaper and cleaned up on exit.

See [Localstack documentation](https://docs.localstack.cloud/references/configuration/) for more details about LAMBDA_DOCKER_FLAGS

### DIFF
--- a/packages/modules/localstack/src/localstack-container.test.ts
+++ b/packages/modules/localstack/src/localstack-container.test.ts
@@ -1,6 +1,7 @@
-import { LOCALSTACK_PORT, LocalstackContainer } from "./localstack-container";
-import { HeadBucketCommand, S3Client, CreateBucketCommand } from "@aws-sdk/client-s3";
+import { CreateBucketCommand, HeadBucketCommand, S3Client } from "@aws-sdk/client-s3";
 import { GenericContainer, log, Network, StartedTestContainer } from "testcontainers";
+import { LABEL_TESTCONTAINERS_SESSION_ID } from "testcontainers/src/utils/labels";
+import { LocalstackContainer, LOCALSTACK_PORT } from "./localstack-container";
 
 const runAwsCliAgainstDockerNetworkContainer = async (
   command: string,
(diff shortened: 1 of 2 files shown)


cristianrgreco (approved):

cristianrgreco:
@julienboulay np, thank you for the PR!

I've merged another PR which was a minor change which hasn't yet been released. Because of that the next version will be 10.19.0. I'll trigger a release once I've merged https://github.com/testcontainers/testcontainers-node/pull/921

Issue #884, opened 2024-12-19
Commit images / duplicate containers / etc


zacaj:
I'm using testcontainers to spin up a postgres instance for testing against, but after every test I need to shut down the container, start up a new one, and then run all my migrations and seed scripts again.  It seems like there should be a way to image the instance once I've run those so I can re-use it in future tests, but I don't see anything in the docs that'd allow this...  or am I missing something?  the ability to take an archive seemed like it'd work, but there doesn't seem to be any way to init a new instance using that archive

cristianrgreco:
It looks like Docker now supports committing a container: https://docs.docker.com/reference/api/engine/sdk/examples/. This is something we should implement

papidb:
Hi @cristianrgreco, Good day, is this something the team is looking to implement? If they are can I give it a try?

cristianrgreco:
> Hi [@cristianrgreco](https://github.com/cristianrgreco), Good day, is this something the team is looking to implement? If they are can I give it a try?

Yes please, PR welcome

PR #923, opened 2025-03-12
feat: add support for container.commit() [#884]

smcroskey:
Adds support for committing a container to an image ([docker container commit](https://docs.docker.com/reference/api/engine/sdk/examples/#commit-a-container)).  

Unfortunately @types/dockerode don't expose the [commit options](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/dockerode/index.d.ts#L54) so I've included the type in types.ts.

### DIFF
--- a/docs/features/containers.md
+++ b/docs/features/containers.md
@@ -357,6 +357,16 @@ const container = await new GenericContainer("alpine").start();
 await container.restart();
 ```
 
+## Committing a container to an image
+
+```javascript
+const container = await new GenericContainer("alpine").start();
+// Do something with the container
+await container.exec(["sh", "-c", `echo 'hello world' > /hello-world.txt`]);
+// Commit the container to an image
+const newImage = await container.commit({ repository: "my-repo", tag: "my-tag" });
+```
+
 ## Reusing a container
 
 Enabling container re-use means that Testcontainers will not start a new container if a Testcontainers managed container with the same configuration is already running.
--- a/packages/testcontainers/src/container-runtime/clients/container/docker-container-client.ts
+++ b/packages/testcontainers/src/container-runtime/clients/container/docker-container-client.ts
@@ -11,7 +11,7 @@ import { IncomingMessage } from "http";
 import { PassThrough, Readable } from "stream";
 import { execLog, log, streamToString } from "../../../common";
 import { ContainerClient } from "./container-client";
-import { ContainerStatus, ExecOptions, ExecResult } from "./types";
+import { CommitOptions, ContainerStatus, ExecOptions, ExecResult } from "./types";
 
 export class DockerContainerClient implements ContainerClient {
   constructor(public readonly dockerode: Dockerode) {}
@@ -260,6 +260,17 @@ export class DockerContainerClient implements ContainerClient {
       log.debug(`Restarted container`, { containerId: container.id });
     } catch (err) {
       log.error(`Failed to restart container: ${err}`, { containerId: container.id });
+      throw err;
+    }
+  }
+
+  async commit(container: Container, opts?: CommitOptions): Promise<void> {
+    try {
+      log.debug(`Committing container...`, { containerId: container.id });
+      await container.commit(opts);
+      log.debug(`Committed container image`, { containerId: container.id });
+    } catch (err) {
+      log.error(`Failed to commit container: ${err}`, { containerId: container.id });
       throw err;
     }
   }
--- a/packages/testcontainers/src/generic-container/generic-container-commit.test.ts
+++ b/packages/testcontainers/src/generic-container/generic-container-commit.test.ts
@@ -1,0 +1,52 @@
+import { RandomUuid } from "../common";
+import { deleteImageByName, getImageInfo, getImageLabelsByName } from "../utils/test-helper";
+import { GenericContainer } from "./generic-container";
+
+describe("GenericContainer commit", { timeout: 180_000 }, () => {
+  const imageName = "cristianrgreco/testcontainer";
+  const imageVersion = "1.1.14";
+
+  it("should commit container changes to a new image", async () => {
+    const testContent = "test content";
+    const newImageTag = `test-commit-${new RandomUuid().nextUuid()}`;
+    const testAuthor = "test-author";
+    const testComment = "test-comment";
+
+    // Start original container and make a change
+    const container = await new GenericContainer(`${imageName}:${imageVersion}`)
+      .withName(`container-${new RandomUuid().nextUuid()}`)
+      .withExposedPorts(8080)
+      .start();
+
+    // Make a change to the container
+    await container.exec(["sh", "-c", `echo '${testContent}' > /test-file.txt`]);
+
+    // Commit the changes to a new image
+    await container.commit({
+      repo: imageName,
+      tag: newImageTag,
+      author: testAuthor,
+      comment: testComment,
+    });
+
+    // Verify image author and comment are set
+    const imageInfo = await getImageInfo(`${imageName}:${newImageTag}`);
+    expect(imageInfo.Author).toBe(testAuthor);
+    expect(imageInfo.Comment).toBe(testComment);
+
+    // Start a new container from the committed image
+    const newContainer = await new GenericContainer(`${imageName}:${newImageTag}`)
+      .withName(`container-${new RandomUuid().nextUuid()}`)
+      .withExposedPorts(8080)
+      .start();
+
+    // Verify the changes exist in the new container
+    const result = await newContainer.exec(["cat", "/test-file.txt"]);
+    expect(result.output.trim()).toBe(testContent);
+
+    // Cleanup
+    await container.stop();
+    await newContainer.stop();
+    await deleteImageByName(`${imageName}:${newImageTag}`);
+  });
+});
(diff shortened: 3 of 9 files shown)


packages/testcontainers/src/test-container.ts
```diff
@@ -60,9 +61,18 @@ export interface StopOptions {
   removeVolumes: boolean;
 }
 
+export interface CommitOptions {
+  repo?: string;
+  tag?: string;
```
cristianrgreco, line 66: What's the use case for these to be undefined?

docs/features/containers.md
```diff
@@ -357,6 +357,16 @@ const container = await new GenericContainer("alpine").start();
...
 ```
 
+## Committing a container to an image
+
+```javascript
+const container = await new GenericContainer("alpine").start();
+// Do something with the container
+await container.exec(["sh", "-c", `echo 'hello world' > /hello-world.txt`]);
+// Commit the container to an image
+const newImage = await container.commit({ repository: "my-repo", tag: "my-tag" });
```
cristianrgreco, line 367: What is `newImage` here? `container.commit` returns `Promise<void>`

packages/testcontainers/src/generic-container/generic-container-commit.test.ts
```diff
@@ -0,0 +1,52 @@
...
+
+  it("should commit container changes to a new image", async () => {
+    const testContent = "test content";
+    const newImageTag = `test-commit-${new RandomUuid().nextUuid()}`;
+    const testAuthor = "test-author";
+    const testComment = "test-comment";
+
+    // Start original container and make a change
+    const container = await new GenericContainer(`${imageName}:${imageVersion}`)
+      .withName(`container-${new RandomUuid().nextUuid()}`)
```
cristianrgreco, line 17: No need to give the container a name, same on line 39

packages/testcontainers/src/container-runtime/clients/container/docker-container-client.ts
```diff
@@ -264,6 +264,17 @@ export class DockerContainerClient implements ContainerClient {
     }
   }
 
+  async commit(container: Container, opts?: CommitOptions): Promise<void> {
+    try {
+      log.debug(`Committing container...`, { containerId: container.id });
+      await container.commit(opts);
```
cristianrgreco, line 270: We should provide labels as we do here, perhaps let the user specify whether to `deleteOnExit`, so that the image is cleaned up by the resource reaper when the tests complete: https://github.com/testcontainers/testcontainers-node/blob/5817d695fe370168ec25a52f9ee1b59b1a3775e1/packages/testcontainers/src/generic-container/generic-container-builder.ts#L66-L82

packages/testcontainers/src/generic-container/generic-container-commit.test.ts
```diff
@@ -0,0 +1,52 @@
...
+      .start();
+
+    // Verify the changes exist in the new container
+    const result = await newContainer.exec(["cat", "/test-file.txt"]);
+    expect(result.output.trim()).toBe(testContent);
+
+    // Cleanup
+    await container.stop();
+    await newContainer.stop();
+    await deleteImageByName(`${imageName}:${newImageTag}`);
```
cristianrgreco, line 50: If we implement the labels such that the image is cleaned up by the resource reaper, then this isn't needed (waiting for this just slows the test down)

cristianrgreco:
@smcroskey thanks for raising the PR, this feature will be very useful. Left a few comments, LMK if you have any Qs.

packages/testcontainers/src/container-runtime/clients/container/docker-container-client.ts
smcroskey: thank you for calling this out!  In order to get this to work, I had to also expose the `changes` option because by default it was actually saving the label with the sessionId to the committed image.  Now, we'll apply the change command to unset the label if `deleteOnExit` is true (added a test case for this as well).  Lmk what you think / if you have concerns with the approach.

docs/features/containers.md
smcroskey: Thanks, good catch.  It was supposed to return the image ID (updated)

packages/testcontainers/src/test-container.ts
smcroskey: I don't think there are any, so I've updated so they're required

packages/testcontainers/src/test-container.ts
```diff
@@ -63,6 +65,7 @@ export interface StopOptions {
 export interface StartedTestContainer {
   stop(options?: Partial<StopOptions>): Promise<StoppedTestContainer>;
   restart(options?: Partial<RestartOptions>): Promise<void>;
+  commit(options: ContainerCommitOptions): Promise<string>;
```
cristianrgreco, line 68: Could we swap these type names around so that `CommitOptions` is used here in the public API, and `ContainerCommitOptions` is used internally? I suggest it mostly for consistency, as all options are implicitly container options

packages/testcontainers/src/generic-container/started-generic-container.ts
```diff
@@ -37,6 +47,40 @@ export class StartedGenericContainer implements StartedTestContainer {
...
+    } else if (!deleteOnExit && this.getLabels()[LABEL_TESTCONTAINERS_SESSION_ID]) {
+      // By default, commit will save the existing labels (including the session ID) to the new image.  If
+      // deleteOnExit is false, we need to remove the session ID label.
+      changes.push(`LABEL ${LABEL_TESTCONTAINERS_SESSION_ID}=`);
+    }
+    return changes.filter(Boolean).join("\n");
+  }
+
+  public async commit(options: ContainerCommitOptions): Promise<string> {
+    log.info(`Committing container image...`, { containerId: this.container.id });
```
cristianrgreco, line 75: Can be removed as it's logged by the container client

packages/testcontainers/src/generic-container/started-generic-container.ts
```diff
@@ -37,6 +47,40 @@ export class StartedGenericContainer implements StartedTestContainer {
...
+      if (!sessionId) {
+        sessionId = await getReaper(client).then((reaper) => reaper.sessionId);
+      }
+      changes.push(`LABEL ${LABEL_TESTCONTAINERS_SESSION_ID}=${sessionId}`);
+    } else if (!deleteOnExit && this.getLabels()[LABEL_TESTCONTAINERS_SESSION_ID]) {
+      // By default, commit will save the existing labels (including the session ID) to the new image.  If
+      // deleteOnExit is false, we need to remove the session ID label.
+      changes.push(`LABEL ${LABEL_TESTCONTAINERS_SESSION_ID}=`);
+    }
+    return changes.filter(Boolean).join("\n");
```
cristianrgreco, line 71: Is there a need to filter here?

packages/testcontainers/src/generic-container/generic-container-commit.test.ts
```diff
@@ -0,0 +1,87 @@
...
+
+    // Commit with deleteOnExit true (default)
+    const imageId = await container.commit({
+      repo: imageName,
+      tag: newImageTag,
+    });
+
+    // Verify session ID label is present
+    const imageInfo = await getImageInfo(imageId);
+    expect(imageInfo.Config.Labels[LABEL_TESTCONTAINERS_SESSION_ID]).toBeDefined();
```
cristianrgreco, line 60: Could we check the value? Should be able to get it from the reaper.

packages/testcontainers/src/container-runtime/clients/image/docker-image-client.ts
```diff
@@ -65,6 +65,18 @@ export class DockerImageClient implements ImageClient {
     return (aPath: string) => !filter(aPath);
   }
 
+  async inspect(imageName: ImageName): Promise<ImageInspectInfo> {
+    try {
+      log.debug(`Inspecting image "${imageName.string}"...`);
+      const imageInfo = await this.dockerode.getImage(imageName.string).inspect();
+      log.debug(`Retrieved image info "${imageName.string}"`);
```
cristianrgreco, line 72:
```suggestion
      log.debug(`Inspected image: "${imageName.string}"`);
```

docs/features/containers.md
```diff
@@ -357,6 +357,16 @@ const container = await new GenericContainer("alpine").start();
...
 ```
 
+## Committing a container to an image
+
+```javascript
+const container = await new GenericContainer("alpine").start();
+// Do something with the container
+await container.exec(["sh", "-c", `echo 'hello world' > /hello-world.txt`]);
+// Commit the container to an image
+const newImageId = await container.commit({ repo: "my-repo", tag: "my-tag", deleteOnExit: false });
```
cristianrgreco, line 367: Could we omit the default and add a line explaining that by default the image will be cleaned up, and give a second example overriding this behaviour?

Could we also give an example how this `newImageId` is used? I think it's not clear what the value is. If we add like `const containerFromCommit = await new GenericContainer(newImageId).start();` I think that'll be nice and clear

cristianrgreco:
@smcroskey thanks for making the changes! Just a few more nitpicks

packages/testcontainers/src/generic-container/started-generic-container.ts
smcroskey: ah nope...leftover from a previous implementation.  Good catch.

cristianrgreco (approved):

cristianrgreco:
@smcroskey thanks for the contribution!

smcroskey:
Thank you for the prompt review(s) @cristianrgreco!

Issue #915, opened 2025-03-04
Add `--database-mode=datastore-mode` option to firestore emulator


jacobg:
Please add `--database-mode=datastore-mode` option to google gcp firestore emulator

cristianrgreco:
After #926 you'll be able to run:

```ts
new FirestoreEmulatorContainer()
  .withFlag("database-mode", `datastore-mode`);
  .start();
```

PR #926, opened 2025-03-12
feat(gloud) add ability to set cmd flags

digital88:
Closes https://github.com/testcontainers/testcontainers-node/issues/915

Added EmulatorFlagsManager class which encapsulates logic of adding arbitrary flags to command used to start gcloud containers:
```
gcloud beta emulators firestore start ...<flags>
```
Instead of hard-coding flags inside container cmd string, users can now add any flags they wish. Existing flags were preserved, existing 3rd party code should be working without changes.

Added tests for EmulatorFlagsManager.


packages/modules/gcloud/src/emulator-flags-manager.ts
```diff
@@ -0,0 +1,42 @@
...
+  private flags: { name: string; value: string }[] = [];
+
+  /**
+   * Adds flag as argument to emulator start command.
+   * Adding same flag name twice replaces existing flag value.
+   * @param name flag name. Must be set to non-empty string. May optionally contain -- prefix.
+   * @param value flag value. May be empty string.
+   * @returns this instance for chaining.
+   */
+  public withFlag(name: string, value: string): this {
```
cristianrgreco, line 11: If `this.flags` were an object of type `{ [name: string]: string }`, I think this could be simplified, and the trimming and adding `—` could also be removed.

packages/modules/gcloud/src/datastore-emulator-container.test.ts
```diff
@@ -25,6 +26,35 @@ describe("DatastoreEmulatorContainer", { timeout: 240_000 }, () => {
 
   // }
```
cristianrgreco, line 28: We shouldn’t be testing the flags manager in the emulator tests. The `flagsManager` is private

packages/modules/gcloud/src/datastore-emulator-container.ts
```diff
@@ -1,18 +1,47 @@
...
   constructor(image = DEFAULT_IMAGE) {
     super(image);
+    this._flagsManager = new EmulatorFlagsManager();
     this.withExposedPorts(EMULATOR_PORT)
-      .withCommand(["/bin/sh", "-c", CMD])
+      .withFlag("host-port", `0.0.0.0:${EMULATOR_PORT}`)
+      .withFlag("database-mode", `datastore-mode`)
+      // explicitly call withCommand() fn here
+      // it will be called implicitly inside prev withFlag() call
+      .withCommand(["/bin/sh", "-c", this.getCmd()])
```
cristianrgreco, line 18: This line of code conflicts with the comment above it. Indeed it does not seem necessary. Let’s remove setting the command from here and from the `withFlag` method and set it once in the start method.

packages/modules/gcloud/src/firestore-emulator-container.test.ts
```diff
@@ -1,8 +1,10 @@
 import * as admin from "firebase-admin";
+import { Wait } from "testcontainers";
 import { FirestoreEmulatorContainer, StartedFirestoreEmulatorContainer } from "./firestore-emulator-container";
 
 describe("FirestoreEmulatorContainer", { timeout: 240_000 }, () => {
-  afterEach(async () => {
+  afterEach(async (ctx) => {
+    if (!["should work using default version", "should work using version 468.0.0"].includes(ctx.task.name)) return;
```
cristianrgreco, line 7: Let’s add `await admin.app().delete();` to the 2 tests that need it instead of doing it like this. Tying this behaviour to the test name isn’t great for maintenance.

packages/modules/gcloud/src/emulator-flags-manager.ts
digital88: done

packages/modules/gcloud/src/datastore-emulator-container.test.ts
digital88: removed redundant tests

packages/modules/gcloud/src/datastore-emulator-container.ts
digital88: moved withCommand to start(), withFlag() no longer calls withCommand()

packages/modules/gcloud/src/firestore-emulator-container.test.ts
digital88: reverted, as this no longer needed after removing redundant tests

PR #934, opened 2025-03-16
Add cosmos db emulator module

mrtnsch:
This adds support for the Azure Cosmos DB emulator. Unlike the existing module for Java, .NET, and Python, this implementation uses the new Linux emulator. The old emulator is very slow and cumbersome to work with, and using the new linux-based one seems like a decent tradeoff to make.

The container requires the mapped port to be passed as an env var. I have found a solution to achieve this, but it seems quite hacky to me. I'm very open to other ideas to solve this problem.

The port has to be passed, because CosmosDB provides redirects to the individual write/read endpoints. Those redirects then target the port passed in the env var, defaulting to 8081. The Azure Cosmos SDK client follows these redirects by default, leading to an ECONREFUSED if the port in the env is different than the mapped port.

This could be avoided by disabling endpoint discovery in the client like this:
```typescript
const cosmosClient = new CosmosClient({
      endpoint: container.getEndpoint(),
      key: container.getKey(),
      connectionPolicy: {
        enableEndpointDiscovery: false,
      },
    });
```

However, this would force users of this testcontainer to specify different configurations depending on whether their client is running in prod or dev/test.

### DIFF
--- a/packages/modules/azurecosmosdb/src/azure-cosmosdb-emulator-container.ts
+++ b/packages/modules/azurecosmosdb/src/azure-cosmosdb-emulator-container.ts
@@ -1,0 +1,101 @@
+import * as net from "node:net";
+import { AbstractStartedContainer, GenericContainer, StartedTestContainer, Wait } from "testcontainers";
+
+type Protocol = "http" | "https";
+const DEFAULT_KEY = "C2y6yDjf5/R+ob0N8A7Cgv30VRDJIWEHLM+4QDU5DE2nQ9nDuVTqobD4b8mGGyPMbIZnqyMsEcaGQy67XIw/Jw=="; // default key for Cosmos DB Emulator
+const DEFAULT_PROTOCOL = "http";
+const DEFAULT_TELEMETRY_ENABLED = false;
+const DEFAULT_EXPLORER_ENABLED = false;
+
+const COSMOS_READY_LOG_MESSAGE = "Now listening on: ";
+
+export class AzureCosmosDbEmulatorContainer extends GenericContainer {
+  private key = DEFAULT_KEY;
+  private protocol: Protocol = DEFAULT_PROTOCOL;
+  private telemetryEnabled = DEFAULT_TELEMETRY_ENABLED;
+  private explorerEnabled = DEFAULT_EXPLORER_ENABLED;
+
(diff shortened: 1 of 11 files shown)


packages/modules/azurecosmosdb/src/azure-cosmosdb-emulator-container.ts
```diff
@@ -0,0 +1,101 @@
...
+
+const COSMOS_READY_LOG_MESSAGE = "Now listening on: ";
+
+export class AzureCosmosDbEmulatorContainer extends GenericContainer {
+  private key = DEFAULT_KEY;
+  private protocol: Protocol = DEFAULT_PROTOCOL;
+  private telemetryEnabled = DEFAULT_TELEMETRY_ENABLED;
+  private explorerEnabled = DEFAULT_EXPLORER_ENABLED;
+
+  constructor(image = "mcr.microsoft.com/cosmosdb/linux/azure-cosmos-emulator:vnext-preview") {
```
cristianrgreco, line 18: Is it possible to pin a version which won't change?

packages/modules/azurecosmosdb/src/azure-cosmosdb-emulator-container.ts
```diff
@@ -0,0 +1,101 @@
...
+    return new StartedAzureCosmosDbEmulatorContainer(await super.start(), this.key, port, this.protocol);
+  }
+
+  /**
+   * The mapped port has to be passed to CosmosDB as an environment variable.
+   * Since the mapped port would only be known after container creation, we need to find an available port ourselves
+   * before starting the container.
+   * @private
+   */
+  private async getFreePort(): Promise<number> {
```
cristianrgreco, line 55: Why does the container need to know the host port?

cristianrgreco, line 55: It is likely fine to use this implementation for getting a port: https://github.com/testcontainers/testcontainers-node/blob/824cd8dbac0615ae00a83d82e8625a108b885ecb/packages/testcontainers/src/utils/port-generator.ts#L17

mrtnsch: Sorry I missed this - PR is updated accordingly.

Issue #850, opened 2024-10-25
Allow configuration on container if it should be automatically cleaned up


cbrevik:
In reference to https://github.com/testcontainers/testcontainers-node/pull/849#issuecomment-2437020917 I'd like to be able to override the default behaviour where a container is cleaned up when it is stopped. 

What I suggest is adding a new function on `GenericContainer`:

- `withRemoveWhenStopped(removeWhenStopped: boolean)`

So the default behaviour is still that the container is removed when you call `container.stop()`. But to override you can explicitly configure it with `.withRemoveWhenStopped(false)` when creating the container. It should probably still be possible to override the configured value by options `container.stop({ remove: true })`.

It also made sense to me that you'd to be able to configure `withAutoRemove(autoRemove: boolean)`, so if the container is stopped in other ways than via the `testcontainers-node` API it is also cleaned up. 
 
I've taken a stab at adding both `withRemoveWhenStopped` and `withAutoRemove` here (including tests): https://github.com/testcontainers/testcontainers-node/compare/main...cbrevik:testcontainers-node:support-configured-auto-remove

cristianrgreco:
@cbrevik Could you add some info why the logic for adding `withRemoveWhenStopped` and `withAutoRemove` cannot just be put in the `container.stop({ ... })`. IMO having the same option in 2 places, with one overriding the other is confusing

cbrevik:
The main use case for me would be if I enable `withReuse()` on a container, I normally don't want it to be removed when stopped. So it made sense to me that I could configure that "in the same place" as I configure re-use.

In other words something like:

```typescript
const container = await new PostgreSqlContainer()
    .withReuse()
    .withRemove(false)
    .start()
```

Or something like that. I normally "pass" the container somewhere else, and then I have to remember to call `container.stop({ remove: false })` in the other module. 

It's not a huge deal though, just made sense that it would be configureable somehow to me.

cristianrgreco:
Completed in https://github.com/testcontainers/testcontainers-node/pull/939

PR #905, opened 2025-01-27
feat: implement `.withAutoRemove` #850

botflux:
Hey, here is a PR to implement the request of #850.
I've implemented the following behavior:
- `.withAutoRemove(false)` disables container's auto removal
- `.stop({ remove: true })` overrides the `.withAutoRemove(false)`, so the container is removed
- By default, containers are automatically removed

I've also added sections about the new behaviors close to existing sections about `.stop({ remove: false })`.

On the other hand, I've noted that docker prefixes container name with a slash, meaning that the test I wrote would fail:
```typescript
  it("should stop but not remove the container", async () => {
    const container = await new GenericContainer("cristianrgreco/testcontainer:1.1.14")
      .withName(`container-${new RandomUuid().nextUuid()}`)
      .withAutoRemove(false)
      .start();

    await container.stop();

    expect(await getRunningContainerNames()).not.toContain(container.getName()); // never fails because container is prefixed
    expect(await getStoppedContainerNames()).toContain(container.getName()); // fails because `container.getName()`' result is prefixed with '/'.
  });
```

Here is the output of jest:
```
Summary of all failing tests
 FAIL  packages/testcontainers/src/generic-container/generic-container.test.ts (90.581 s)
  ● GenericContainer › should stop but not remove the container

    expect(received).toContain(expected) // indexOf

    Expected value: "/container-4e957eea11e6"
    Received array: ["container-4e957eea11e6", "testcontainers-22930a5025fc-service_2-1", "testcontainers-686a91976fd7-service_1-1", "testcontainers-686a91976fd7-service_2-1", "container-797b0e580f81", "testcontainers-a1999481affa-service_2-1", "testcontainers-39d1bd573914-service_1-1", "testcontainers-39d1bd573914-service_2-1", "goofy_aryabhata", "silly_hofstadter", …]

      530 |
      531 |     expect(await getRunningContainerNames()).not.toContain(container.getName());
    > 532 |     expect(await getStoppedContainerNames()).toContain(container.getName());
          |                                              ^
      533 |   });
      534 |
      535 |   it("should stop and override .withAutoRemove", async () => {

      at Object.<anonymous> (src/generic-container/generic-container.test.ts:532:46)


Test Suites: 1 failed, 44 passed, 45 total
Tests:       1 failed, 3 skipped, 323 passed, 327 total
Snapshots:   0 total
Time:        114.873 s
Ran all test suites.
```

To fix the test I've removed the leading '/'
```typescript
[...]

PR #936, opened 2025-03-17
add withRemove() and tests

digital88:
Related issue https://github.com/testcontainers/testcontainers-node/issues/850

Add withRemove() method to GenericContainer and tests. This allows to control autoremoval behaviour of a container.

Containers are removed by default if `remove` option is not set when container is being stopped.
```
await container.stop(); // will remove stopped container.
```

`withRemove(true)` produces the same effect as explicitly setting remove flag when stopping container.
```
container.withRemove(true).start();
...
await container.stop();

// is equivalent of doing:

await container.stop({ remove: true });

// is equivalent of doing:

await container.stop();
```

`withRemove(false)` disables removing container by default.
```
container.withRemove(false).start();
...
await container.stop(); // will NOT remove stopped container.
```

`container.stop()` options have higher priority than `withRemove()`
```
container.withRemove(false).start();
...
await container.stop({ remove: true }); // will override withRemove(false) and remove stopped container.
```

Keep in mind that Ryuk will remove stopped container after tests finished regardless of value passed in `withRemove` method, unless Ryuk is disabled. Setting `withRemove(false)` and disabling Ryuk is not recommeded.

### DIFF
--- a/packages/testcontainers/src/docker-compose-environment/docker-compose-environment.ts
+++ b/packages/testcontainers/src/docker-compose-environment/docker-compose-environment.ts
@@ -169,6 +169,7 @@ export class DockerComposeEnvironment {
             boundPorts,
             containerName,
             waitStrategy
+            // not sure how to control 'remove' option for the whole compose stack, will use default value here (true)
           );
         })
       )
--- a/packages/testcontainers/src/generic-container/generic-container-reuse.test.ts
+++ b/packages/testcontainers/src/generic-container/generic-container-reuse.test.ts
@@ -153,6 +153,27 @@ describe("GenericContainer reuse", { timeout: 180_000 }, () => {
     await container2.stop();
   });
 
+  it("should reuse container when an existing reusable container created using withRemove(false) has stopped", async () => {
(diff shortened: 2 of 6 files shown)


cristianrgreco:
Hi @digital88, thanks for raising the PR, but it is very similar to https://github.com/testcontainers/testcontainers-node/pull/905. The other PR is a bit further progressed, e.g it updates the docs, and the testcontainer interface. It looks like both PRs are now failing for the same reason, some difference with the container name.

Instead of asserting the container name does or does not exist, is it not possible to assert the container IDs? That should resolve any inconsistency with whether the name is prefixed with `\` or not.

Would it be possible for you to collaborate on PR #905 instead of creating a new one?

digital88:
> Would it be possible for you to collaborate on PR #905 instead of creating a new one?

Yes, I'll look into it.


> Instead of asserting the container name does or does not exist, is it not possible to assert the container IDs?

Yes, I was thinking about using container id too. Will try this idea in PR #905

digital88:
See https://github.com/testcontainers/testcontainers-node/pull/939

Issue #830, opened 2024-09-02
Cannot enable verbose mode for ryuk


FelixZY:
Ryuk supports [verbose output via the `RYUK_VERBOSE` envvar](https://github.com/testcontainers/moby-ryuk/blob/e8ed9c9172b99aacc732cf0d09414e97e1294d64/main.go#L32). It should be possible to request this verbose output somehow.

I would suggest automatically enabling verbose output when enabling verbose output via `DEBUG`, alternatively adding `TESTCONTAINERS_RYUK_VERBOSE` to the [list of supported configuration variables](https://node.testcontainers.org/configuration/)

**Environment Information**
- Testcontainers version: 10.13.0

cristianrgreco:
PR welcome

PR #941, opened 2025-03-19
Add `TESTCONTAINERS_RYUK_VERBOSE` setting

digital88:
As requested in #830

Tests are somewhat possible, but please see comment in .test.ts file.

### DIFF
--- a/packages/testcontainers/src/reaper/reaper.test.ts
+++ b/packages/testcontainers/src/reaper/reaper.test.ts
@@ -1,8 +1,6 @@
 import { ContainerRuntimeClient, getContainerRuntimeClient } from "../container-runtime";
-import { Reaper } from "./reaper";
 
 describe("Reaper", { timeout: 120_000 }, () => {
-  let reaper: Reaper;
   let client: ContainerRuntimeClient;
 
   const getReaper = async () => await (await import("./reaper")).getReaper(client);
@@ -16,7 +14,7 @@ describe("Reaper", { timeout: 120_000 }, () => {
 
   it("should create Reaper container without RYUK_VERBOSE env var by default", async () => {
     vi.spyOn(client.container, "list").mockResolvedValue([]);
-    reaper = await getReaper();
+    const reaper = await getReaper();
 
     const reaperContainer = client.container.getById(reaper.containerId);
     const reaperContainerEnv = (await reaperContainer.inspect()).Config.Env;
@@ -28,7 +26,7 @@ describe("Reaper", { timeout: 120_000 }, () => {
     vitest.stubEnv("TESTCONTAINERS_RYUK_VERBOSE", "true");
 
     vi.spyOn(client.container, "list").mockResolvedValue([]);
-    reaper = await getReaper();
+    const reaper = await getReaper();
 
     const reaperContainer = client.container.getById(reaper.containerId);
     expect((await reaperContainer.inspect()).Config.Env).toContain("RYUK_VERBOSE=true");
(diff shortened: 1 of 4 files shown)


packages/testcontainers/src/reaper/reaper.test.ts
```diff
@@ -0,0 +1,30 @@
+import { getContainerRuntimeClient } from "../container-runtime";
+import { getReaper } from "./reaper";
+
+// ryuk container is created as global variable
+// so it's impossible to test against it in isolation.
+// both of this tests work if run manually one after another
+// and ensuring that ryuk container does not exist before either test.
```
cristianrgreco, line 7: You may be able to do something like this to reset the module for each test:

https://github.com/testcontainers/testcontainers-node/blob/0224ba70c00dd2619baa8708a945a051326219d8/packages/testcontainers/src/container-runtime/strategies/utils/config.test.ts#L8-L16

packages/testcontainers/src/reaper/reaper.ts
```diff
@@ -16,6 +16,8 @@ export interface Reaper {
   addSession(sessionId: string): void;
 
   addComposeProject(projectName: string): void;
+
+  get reaperContainerId(): string;
```
cristianrgreco, line 20: Let's rename to `id` or `containerId`, `reaper` is implied

digital88: done

packages/testcontainers/src/reaper/reaper.test.ts
digital88: I had to do this for `getReaper` fn AND also have to pass custom FindReaperContainerFn as `getReaper` argument.

packages/testcontainers/src/reaper/reaper.test.ts
```diff
@@ -0,0 +1,45 @@
...
+    getReaperFn = (await import("./reaper")).getReaper;
+  });
+
+  afterEach(() => {
+    vi.resetModules();
+  });
+
+  it("should create Reaper container without RYUK_VERBOSE env var by default", async () => {
+    const client = await getContainerRuntimeClient();
+    const reaper = await getReaperFn(client, async () => undefined);
```
cristianrgreco, line 25: What if we mock `getContainerRuntimeClient` instead, have `client.container.list()` return an empty list for our tests. That way we don't change the implementation for the tests, don't need to expose the inner type, etc.

packages/testcontainers/src/reaper/reaper.ts
```diff
@@ -135,9 +146,14 @@ async function connectToReaperSocket(host: string, port: number, containerId: st
 class RyukReaper implements Reaper {
   constructor(
     public readonly sessionId: string,
-    private readonly socket: Socket
+    private readonly socket: Socket,
+    private readonly id: string
```
cristianrgreco, line 150: Rename to `containerId` and make type `public`? That should still satisfy the interface and let us get rid of the getter

packages/testcontainers/src/reaper/reaper.ts
```diff
@@ -16,19 +16,28 @@ export interface Reaper {
   addSession(sessionId: string): void;
 
   addComposeProject(projectName: string): void;
+
+  get containerId(): string;
```
cristianrgreco, line 20: This can just be a field like on line 14

packages/testcontainers/src/reaper/reaper.test.ts
digital88: Done using mocks, thanks for hint

packages/testcontainers/src/reaper/reaper.ts
digital88: done

packages/testcontainers/src/reaper/reaper.ts
digital88: done

digital88:
Not sure why this chromadb check failed. Can you please re-run it?

PR #953, opened 2025-03-23
Build images with Buildkit

cristianrgreco:
Pull policy tests have had to be removed as we can no longer rely on Docker events to tell us if an image is being pulled.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -8,7 +8,6 @@ on:
       - "docs/**"
       - "README.md"
   pull_request:
-    branches: [main]
     paths-ignore:
       - "mkdocs.yml"
       - "docs/**"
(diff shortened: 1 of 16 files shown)


eddumelendez:
Hi @cristianrgreco, buildkit support via http doesn't take full advantage of buildkit capabilities. gRPC must be used instead.

Just want to make this clear if this is going to be supported in tc-node, in order to avoid misunderstandings.

cristianrgreco:
Hi @eddumelendez, the underlying PR to dockerode was implemented using gRPC (https://github.com/apocas/dockerode/pull/766)

cristianrgreco:
Replaced by https://github.com/testcontainers/testcontainers-node/pull/955

Issue #683, opened 2023-12-01
Improve error logging when a container fails to start


cristianrgreco:
Related issue https://github.com/testcontainers/testcontainers-node/issues/682, you can see from the error log that the wait strategy is failing because the image is distroless. 
It'd be nice to surface this without having to enable logging.

akkie:
That would be very helpful. I have a compose file that starts 10 containers. Sometimes my pipeline fails with:

```
[test](https://github.com/actions/runs/11874097483/job/33089644145#step:11:6220)
Error: Health check failed: unhealthy

   at testcontainers/compose.ts:11

   9 |   const composeFilePath = resolve(rootDir, 'docker/env-test');
  10 |
> 11 |   const environment = await new DockerComposeEnvironment(composeFilePath, 'docker-compose.yml')
     |                       ^
  12 |     .withStartupTimeout(120000)
  13 |     .withWaitStrategy('web-1', Wait.forHealthCheck())
  14 |     .withWaitStrategy(

    at HealthCheckWaitStrategy.waitUntilReady (/home/runner/work/apps/web/node_modules/.pnpm/testcontainers@10.14.0/node_modules/testcontainers/src/wait-strategies/health-check-wait-strategy.ts:26:13)
    at waitForContainer (/home/runner/work/apps/web/node_modules/.pnpm/testcontainers@10.14.0/node_modules/testcontainers/src/wait-strategies/wait-for-container.ts:17:5)
    at /home/runner/work/apps/web/node_modules/.pnpm/testcontainers@10.14.0/node_modules/testcontainers/src/docker-compose-environment/docker-compose-environment.ts:155:13
    at DockerComposeEnvironment.up (/home/runner/work/apps/web/node_modules/.pnpm/testcontainers@10.14.0/node_modules/testcontainers/src/docker-compose-environment/docker-compose-environment.ts:133:7)
    at runCompose (/home/runner/work/apps/web/e2e/testcontainers/compose.ts:11:23)
    at globalSetup (/home/runner/work/apps/web/e2e/globalSetup.ts:19:16)
```

There is absolutely no information which container failed.

PR #954, opened 2025-03-23
Improve PortCheck logging for distroless images

digital88:
See #683 

The error `The HostPortWaitStrategy will not work on a distroless image` will be logged regardless of `log.enabled()`

Distroless containers also return error code 127 when exec non-existent command. Previous logic only tested for error code 126.


packages/testcontainers/src/wait-strategies/utils/port-check-real.test.ts
```diff
@@ -0,0 +1,49 @@
+import path from "path";
+import { getContainerRuntimeClient } from "../../container-runtime";
+import { GenericContainer } from "../../generic-container/generic-container";
+import { checkContainerIsHealthy } from "../../utils/test-helper";
+import { Wait } from "../wait";
+import { InternalPortCheck } from "./port-check";
+
+describe("PortCheck", { timeout: 120_000 }, () => {
+  describe("InternalPortCheck", () => {
```
cristianrgreco, line 9: I don't think an integration test is necessary for the `PortCheck` in addition to the existing unit tests. There are integration tests which cover this functionality here: https://github.com/testcontainers/testcontainers-node/blob/7e302a510eb24999cc2df3916412c590fcfb13ab/packages/testcontainers/src/wait-strategies/host-port-wait-strategy.test.ts#L5-L28

digital88:
Oh, I just pushed a commit to this pr, but was late by a minute ) docker-no-shell fixture is also not needed, if we removed integration test.

cristianrgreco:
@digital88 Sorry about that, I want to merge all that I can for the next release, was hoping to get it out today. I've raised #959 to remove those fixtures.

Issue #817, opened 2024-08-11
Docs and Implementation Do Not Match for Container Stop Timeout


vbaier-mitll:
When trying to figure out why my containers weren't getting killed in my test suite I stumbled across a feature request for killing containers #716 . Looking into the resolution it seems that after the timeout, badly behaved containers should be killed automatically. I used some curl commands against the docker socket and discovered that this is in fact the case:

```
Aug 11 21:14:01 ip-10-0-4-201 dockerd[685]: time="2024-08-11T21:14:01.777463682Z" level=debug msg="Calling POST /containers/39fb06f06a3bb89fc1ed11fe0c66c0f90ec7bfede2ee6cab3b31ba430f2af73e/stop" spanID=601c8a9163eab753 traceID=824c04617c6e3e815632d2bf71c98dd7
Aug 11 21:14:01 ip-10-0-4-201 dockerd[685]: time="2024-08-11T21:14:01.777527926Z" level=debug msg="Sending kill signal 15 to container 39fb06f06a3bb89fc1ed11fe0c66c0f90ec7bfede2ee6cab3b31ba430f2af73e"
Aug 11 21:14:02 ip-10-0-4-201 dockerd[685]: time="2024-08-11T21:14:02.403595264Z" level=debug msg="Calling HEAD /_ping" spanID=e24d59e340674d23 traceID=5dd89daf405db44535cdece881b1af7a
Aug 11 21:14:02 ip-10-0-4-201 dockerd[685]: time="2024-08-11T21:14:02.403919114Z" level=debug msg="Calling GET /v1.45/containers/json?all=1" spanID=10ce79e914bdbe42 traceID=fca4359daec9dbcda1de0c275251be21
Aug 11 21:14:11 ip-10-0-4-201 dockerd[685]: time="2024-08-11T21:14:11.789391839Z" level=info msg="Container failed to exit within 10s of signal 15 - using the force" container=39fb06f06a3bb89fc1ed11fe0c66c0f90ec7bfede2ee6cab3b31ba430f2af73e spanID=601c8a9163eab753 traceID=824c04617c6e3e815632d2bf71c98dd7
Aug 11 21:14:11 ip-10-0-4-201 dockerd[685]: time="2024-08-11T21:14:11.789438188Z" level=debug msg="Sending kill signal 9 to container 39fb06f06a3bb89fc1ed11fe0c66c0f90ec7bfede2ee6cab3b31ba430f2af73e"
[...]

PR #958, opened 2025-03-24
Docs: clarified timeout unit in different places

digital88:
#817

Docker API itself is a little inconsistent, it uses **seconds** here in `stop` https://docs.docker.com/reference/api/engine/version/v1.48/#tag/Container/operation/ContainerStop
and also **seconds** here in compose up/down https://docs.docker.com/reference/cli/docker/compose/down/

Docker is using **nanoseconds** in HEALTHCHECK (which testcontainers convert from **milliseconds** and pass down) here https://docs.docker.com/reference/api/engine/version/v1.48/#tag/Container/operation/ContainerCreate

Testcontainers APIs use **millisecond** consistently. I have documented timeout values to say which units they are expecting (seconds or millis, etc).

### DIFF
--- a/docs/features/compose.md
+++ b/docs/features/compose.md
@@ -140,11 +140,11 @@ const environment = await new DockerComposeEnvironment(composeFilePath, composeF
 await environment.down();
 ```
 
-If you need to wait for the environment to be downed, you can provide a timeout:
+If you need to wait for the environment to be downed, you can provide a timeout. The unit of timeout here is **second**:
 
 ```javascript
 const environment = await new DockerComposeEnvironment(composeFilePath, composeFile).up();
-await environment.down({ timeout: 10000 }); // ms
+await environment.down({ timeout: 10 }); // timeout after 10 seconds
 ```
 
 Volumes created by the environment are removed when stopped. This is configurable:
--- a/docs/features/containers.md
+++ b/docs/features/containers.md
@@ -334,11 +334,11 @@ const container = await new GenericContainer("alpine").start();
 await container.stop();
 ```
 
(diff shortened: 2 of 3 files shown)


docs/features/compose.md
```diff
@@ -140,11 +140,11 @@ const environment = await new DockerComposeEnvironment(composeFilePath, composeF
 await environment.down();
 ```
 
-If you need to wait for the environment to be downed, you can provide a timeout:
+If you need to wait for the environment to be downed, you can provide a timeout. The unit of timeout here is **second**:
 
 ```javascript
 const environment = await new DockerComposeEnvironment(composeFilePath, composeFile).up();
-await environment.down({ timeout: 10000 }); // ms
+await environment.down({ timeout: 10 }); // timeout after 10 seconds
```
cristianrgreco, line 147: I wasn't aware this was in seconds, same in `containers.md`.

I've created an issue to convert them from seconds to milliseconds (https://github.com/testcontainers/testcontainers-node/issues/960). It'll be a breaking change, but I'm planning one anyway. Please feel free to pick it up.

cristianrgreco (approved):

docs/features/compose.md
digital88: I mean the docker API expects seconds from caller (testcontainers). Tescontainers just pass down the value user provided to docker API. You want users to pass value here as milliseconds explicitly?

cristianrgreco: > You want users to pass value here as milliseconds explicitly?

Yeah I'd say so. We already convert from other units for consistency, seems we just missed this one.

Issue #960, opened 2025-03-24
Container stop timeouts should be in milliseconds instead of seconds for consistency


cristianrgreco:
Remember to update the docs as well.

cristianrgreco:
Completed in 11.0.0.

PR #962, opened 2025-03-26
Container stop timeouts should be in milliseconds

digital88:
#960

+ Added class `Ms` which provides conversion from ms to nanos and seconds.
+ Explicitly call this class' conversion methods when calling Docker APIs. Example:

```js
...
log.debug(`Stopping container...`, { containerId: container.id });
const t = new Ms(opts?.timeout ?? 0).seconds();
await container.stop({ t });
...
```

+ Renamed parameters in all public APIs to tell that value in ms expected to be passed.
+ Renamed some internal fields and methods to tell that value in ms expected there.
+ Added docs for timeouts.
+ Updated old docs where timeouts are used.


packages/testcontainers/src/container-runtime/clients/container/docker-container-client.ts
```diff
@@ -131,7 +130,7 @@ export class DockerContainerClient implements ContainerClient {
   async stop(container: Container, opts?: { timeout: number }): Promise<void> {
     try {
       log.debug(`Stopping container...`, { containerId: container.id });
-      await container.stop({ t: opts?.timeout });
+      await container.stop({ t: toSeconds(opts?.timeout ?? 0) });
```
cristianrgreco, line 133: Note to self: these are the reasons why this is a breaking change

PR #990, opened 2025-04-16
Allow overriding default compose wait strategy

jalaziz:
Allow the default wait strategy to be overriden for docker compose environments. The default "listening ports" strategy doesn't work for distroless images and in complex setups, having to explicitly override the wait strategy for each service is error-prone.


jalaziz:
pushed a change to fix the linting failure.

jalaziz:
All tests passed with the previous commit. I just added docs to hopefully get this over the line 🙏🏼

anapsix:
This change is extremely helpful, especially while working with distroless containers. Hoping to see it merged soon.

cristianrgreco (approved):

PR #994, opened 2025-05-01
Support Neo4j plugins other than Apoc

### LINKED ISSUE #993: Neo4j container: support plugins other than Apoc
The standard Neo4j docker image supports [a selection of commonly used plugins](https://neo4j.com/docs/operations-manual/current/configuration/plugins/). They can be set up by passing a list of plugins as env variables.

Currently the Neo4j Testcontainer for Node only supports standaard Apoc, by using `withApoc `method. It works fine, but other plugins like Apoc Extended or Graph Algorithms can not be setup similarly. They probably could be set using raw env values, but that would then conflict with what `withApoc` sets up.

Instead of adding convenience methods for them all, I would suggest we add `withPlugins` method that accepts a list of plugins to setup. Could be based on enumeration of supported plugins, or just left to the image used. It should error if incorrect plugins are included. I believe this is how Java and Go implementations of TestContainers handle other plugins.

I'll probably just open PR about this myself, but adding this issue in advance for possible comments.

stscoundrel:
Add withPlugins method to Neo4jContainer for setting up plugins included in the docker image. Previously only withApoc was provided for setting up Apoc-plugin. Now withApoc uses the same plugin list under-the-hood.

Availability of specific plugins may depend on what version of Neo4j image is used.

Closes #993

### DIFF
--- a/packages/modules/neo4j/src/neo4j-container.test.ts
+++ b/packages/modules/neo4j/src/neo4j-container.test.ts
@@ -1,5 +1,5 @@
 import neo4j from "neo4j-driver";
-import { Neo4jContainer } from "./neo4j-container";
+import { Neo4jContainer, Neo4jPlugin } from "./neo4j-container";
 
 describe("Neo4jContainer", { timeout: 180_000 }, () => {
   // createNode {
@@ -77,6 +77,36 @@ describe("Neo4jContainer", { timeout: 180_000 }, () => {
     const result = await session.run("CALL apoc.help('text')");
     const singleRecord = result.records[0];
     expect(singleRecord.length).toBeGreaterThan(0);
+
+    await session.close();
+    await driver.close();
+    await container.stop();
+  });
+  // }
+
+  // pluginsList {
+  it("should work with plugin list", async () => {
+    const container = await new Neo4jContainer("neo4j:5.26.5")
+      .withPlugins([Neo4jPlugin.APOC_EXTENDED, Neo4jPlugin.GRAPH_DATA_SCIENCE])
+      .withStartupTimeout(120_000)
+      .start();
+    const driver = neo4j.driver(
+      container.getBoltUri(),
+      neo4j.auth.basic(container.getUsername(), container.getPassword())
+    );
+
+    const session = driver.session();
+
+    // Monitor methods are only available in extended Apoc.
+    const result = await session.run("CALL apoc.monitor.ids()");
+    expect(result.records[0].get("nodeIds")).toEqual({ high: 0, low: 0 });
+
+    // Insert one node.
+    await session.run("CREATE (a:Person {name: $name}) RETURN a", { name: "Some dude" });
+
+    // Monitor result should reflect increase in data.
+    const result2 = await session.run("CALL apoc.monitor.ids()");
+    expect(result2.records[0].get("nodeIds")).toEqual({ high: 0, low: 1 });
 
     await session.close();
     await driver.close();
--- a/packages/modules/neo4j/src/neo4j-container.ts
+++ b/packages/modules/neo4j/src/neo4j-container.ts
@@ -4,10 +4,28 @@ const BOLT_PORT = 7687;
 const HTTP_PORT = 7474;
 const USERNAME = "neo4j";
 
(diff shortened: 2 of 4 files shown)


packages/modules/neo4j/src/neo4j-container.test.ts
```diff
@@ -83,4 +83,34 @@ describe("Neo4jContainer", { timeout: 180_000 }, () => {
     await container.stop();
   });
   // }
+
+  // plugins list {
```
cristianrgreco, line 87: Could you please add this example to the module docs?

packages/modules/neo4j/src/neo4j-container.ts
```diff
@@ -44,10 +56,29 @@ export class Neo4jContainer extends GenericContainer {
         NEO4J_apoc_ttl_schedule: this.ttl.toString(),
       });
     }
-    if (this.apoc) {
+
+    if (this.plugins.length > 0) {
+      const whitelists: string[] = [];
+
+      this.plugins.forEach((plugin) => {
+        if (plugin === Neo4jPlugin.APOC || plugin === Neo4jPlugin.APOC_EXTENDED) {
```
cristianrgreco, line 64: We could simplify/reduce the code a bit if we combined the enum with the whitelist values. Something like:

```ts
export const PLUGINS = {
  APOC: { whitelist: "apoc.*" },
  APOC_EXTENDED: { whitelist: "apoc.*" },
  BLOOM: { whitelist: "bloom.*" },
  GRAPH_DATA_SCIENCE: { whitelist: "gds.*" },
} as const;

type Neo4jPlugin = keyof typeof PLUGINS;
```

That way it'll be easier to add new plugins. This code can also then be refactored like:

```ts
const whitelists = this.plugins.reduce((acc, plugin) => {
  const entry = plugins[plugin];
  if (entry?.whitelist) {
    acc.push(entry.whitelist);
  }
  return acc;
}, []);
```

packages/modules/neo4j/src/neo4j-container.test.ts
stscoundrel: Whoops, forgot these had to be manually added. Added to docs 👍

packages/modules/neo4j/src/neo4j-container.ts
stscoundrel: I agree that this should be made a bit easier to maintain.

While your approach seems good, I opted to keep the plugins & whitelist separate. Whitelists are promoted to top level record, looping of them is done with combination of map, filter and join. Should allow people to configure whitelists without touching the methods.

cristianrgreco: > Whitelists are promoted to top level record, looping of them is done with combination of map, filter and join. Should allow people to configure whitelists without touching the methods.

Sorry I didn't understand this. When I say "that way it'll be easier to add new plugins", I mean for someone to contribute a new plugin to this code. They just add a new line to the `PLUGINS`, instead of scrolling down and adding another `if` statement.

stscoundrel: That is quite close to how I meant it. Instead of it being one variable (combination of plugin name and a possible whitelist) it is just two different declarations next to each other. Meaning this part:

```
export enum Neo4jPlugin {
  APOC = "apoc",
  APOC_EXTENDED = "apoc-extended",
  BLOOM = "bloom",
  GEN_AI = "genai",
  GRAPHQL = "graphql",
  GRAPH_ALGORITHMS = "graph-algorithms",
  GRAPH_DATA_SCIENCE = "graph-data-science",
  NEO_SEMANTICS = "n10s",
}

const pluginWhitelists: Partial<Record<Neo4jPlugin, string>> = {
  [Neo4jPlugin.APOC]: "apoc.*",
  [Neo4jPlugin.APOC_EXTENDED]: "apoc.*",
  [Neo4jPlugin.BLOOM]: "bloom.*",
  [Neo4jPlugin.GRAPH_DATA_SCIENCE]: "gds.*",
};
```

So there is no need to touch `if` statements.

cristianrgreco (approved):

PR #1004, opened 2025-05-18
Add etcd module

prenaissance:
This PR adds [etcd](https://etcd.io/) as a new module.

I've implemented a single-node initialization only. If anyone would benefit from a multi-node cluster testcontainer, please let me know through comments.


prenaissance:
Could we try a re-run on the failed jobs? I'll look for workarounds if this keeps failing on CI. Tested locally using `act` and `npm run test:ci packages/modules/etcd`:

image

image

cristianrgreco:
The test is failing because it took longer than 5s to download the etcd image. Increase the test timeout, for example as done here: https://github.com/testcontainers/testcontainers-node/blob/87a5e5eb0792a45612a4489441d33f289df2bad4/packages/modules/kafka/src/kafka-container.test.ts#L7

packages/modules/etcd/src/etcd-container.ts
```diff
@@ -0,0 +1,46 @@
...
+      `http://0.0.0.0:${ETCD_CLIENT_PORT}`,
+      "--listen-peer-urls",
+      `http://0.0.0.0:${ETCD_PEER_PORT}`,
+      "--listen-client-urls",
+      `http://0.0.0.0:${ETCD_CLIENT_PORT}`,
+    ]);
+  }
+
+  public override async start(): Promise<StartedEtcdContainer> {
+    this.withWaitStrategy(Wait.forLogMessage(/"status":"SERVING"/));
```
cristianrgreco, line 25: Should move this to the constructor. Setting the wait strategy here will override any user-provided wait strategy.

prenaissance:
Thanks for the hint! With the image removed it took 13 seconds locally to run the spec. I set the timeout to 30 seconds, hopefully that works for the CI run.

cristianrgreco:
@prenaissance Thanks for the contribution!

cristianrgreco:
@prenaissance I just realised the containers aren't exported from the `index.ts` https://github.com/testcontainers/testcontainers-node/blob/main/packages/modules/etcd/src/index.ts. Please feel free to raise a PR to export them 🙂

prenaissance:
My bad! Omw to fix this.

Issue #827, opened 2024-08-28
Update redis module so that it can work with redis-stack similar to testcontainers-go


kevbook:
**Expected Behaviour**

```js
const container = new RedisContainer('redis/redis-stack-server:7.2.0-v12')
```

This will fail because of the `redis-server` command. https://github.com/testcontainers/testcontainers-node/blob/9f7fd4eff5e6f05440016010964a78aa80557cb1/packages/modules/redis/src/redis-container.ts#L36-L39

As you can see from the docker images, `redis` uses `CMD [redis-server]` but `redis-stack-server` uses a custom entrypoint script. So I'm happy to open a PR, basically based on image string passed, we'll need to change how `container.start()` is handled (It's messy but not sure how else to solve it when you have 2 Redis images now!)

- https://hub.docker.com/layers/library/redis/7.2/images/sha256-74de23f6746eb84af59a6ae86e5d3c4306078dd2a45cb02bf1556ac438436e95?context=explore
- https://hub.docker.com/layers/redis/redis-stack-server/7.2.0-v12/images/sha256-af91865c72c3c51b62a15fb3bcf9eae8d958672820fc91ae5c0b4c39ff5c9b53?context=explore (redis/redis-stack-server:7.2.0-v12)

kevbook:
Just add this line. All tests should pass

```js
/* Make sure the redis image is a not a `redis-stack` image as `redis-stack` images have a custom entrypoint, see https://hub.docker.com/r/redis/redis-stack-server */
    if (!this.imageName.string.includes("redis-stack")) {
      this.withCommand([
        "redis-server",
        ...(this.password ? [`--requirepass "${this.password}"`] : []),
        ...(this.persistenceVolume ? ["--save 1 1 ", "--appendonly yes"] : []),
      ]);
    }
```

cristianrgreco:
@kevbook PR is welcome

digital88:
@kevbook Hi, this runs without errors (at least on e7c499ebd06763c8ce9062db6e7c08798cc248aa):

```ts
  it("should start from redis-stack-server image", async () => {
    const container = await new RedisContainer("redis/redis-stack-server:7.2.0-v12").start();

    await container.stop();
  });
```

crcarlo:
@digital88 @cristianrgreco Hi! Event though the container starts by using `redis/redis-stack-server` image, it doesn't seem to have JSON module enabled as you can verify from the test below.

[...]

PR #1005, opened 2025-05-20
Add support for redis-stack containers

crcarlo:
Using the current version of `RedisContainer` with `redis/redis-stack-server` or `redis/redis-stack` images, the container lacks Redis modules because `start` method overrides the entrypoint script of redis-stack.

This problem has been described in this thread https://github.com/testcontainers/testcontainers-node/issues/827#issuecomment-2894254729

This PR:
- Uses default redis-stack entrypoint script on start
- Adds a test for redis-stack with JSON module
- Updates the documentation to show the test as an example

### DIFF
--- a/packages/modules/redis/src/redis-container.test.ts
+++ b/packages/modules/redis/src/redis-container.test.ts
@@ -102,6 +102,7 @@ describe("RedisContainer", { timeout: 240_000 }, () => {
   });
   // }
 
+  // startWithRedisStack {
   it("should start with redis-stack-server and json module", async () => {
     const container = await new RedisContainer("redis/redis-stack-server:latest").withPassword("testPassword").start();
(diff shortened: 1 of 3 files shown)


packages/modules/redis/src/redis-container.test.ts
```diff
@@ -102,6 +102,20 @@ describe("RedisContainer", { timeout: 240_000 }, () => {
   });
   // }
 
+  // startWithRedisStack {
+  it("should start with redis-stack-server and json module", async () => {
+    const container = await new RedisContainer("redis/redis-stack-server:latest").withPassword("testPassword").start();
```
cristianrgreco, line 107: Let's pin a version instead of using `latest`, otherwise this test will break over time

cristianrgreco (approved):

PR #1044, opened 2025-06-12
Add support for custom client options in DockerComposeEnvironment

### LINKED ISSUE #1043: Feature Request: Add standalone Option to DockerComposeClient in DockerComposeEnvironment
**Is your feature request related to a problem? Please describe.**

My team utilizes `docker-compose` as a standalone binary, and we currently lack a mechanism to pass the `standalone` option through the `DockerComposeClient` when using `DockerComposeEnvironment`. This limits our ability to leverage Testcontainers-Node with our existing `docker-compose` setup.

The `standalone` flag is documented in the underlying `docker-compose` library's README: [https://github.com/PDMLab/docker-compose?tab=readme-ov-file#standalone-mode](https://github.com/PDMLab/docker-compose?tab=readme-ov-file#standalone-mode).

**Describe the solution you'd like**

I would like to see an option added to `DockerComposeClient` (and consequently accessible via `DockerComposeEnvironment`) that allows users to specify the `standalone` flag for the `docker-compose` executable.

Ideally, this could be achieved by extending the `ComposeOptions` interface to include an `executable` property, similar to how it's handled in the `PDMLab/docker-compose` library:

```typescript
interface ComposeOptions{
  // ... existing options
  executable?: {
    standalone?: boolean;
    executablePath?: string;
  };
}
```

This would allow users to configure `DockerComposeEnvironment` like so:

```typescript
const environment = new DockerComposeEnvironment(
  path.resolve(__dirname, './compose-files'),
  'docker-compose.yml'
);

environment.withClientOptions({
  executable: {
    standalone: true,
    executablePath: '/usr/local/bin/docker-compose', // Example custom path
  },
});

await environment.up();
```

**Describe alternatives you've considered**

Currently, we are constrained to keep using Testcontainers-Node v10, as v11 updates the underlying library `docker-compose` from v0 to v1, which removed the support of `docker compose` version 1 and standalone binary out of the box.

**Additional context**

This feature would significantly improve the flexibility and usability of Testcontainers-Node for teams that rely on standalone `docker-compose` installations. It would enable seamless integration and allow us to leverage the benefits of Testcontainers for managing our testing infrastructure.

Dyskal:
Fixes #1043

### DIFF
--- a/packages/testcontainers/src/container-runtime/clients/compose/types.ts
+++ b/packages/testcontainers/src/container-runtime/clients/compose/types.ts
@@ -1,4 +1,3 @@
-import { IDockerComposeExecutableOptions } from "docker-compose";
 import { Logger } from "../../../common";
 
 export type ComposeOptions = {
@@ -9,8 +8,20 @@ export type ComposeOptions = {
   composeOptions?: string[];
   environment?: NodeJS.ProcessEnv;
   logger?: Logger;
-  executable?: IDockerComposeExecutableOptions;
+  executable?: ComposeExecutableOptions;
 };
+
+export type ComposeExecutableOptions =
+  | {
+      executablePath: string;
+      options?: string[] | (string | string[])[];
+      standalone?: never;
+    }
+  | {
+      executablePath?: string;
+      options?: never;
+      standalone: true;
+    };
 
 export type ComposeDownOptions = {
   timeout: number;
--- a/packages/testcontainers/src/container-runtime/index.ts
+++ b/packages/testcontainers/src/container-runtime/index.ts
@@ -1,6 +1,6 @@
 export { getAuthConfig } from "./auth/get-auth-config";
 export { ContainerRuntimeClient, getContainerRuntimeClient } from "./clients/client";
 export { parseComposeContainerName } from "./clients/compose/parse-compose-container-name";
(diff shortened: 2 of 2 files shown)


packages/testcontainers/src/container-runtime/clients/compose/types.ts
```diff
@@ -8,6 +9,7 @@ export type ComposeOptions = {
   composeOptions?: string[];
   environment?: NodeJS.ProcessEnv;
   logger?: Logger;
+  executable?: IDockerComposeExecutableOptions;
```
cristianrgreco: Let's create and expose (add export to `index.ts`) our own `DockerComposeClientOptions` interface. 

That way our users:
1. Don't have to import types from a transitive dependency
2. If docker-compose introduce a breaking change tomorrow, we have the option to handle it without a breaking change for our users

cristianrgreco:
I wonder if this can be tested by spinning up an image of Docker (https://hub.docker.com/_/docker), mounting the compose executable somewhere in the host, and specifying that as the executable.

cristianrgreco:
@Dyskal are you planning on adding tests?

Dyskal:
Sorry for the delay, I don't know if I have the knowledge in docker to do such tests...

PR #1051, opened 2025-07-04
Change KurrentDB protocol from `esdb` to `kurrentdb`

botflux:
## Changes
`getConnectionString` now:
- Uses `kurrentdb://` by default
- Falls back to `esdb://` if the image name contains `eventstore`

## Questions for review
1. **String matching**: Should I use a stricter pattern like `eventstore/eventstore` instead of just checking if the image name contains `eventstore`?

2. **Compatibility concern**: Looking at my tests, `KurrentDBContainer` cannot start with old EventStore images without overriding environment variables (the env prefix changed from `EVENTSTORE_` to `KURRENT_`). Given this limitation, should we even bother with the `esdb://` fallback since the container won't work out of the box with EventStore images anyway?

What are your thoughts @cristianrgreco?


cristianrgreco:
> Compatibility concern: Looking at my tests, KurrentDBContainer cannot start with old EventStore images without overriding environment variables (the env prefix changed from EVENTSTORE_ to KURRENT_). Given this limitation, should we even bother with the esdb:// fallback since the container won't work out of the box with EventStore images anyway?

Great spot! My only reason for maintaining `esdb` was backward compatibility with eventstore, but if there isn't any (it was a major release after all) then it's not necessary. Looks like we can simply `s/esdb/kurrentdb/`. Thanks @botflux

botflux:
Done! Now `kurrentdb://` is hardcoded.

PR #1053, opened 2025-07-05
Add support for Confluent Kafka 8

cristianrgreco:
TODO: Fix doc format https://deploy-preview-1053--testcontainers-node.netlify.app/modules/kafka/


packages/modules/kafka/src/kafka-container.ts
```diff
@@ -64,6 +65,12 @@ export class KafkaContainer extends GenericContainer {
 
   constructor(image: string) {
     super(image);
+
+    const parsedImage = ImageName.fromString(image);
```
cristianrgreco: TODO: refactor this and `isLessThanCP` to some semantic version comparison util. This looks good https://www.npmjs.com/package/compare-versions. Something like `satisfies(tag, '>=8.0.0'); // true/false`

PR #1068, opened 2025-07-08
Implement support for specifying the protocol when exposing ports

technocidal:
Relates to #556. 

Follows the commonly known Docker format for specifying ports with a protocol. Backwards compatibility has been tested. If no protocol is specified, it'll default to `tcp` which is all that was possible with `testcontainers-node` up until now anyway.

### DIFF
--- a/packages/testcontainers/src/test-container.ts
+++ b/packages/testcontainers/src/test-container.ts
@@ -71,7 +71,7 @@ export interface StartedTestContainer extends AsyncDisposable {
   getHost(): string;
   getHostname(): string;
   getFirstMappedPort(): number;
-  getMappedPort(port: number, protocol?: string): number;
+  getMappedPort(port: string | number, protocol?: string): number;
   getName(): string;
   getLabels(): Labels;
   getId(): string;
--- a/packages/testcontainers/src/test-container.ts
+++ b/packages/testcontainers/src/test-container.ts
@@ -71,7 +71,7 @@ export interface StartedTestContainer extends AsyncDisposable {
   getHost(): string;
   getHostname(): string;
   getFirstMappedPort(): number;
-  getMappedPort(port: number): number;
+  getMappedPort(port: number, protocol?: string): number;
   getName(): string;
   getLabels(): Labels;
   getId(): string;
--- a/packages/testcontainers/src/test-container.ts
+++ b/packages/testcontainers/src/test-container.ts
@@ -71,7 +71,8 @@ export interface StartedTestContainer extends AsyncDisposable {
   getHost(): string;
   getHostname(): string;
   getFirstMappedPort(): number;
-  getMappedPort(port: string | number, protocol?: string): number;
+  getMappedPort(port: number, protocol?: string): number;
+  getMappedPort(portWithProtocol: `${number}/${"tcp" | "udp"}`): number;
   getName(): string;
   getLabels(): Labels;
   getId(): string;
--- a/packages/testcontainers/src/utils/bound-ports.ts
+++ b/packages/testcontainers/src/utils/bound-ports.ts
@@ -10,8 +10,6 @@ export class BoundPorts {
     const normalizedProtocol = protocol.toLowerCase();
     const key = typeof port === "string" ? port : `${port}/${normalizedProtocol}`;
     let binding = this.ports.get(key);
-
-    // Try case-insensitive lookup if not found and port is a string that might contain a protocol
     if (!binding && typeof port === "string" && port.includes("/")) {
       const [portNumber, portProtocol] = port.split("/");
       const normalizedKey = `${portNumber}/${portProtocol.toLowerCase()}`;
@@ -36,16 +34,13 @@ export class BoundPorts {
   }
 
   public setBinding(key: string | number, value: number, protocol: string = "tcp"): void {
-    // Normalize protocol to lowercase for consistency
     const normalizedProtocol = protocol.toLowerCase();
 
     if (typeof key === "string" && key.includes("/")) {
-      // If key contains protocol, normalize it
       const [portNumber, portProtocol] = key.split("/");
       const normalizedKey = `${portNumber}/${portProtocol.toLowerCase()}`;
       this.ports.set(normalizedKey, value);
     } else {
-      // Otherwise use the provided key/protocol
       const portKey = typeof key === "string" ? key : `${key}/${normalizedProtocol}`;
       this.ports.set(portKey, value);
     }
@@ -57,8 +52,6 @@ export class BoundPorts {
 
   public filter(ports: PortWithOptionalBinding[]): BoundPorts {
     const boundPorts = new BoundPorts();
-
-    // Create map of port to protocol for lookup
     const containerPortsWithProtocol = new Map<number, string>();
     ports.forEach((port) => {
       const containerPort = getContainerPort(port);
@@ -69,8 +62,6 @@ export class BoundPorts {
     for (const [internalPortWithProtocol, hostPort] of this.iterator()) {
       const [internalPortStr, protocol] = internalPortWithProtocol.split("/");
       const internalPort = parseInt(internalPortStr, 10);
-
-      // Case-insensitive protocol comparison
       if (
         containerPortsWithProtocol.has(internalPort) &&
         containerPortsWithProtocol.get(internalPort)?.toLowerCase() === protocol?.toLowerCase()
--- a/packages/testcontainers/src/utils/bound-ports.ts
+++ b/packages/testcontainers/src/utils/bound-ports.ts
@@ -1,16 +1,17 @@
 import net from "net";
 import { HostIp } from "../container-runtime";
 import { HostPortBindings, InspectResult } from "../types";
-import { getContainerPort, PortWithOptionalBinding } from "./port";
+import { getContainerPort, getProtocol, PortWithOptionalBinding } from "./port";
 
 export class BoundPorts {
-  private readonly ports = new Map<number, number>();
+  private readonly ports = new Map<string, number>();
 
-  public getBinding(port: number): number {
-    const binding = this.ports.get(port);
+  public getBinding(port: number | string, protocol: string = "tcp"): number {
+    const key = typeof port === "string" ? port : `${port}/${protocol}`;
+    const binding = this.ports.get(key);
 
     if (!binding) {
-      throw new Error(`No port binding found for :${port}`);
+      throw new Error(`No port binding found for :${key}`);
     }
 
     return binding;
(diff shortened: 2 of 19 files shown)


technocidal:
@cristianrgreco I'll hopefully get this into a publishable state very soon. It's currently logging way too much and some stuff needs cleaning up. I'm however using this version already at my job (telephony/voice engineering where UDP is quite important) for integration testing and it works quite well! Thanks for your patience.

cristianrgreco:
@technocidal No problem at all. Thanks for working on it! Let me know if if there's anything I can help with

packages/testcontainers/src/utils/port.ts
```diff
@@ -1,13 +1,39 @@
...
-export type PortWithOptionalBinding = number | PortWithBinding;
+export type PortWithOptionalBinding = number | string | PortWithBinding;
 
-export const getContainerPort = (port: PortWithOptionalBinding): number =>
-  typeof port === "number" ? port : port.container;
+export const getContainerPort = (port: PortWithOptionalBinding): number => {
+  if (typeof port === "number") {
+    return port;
+  } else if (typeof port === "string") {
+    const match = port.match(/^(\d+)(?:\/(udp|tcp))?$/);
```
cristianrgreco: Missing the case insensitive flag here.

The unit tests that have been added are great but we should add an integration test too. For example the following test is failing with this error:

```ts
await using container = await new GenericContainer("mendhak/udp-listener")
  .withExposedPorts("5005/UDP")
  .start();
```

Error: Invalid port format: "5005/UDP"

---

May be worth extracting the `RegExp` to a variable, as it's used in both functions. Something like:

```ts
const portRegex = RegExp(/^(\d+)(?:\/(udp|tcp))?$/).exec(port)
...
const match = portRegex.exec(port)
```

packages/testcontainers/src/utils/port.ts
```diff
@@ -1,13 +1,39 @@
...
 export const hasHostBinding = (port: PortWithOptionalBinding): port is PortWithBinding => {
   return typeof port === "object" && port.host !== undefined;
 };
+
+export const getProtocol = (port: PortWithOptionalBinding): string => {
+  if (typeof port === "number") {
+    return "tcp";
+  } else if (typeof port === "string") {
+    const match = port.match(/^(\d+)(?:\/(udp|tcp))?$/i);
+    if (match && match[2]) {
```
cristianrgreco:
```suggestion
    if (match?.[2]) {
```

packages/testcontainers/src/wait-strategies/host-port-wait-strategy.test.ts
```diff
@@ -20,7 +20,7 @@ describe("HostPortWaitStrategy", { timeout: 180_000 }, () => {
         .withExposedPorts(8081)
         .withStartupTimeout(0)
         .start()
-    ).rejects.toThrowError(/Port \d+ not bound after 0ms/);
+    ).rejects.toThrowError(/Port (?:\d+|\d+\/\w+) not bound after 0ms/);
```
cristianrgreco: There are no cases where the protocol won't be included

```suggestion
    ).rejects.toThrowError(/Port \d+\/(tcp|udp) not bound after 0ms/);
```

packages/testcontainers/src/wait-strategies/host-port-wait-strategy.ts
```diff
@@ -36,22 +42,55 @@ export class HostPortWaitStrategy extends AbstractWaitStrategy {
...
-      () => portCheck.isBound(port),
-      (isBound) => isBound,
-      () => {
-        const message = `Port ${port} not bound after ${this.startupTimeoutMs}ms`;
-        log.error(message, { containerId: container.id });
-        throw new Error(message);
-      },
-      this.startupTimeoutMs
-    );
+  private async waitForPort(
```
cristianrgreco, line 58: The changes to this method aren't needed. The skipping of the UDP ports is already handled by the `waitForInternalPorts` and `waitForHostPorts`.

packages/testcontainers/src/wait-strategies/host-port-wait-strategy.ts
```diff
@@ -36,22 +42,55 @@ export class HostPortWaitStrategy extends AbstractWaitStrategy {
     boundPorts: BoundPorts
   ): Promise<void> {
     for (const [internalPort] of boundPorts.iterator()) {
+      if (typeof internalPort === "string" && internalPort.toLowerCase().includes("/udp")) {
```
cristianrgreco: `internalPort` is always a `string`

```suggestion
      if (internalPort.toLowerCase().includes("/udp")) {
```

packages/testcontainers/src/utils/bound-ports.ts
```diff
@@ -1,16 +1,23 @@
...
+import { getContainerPort, getProtocol, PortWithOptionalBinding } from "./port";
 
 export class BoundPorts {
-  private readonly ports = new Map<number, number>();
-
-  public getBinding(port: number): number {
-    const binding = this.ports.get(port);
+  private readonly ports = new Map<string, number>();
+
+  public getBinding(port: number | string, protocol: string = "tcp"): number {
```
cristianrgreco, line 9: We can avoid the double lookup (where protocol is not provided, but the binding has the protocol), with something like this:

```ts
public getBinding(port: number | string, protocol: string = "tcp"): number {
  let key: string;

  if (typeof port === "string" && port.includes("/")) {
    const [portNumber, portProtocol] = port.split("/");
    key = `${portNumber}/${portProtocol.toLowerCase()}`;
  } else {
    key = `${port}/${protocol.toLowerCase()}`;
  }

  const binding = this.ports.get(key);

  if (!binding) {
    throw new Error(`No port binding found for :${key}`);
  }

  return binding;
}
```

packages/testcontainers/src/utils/port.ts
technocidal: Yes. 

Integration test is definitely missing. I thought about adding UDP to the helloworld container, this looks like a valid option for now. I can always switch it over later.

packages/testcontainers/src/test-container.ts
```diff
@@ -71,7 +71,7 @@ export interface StartedTestContainer extends AsyncDisposable {
   getHost(): string;
   getHostname(): string;
   getFirstMappedPort(): number;
-  getMappedPort(port: number): number;
+  getMappedPort(port: string | number, protocol?: string): number;
```
cristianrgreco: The existing `getMappedPort` function only accepts a number. Meaning we can be restrictive of the new string type we're introducing here. 

The problem I'd like to solve is, at the moment the following is valid:

```ts
container.getMappedPort("5005/udp", "tcp")
```

We can prevent this by introducing the following overloads:

```ts
getMappedPort(portWithProtocol: `${number}/${"tcp" | "udp"}`): number;
getMappedPort(port: number, protocol?: string): number;
```

So basically, if the user provides a string, it must include the protocol (fair enough, a number makes more sense if you don't want to specify the protocol). If they provide a number, then they can optionally provide the protocol, if not we default it. We no longer have this issue where the user can provide two conflicting values. WDYT?

technocidal: That's a good idea. And we can probably do this in for `withExposedPorts`/`PortWithOptionalBinding` as well. Doesn't get rid of the Regex but it's a better experience using the API imo.

packages/testcontainers/src/wait-strategies/host-port-wait-strategy.ts
technocidal: I did revert this as well. GitHub is probably not showing this as outdated, because the comment is on the first line of the function.

packages/testcontainers/src/utils/port.test.ts
```diff
@@ -0,0 +1,75 @@
+import { describe, expect } from "vitest";
+import { getContainerPort, getProtocol, hasHostBinding } from "./port";
+
+describe("port utilities", () => {
```
cristianrgreco, line 4: There are compilation issues in this test suite after the API change, e.g. `parses port from string`. There's basically one less case to handle (where port is just a string without protocol).

Another one:

```ts
it("throws error for invalid port format", () => {
  expect(() => getContainerPort("invalid")).toThrow("Invalid port format: invalid");
});
```

Actually it's no longer possible to have this runtime error as we now catch it at compile time

packages/testcontainers/src/wait-strategies/host-port-wait-strategy.ts
```diff
@@ -22,7 +22,13 @@ export class HostPortWaitStrategy extends AbstractWaitStrategy {
     container: Dockerode.Container,
     boundPorts: BoundPorts
   ): Promise<void> {
-    for (const [, hostPort] of boundPorts.iterator()) {
+    for (const [portKey, hostPort] of boundPorts.iterator()) {
+      if (portKey.toLowerCase().includes("/udp")) {
```
cristianrgreco: It's more precise to replace `includes` with `endsWith`, same in `waitForInternalPorts`

packages/testcontainers/src/wait-strategies/utils/port-check.ts
```diff
@@ -4,15 +4,21 @@ import { log } from "../../common";
...
-  isBound(port: number): Promise<boolean>;
+  isBound(port: number | string): Promise<boolean>;
 }
 
 export class HostPortCheck implements PortCheck {
   constructor(private readonly client: ContainerRuntimeClient) {}
 
-  public isBound(port: number): Promise<boolean> {
+  public isBound(port: number | string): Promise<boolean> {
+    if (typeof port === "string" && port.toLowerCase().includes("/udp")) {
```
cristianrgreco: It's more precise to replace `includes` with `endsWith`, same in `InternalPortCheck`

packages/testcontainers/src/wait-strategies/host-port-wait-strategy.ts
technocidal: Changed

packages/testcontainers/src/wait-strategies/utils/port-check.ts
technocidal: Changed

packages/testcontainers/src/utils/port.test.ts
technocidal: Removed all broken tests. Sorry for not catching this before pushing.

cristianrgreco:
Looks like there’s a compilation issue

technocidal:
> Looks like there’s a compilation issue

Should be addressed in the latest commit

cristianrgreco:
Thanks @technocidal for your contribution, and patience with reviewing the PR!

technocidal:
@cristianrgreco And thanks for your patience with all of my little mistakes and oversights and especially for your valuable feedback. Have a nice weekend!

PR #1070, opened 2025-07-09
Add support for MongoDBContainer credentials

### LINKED ISSUE #988: @testcontainers/mongodb - how can creds to require be passed?
I cannot find an example for it in the docs.

I tried to 
```js
const mongo = new MongoDBContainer()
      .withEnvironment({ 
         MONGO_INITDB_ROOT_USERNAME: dbUsername, 
         MONGO_INITDB_ROOT_PASSWORD: dbPassword,
      })
      .start();
```

but it appears that `mongo.getConnectionString()` does not include the creds...

digital88:
Hopefully closes #988

This implementation supports credentials only when replica set is disabled. Because using credentials with replica set requires rather complicated setup I think it could be implemented later as separate PR. I'm not even sure _why_ replica set was implemented in first place, as TC creates single container for running tests it does not make much sense IMO. Either way the old replica set code remains untouched and is used by default, so nothing should break for existing users.

### DIFF
--- a/packages/modules/mongodb/src/mongodb-container.test.ts
+++ b/packages/modules/mongodb/src/mongodb-container.test.ts
@@ -6,10 +6,9 @@ const IMAGE = getImage(__dirname);
 
 describe("MongoDBContainer", { timeout: 240_000 }, () => {
   it.each([IMAGE, "mongo:6.0.25", "mongo:4.4.29"])("should work with %s", async (image) => {
-    // mongoConnect {
+    // connectMongo {
     await using mongodbContainer = await new MongoDBContainer(image).start();
 
-    // directConnection: true is required as the container is created as a MongoDB Replica Set.
     const db = mongoose.createConnection(mongodbContainer.getConnectionString(), { directConnection: true });
     const fooCollection = db.collection("foo");
     const obj = { value: 1 };
@@ -25,7 +24,7 @@ describe("MongoDBContainer", { timeout: 240_000 }, () => {
   });
 
   it("should connect with credentials", async () => {
-    // mongoConnectWithCredentials {
+    // connectWithCredentials {
     await using mongodbContainer = await new MongoDBContainer(IMAGE)
       .withUsername("mongo_user")
       .withPassword("mongo_password")
--- a/docs/modules/mongodb.md
+++ b/docs/modules/mongodb.md
@@ -11,9 +11,9 @@ npm install @testcontainers/mongodb --save-dev
 ## Examples
 
 <!--codeinclude-->
-[Connect:](../../packages/modules/mongodb/src/mongodb-container.test.ts) inside_block:mongoConnect
+[Connect:](../../packages/modules/mongodb/src/mongodb-container.test.ts) inside_block:connectMongo
 <!--/codeinclude-->
 
 <!--codeinclude-->
-[Connect with credentials:](../../packages/modules/mongodb/src/mongodb-container.test.ts) inside_block:mongoConnectWithCredentials
+[Connect with credentials:](../../packages/modules/mongodb/src/mongodb-container.test.ts) inside_block:connectWithCredentials
 <!--/codeinclude-->
--- a/mkdocs.yml
+++ b/mkdocs.yml
@@ -29,7 +29,7 @@ markdown_extensions:
   - pymdownx.details
   - pymdownx.superfences
   - pymdownx.tabbed:
-      alternate_style: true
+      alternate_style: false
   - toc:
       permalink: true
 


packages/modules/mongodb/src/mongodb-container.test.ts
```diff
@@ -64,4 +64,47 @@ describe("MongodbContainer", { timeout: 240_000 }, () => {
     await mongodbContainer.stop();
   });
   // }
+  test.for([["mongo:4.0.1"], ["mongo:6.0.1"], ["mongo:8.0"]])(
+    "should connect to %s with credentials and disabled replica set",
+    async ([image]) => {
+      const mongodbContainer = await new MongoDBContainer(image)
+        .withUsername("mongo_user")
+        .withPassword("mongo_password")
+        .withReplicaSetEnabled(false)
```
cristianrgreco: Once we implement credentials support with replica sets, I guess we wouldn’t ever need to disable it. I’m therefore hesitant to add this to the API just to support credentials. WDYT, is it a lot more work to make it work with replica sets? Anything I can help with?

packages/modules/mongodb/src/mongodb-container.test.ts
```diff
@@ -64,4 +64,47 @@ describe("MongodbContainer", { timeout: 240_000 }, () => {
     await mongodbContainer.stop();
   });
   // }
+  test.for([["mongo:4.0.1"], ["mongo:6.0.1"], ["mongo:8.0"]])(
```
cristianrgreco: As this is a new feature, we only need to test it against the latest image

packages/modules/mongodb/src/mongodb-container.test.ts
digital88: Ok, I'll take some more approaches. Will try to make RS + auth possible. I will document what works and what is not.

packages/modules/mongodb/src/mongodb-container.test.ts
digital88: Thanks, I will use only 8.0 image.

digital88:
Also very interesting that 4.0 image fails tests with:
```
× MongodbContainer > should connect to mongo:4.0.1 with credentials and disabled replica set 127121ms (retry x3)
     → Server at localhost:41731 reports maximum wire version 7, but this version of the Node.js Driver requires at least 8 (MongoDB 4.2)
```
I do not encounter this on local machine. Need to investigate more.

digital88:
Ok, I decided to start from scratch here. I wrote a new `mongodb-container-2.ts` file with mongo container implementation. This version works for replica set both with and without auth. Mongo 5.0 and later is supported. Earlier images unfortunately don't work because of old mongo shell.
The main problem that I encountered was that mongo is doing some stuff in background after container is started and this causes eval commands to fail with `Authentication failed` errors. If I wait a little on breakpoint, tests suddenly start passing without errors. So I made basic retries logic in `executeMongoEvalCommand` and it actually solved this random errors I got earlier.
Edit: I'll try to port old code to this 'new' implementation.

digital88:
> Also very interesting that 4.0 image fails tests with:

Had to pin mongoose to `"8.15.0"`. Newer versions [do not support](https://mongoosejs.com/docs/compatibility.html) mongo 4.

I moved all logic in `containerStarted` to healthcheck and changed wait strategy to wait for healthcheck. Took inspiration from searching how to run mongo replica set with auth in compose. Not sure about default timeouts tho.

Credentials work with images from 4.0.1 up to 8.0

digital88:
Tests with podman fail healthcheck even in older cases without credentials. I think my timeouts may be the reason? I have only docker installed locally. Will try later to test in podman too.

digital88:
Changed `CMD` to `CMD-SHELL` as used in other modules. Tests pass in podman locally.

cristianrgreco:
@digital88 do you mind resolving the conflicts? I would but I’m afk at the moment. Should just be a testcontainers version bump

digital88:
> @digital88 do you mind resolving the conflicts? I would but I’m afk at the moment. Should just be a testcontainers version bump

Resolved

digital88:
> > @digital88 do you mind resolving the conflicts? I would but I’m afk at the moment. Should just be a testcontainers version bump
> 
> Resolved

And resolved again 😄

cristianrgreco:
Apologies @digital88, things are a bit busy at the moment, I know I need to review this PR, I will get to it soon.

cristianrgreco:
> Had to pin mongoose to "8.15.0". Newer versions [do not support](https://mongoosejs.com/docs/compatibility.html) mongo 4.

Let's not pin the mongoose version. It's pinned because of a Mongo version which was EOL Feb 2024, so when we need to update mongoose for some other reason we'll be stuck. As this is a new behaviour we don't need to test it with older Mongo versions as there are no existing users using it, so let's remove the credentials test for Mongo 4. The existing Mongo 4 test continues to pass so that's good enough from my perspective that this isn't a breaking change. If a user wants to use an older version of Mongoose so that it works with their Mongo 4 then great, but let's not tie ourselves to it.

packages/modules/mongodb/src/mongodb-container.ts
```diff
@@ -1,68 +1,85 @@
-import { AbstractStartedContainer, ExecResult, GenericContainer, StartedTestContainer, Wait } from "testcontainers";
+import { satisfies } from "compare-versions";
+import { AbstractStartedContainer, GenericContainer, StartedTestContainer, Wait } from "testcontainers";
 
 const MONGODB_PORT = 27017;
 
 export class MongoDBContainer extends GenericContainer {
+  private username = "";
+  private password = "";
```
cristianrgreco: Let's remove the default values and have the type `string | undefined`, so I don't have to double check `if (this.username && this.password)`  works as expected

packages/modules/mongodb/src/mongodb-container.ts
```diff
@@ -1,68 +1,85 @@
...
+    this.password = password;
+    return this;
   }
 
-  private async executeMongoEvalCommand(startedTestContainer: StartedTestContainer, command: string) {
-    const execResult = await startedTestContainer.exec(this.buildMongoEvalCommand(command));
-    this.checkMongoNodeExitCode(execResult);
+  public override async start(): Promise<StartedMongoDBContainer> {
+    const cmdArgs = ["--replSet", "rs0", "--bind_ip_all"];
+    this.withHealthCheck({
```
cristianrgreco: We can't set the healthcheck and wait strategy in the `start`, otherwise there's no way for a user to override it. We could follow a similar approach to the PostgreSQL container, and set the wait strategy to healthcheck in the constructor (so it can be overridden), and conditionally set the healthcheck _if_ it hasn't already been set: https://github.com/testcontainers/testcontainers-node/blob/c9ded058ccdfb21a2b9c5c77531f5693359f1c33/packages/modules/postgresql/src/postgresql-container.ts#L38-L48

packages/modules/mongodb/src/mongodb-container.ts
```diff
@@ -1,68 +1,85 @@
...
   }
 
   private buildMongoEvalCommand(command: string) {
-    return [this.getMongoCmdBasedOnImageTag(), "--eval", command];
+    const useMongosh = satisfies(this.imageName.tag, ">=5.0.0");
+    const args = [];
+    if (useMongosh) args.push("mongosh");
+    else args.push("mongo", "admin");
+    if (this.username && this.password) args.push("-u", this.username, "-p", this.password);
+    args.push("--host", "localhost", "--quiet", "--eval", command);
```
cristianrgreco: How comes `host` was set? If Mongo is able to deduce it itself I would opt for that

packages/modules/mongodb/src/mongodb-container.ts
```diff
@@ -1,68 +1,85 @@
...
+    if (password === "") throw new Error("Password should not be empty.");
+    this.password = password;
+    return this;
   }
 
-  private async executeMongoEvalCommand(startedTestContainer: StartedTestContainer, command: string) {
-    const execResult = await startedTestContainer.exec(this.buildMongoEvalCommand(command));
-    this.checkMongoNodeExitCode(execResult);
+  public override async start(): Promise<StartedMongoDBContainer> {
+    const cmdArgs = ["--replSet", "rs0", "--bind_ip_all"];
```
cristianrgreco: How comes `bind_ip_all` was set, could it be a breaking change for existing users?

digital88: Leftover when I was experimenting. `bind_ip_all` equals to "bind to 0.0.0.0" inside container. Removed.

packages/modules/mongodb/src/mongodb-container.ts
digital88: Leftover when I was experimenting. Removed.

packages/modules/mongodb/src/mongodb-container.ts
digital88: Fixed.

packages/modules/mongodb/src/mongodb-container.ts
digital88: Fixed

digital88:
> > Had to pin mongoose to "8.15.0". Newer versions [do not support](https://mongoosejs.com/docs/compatibility.html) mongo 4.
> 
> Let's not pin the mongoose version. It's pinned because of a Mongo version which was EOL Feb 2024, so when we need to update mongoose for some other reason we'll be stuck. As this is a new behaviour we don't need to test it with older Mongo versions as there are no existing users using it, so let's remove the credentials test for Mongo 4. The existing Mongo 4 test continues to pass so that's good enough from my perspective that this isn't a breaking change. If a user wants to use an older version of Mongoose so that it works with their Mongo 4 then great, but let's not tie ourselves to it.

I unpinned mongoose. Made tests with creds to run just on mongo 8.0

digital88:
Oh, sorry, TC patch version just went up again...Will fix it in a minute.

digital88:
Looks like I messed lockfile..I'll fix it.

packages/modules/mongodb/src/mongodb-container.test.ts
```diff
@@ -63,3 +63,27 @@ describe("MongodbContainer", { timeout: 240_000 }, () => {
   });
   // }
 });
+
+describe("MongodbContainer connect with credentials", { timeout: 240_000 }, () => {
+  it("should connect to %s with credentials", async () => {
```
cristianrgreco: This no longer needs to be in a describe, the `%s` removed, and the image `mongo:8.0` should be replaced with `IMAGE` (so it'll be auto-upgraded by Dependabot)

digital88: Ok. I see that test named `should work using default version 4.0.1` seems to use latest image 8.x.x too (It uses IMAGE variable)? Maybe we should hardcode mongo version to 4 for this test or rename this test?

cristianrgreco: You're right, we should have a test for Mongo 4, 6, and the latest. So need to add a new test. Regardless, this credentials test just needs to run against the latest (`IMAGE`)

cristianrgreco (approved):

cristianrgreco:
@digital88 Are there any more changes you'd like to make?

digital88:
> @digital88 Are there any more changes you'd like to make?

I think it's good 👍

cristianrgreco:
Thanks for the contribution, always a pleasure!

Issue #1071, opened 2025-07-12
Add testcontainer for OpenSearch


JohanObrink:
PR incoming 😸

cristianrgreco:
Released in 11.3.0

PR #1072, opened 2025-07-12
Add Opensearch module

JohanObrink:
Sorry I didn't read the CONTRIB guidelines before starting work on the PR but I hope you will accept it without a prior Issue.

I looked at how the Testcontainer for Elasticsearch was built and tried to maintain the same style. I also basically copied the tests almost verbatim.

One thing added to the container that differs from the Elasticsearch implementation is a validation of the password which, if set, has to adhere to the requirements of the OpenSearch demo container. I added a check for this in TS to make it fail faster and with a more readable error message.

Otherwise, the container implementation follows the settings of the Dotnet OpenSearch container.

Two tests fail - for Kafka and BigQuery. It did not seem right to touch those tests in this PR.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -39,6 +39,7 @@ updates:
       - "/packages/modules/nats"
       - "/packages/modules/neo4j"
       - "/packages/modules/ollama"
+      - "/packages/modules/opensearch"
       - "/packages/modules/postgresql"
       - "/packages/modules/qdrant"
       - "/packages/modules/rabbitmq"


JohanObrink:
Fixes #1071

cristianrgreco:
LGTM! Just missing an entry in dependabot.yml https://github.com/testcontainers/testcontainers-node/blob/3f3eed17742d4d48118959542571b3480811d138/.github/dependabot.yml

cristianrgreco:
Thanks @JohanObrink

PR #1076, opened 2025-07-14
Add Spanner module

JohanObrink:
Added SpannerEmulatorContainer. It returns a StartedSpannerEmulatorContainer which in turn exposes a .helper for convenience methods for instance/database creation and deletion.

This PR also adds .withPlatform for BigQueryEmulatorContainer to make it work on M-series Macs.


cristianrgreco:
There's an msw dependency override in the root package.json. If you remove the override and the tests pass then great, if not we need to keep the previous msw version. This is because of some issue with msw and Unix sockets which break the docker integration: https://github.com/mswjs/msw/issues/1600

packages/modules/gcloud/src/bigquery-emulator-container.ts
```diff
@@ -8,6 +8,12 @@ export class BigQueryEmulatorContainer extends GenericContainer {
   constructor(image: string) {
     super(image);
     this.withExposedPorts(EMULATOR_PORT).withWaitStrategy(Wait.forListeningPorts()).withStartupTimeout(120_000);
+
+    // The BigQuery emulator image is not multi platform
+    // so this fix is needed for M Macs
```
cristianrgreco:
```suggestion
    // The BigQuery emulator image is not multi platform
    // so this fix is needed for ARM architectures
```

packages/modules/gcloud/src/spanner-emulator-container.test.ts
```diff
@@ -0,0 +1,57 @@
...
+// select the fourth FROM in the Dockerfile (spanner emulator)
+const IMAGE = getImage(__dirname, 3);
+
+describe("SpannerEmulatorContainer", { timeout: 240_000 }, () => {
+  it("should start and expose endpoints", async () => {
+    // <example startup> {
+    const container = await new SpannerEmulatorContainer(IMAGE).withProjectId("test-project").start();
+    const grpcEndpoint = container.getEmulatorGrpcEndpoint();
+
+    expect(grpcEndpoint).toMatch(/localhost:\d+/);
```
cristianrgreco: This test will fail depending on where it's run (host is not always `localhost`, e.g. when running DIND). 

Instead of testing the URL, is it possible to test the behaviour instead - i.e. some integration test that we're able to talk to Spanner via GRPC?

packages/modules/gcloud/src/spanner-emulator-container.test.ts
```diff
@@ -0,0 +1,57 @@
...
+
+    await container.stop();
+    // }
+  });
+
+  it("should create and delete instance and database via helper", async () => {
```
cristianrgreco: Do you mind moving the helper and its tests to separate files? I think it'd make the separation between the container and its utilities a bit more clear

packages/modules/gcloud/src/spanner-emulator-container.ts
```diff
@@ -0,0 +1,179 @@
...
+  private databaseAdminClientInstance?: ReturnType<Spanner["getDatabaseAdminClient"]>;
+  private instanceConfigValue?: string;
+
+  constructor(private readonly emulator: StartedSpannerEmulatorContainer) {}
+
+  /**
+   * Must be called by the user to configure the SPANNER_EMULATOR_HOST env var.
+   */
+  public setAsEmulatorHost(): void {
+    process.env.SPANNER_EMULATOR_HOST = this.emulator.getEmulatorGrpcEndpoint();
```
cristianrgreco: Does it make sense to extend the `AbstractGcloudEmulator` class for the Spanner module? 

I see there they're setting the host via a flag: https://github.com/testcontainers/testcontainers-node/blob/4b7b9d0a5ee7f8482936cef8a457039ea616129c/packages/modules/gcloud/src/abstract-gcloud-emulator.ts#L15

If possible, this would be preferable to setting the env var, and then we don't need this setter, or the checks

JohanObrink: The Abstract base class is for using the google-cloud-cli:emulators image with different entry points. Spanner is on its way into that image but isn't there yet. So the EmulatorContainer uses its own, dedicated image, which means it extends Generic instead. Check f.e BigQueryEmulatorContainer for a similar solution.

The host flag is actually not enough. Setting the emulator env variable serves dual purpose. It points to the emulator _and_ tells the library it is connecting to an emulator, removing the requirement of credentials.

packages/modules/gcloud/src/spanner-emulator-container.test.ts
```diff
@@ -0,0 +1,33 @@
...
+describe("SpannerEmulatorContainer", { timeout: 240_000 }, () => {
+  it("should start, expose endpoints and accept real client connections", async () => {
+    // startup {
+    const container = await new SpannerEmulatorContainer(IMAGE).withProjectId("test-project").start();
+
+    // 1) get the endpoint
+    const grpcEndpoint = container.getEmulatorGrpcEndpoint();
+
+    // 2) configure the client to talk to our emulator
+    process.env.SPANNER_EMULATOR_HOST = grpcEndpoint;
```
cristianrgreco: I'm able to get the tests passing without needing to set this env var with the following:

```shell
npm install -w packages/modules/gcloud @grpc/grpc-js --save-dev
```

```ts
import { credentials } from "@grpc/grpc-js";

const client = new Spanner({
  projectId: container.getProjectId(),
  apiEndpoint: container.getHost(),
  port: container.getMappedPort(9010),
  sslCreds: credentials.createInsecure(),
});
```

JohanObrink: Yes that is correct. SPANNER_EMULATOR_HOST just simplifies things a bit. Especially when running production code locally (where you wouldn't want to explicitly set insecure credentials). So I just prefer to use it in testing as well. And that seems to be the standard recommendation for the GCloud emulators in general.

When connecting to PubSub, Firestore, Spanner etc from within GCP you generally just instantiate your client without passing any arguments and it is automatically wired up by environment variables provided by your GCP project.

cristianrgreco: The issue I have with modifying the environment is that it is a shared state, and Testcontainers has many features to support running tests in parallel. For example, if I duplicate the test in `spanner-emulator-container.test.ts` and run them in parallel, with some timings I can reproduce that the first test is actually talking to the second tests's container, because the second test overwrote the process.env.SPANNER_EMULATOR_HOST, and yet the tests pass. For a user, this would be very confusing, gives false confidence when tests pass, or will be a debugging headache when tests flake.

If a user wants to provide SSL credentials then we can add support for that via the container's methods. We do this in several other modules. We can add another helper method for providing insecure creds if needed. 

In terms of simplicity when configuring the client, I think setting the host and port isn't too much of an issue, especially when considering the parallelism issues of doing otherwise

JohanObrink: But that is why I explicitly set the variable in the test/example for the SpannerEmulatorContainer and only auto include it when you use the SpannerEmulatorHelper which encapsulates a metric tonne of Spanner weirdness anyway.

For me it seems stranger to recommend an implementation which flies in the face of what Google recommends themselves.

cristianrgreco: It seems I'm missing something, so please just correct me where I'm wrong

> But that is why I explicitly set the variable in the test/example for the SpannerEmulatorContainer and only auto include it when you use the SpannerEmulatorHelper

The test doesn't use the helper, it uses just the container and the official client from @google-cloud/spanner, so I don't understand what you meant there. The test fails if the env var isn't set.

This test is the "front-face" of the module's documentation, so if users copy/paste it, and start to grow their test suite, they will run into the parallelisation issues I mentioned - that is not right.

> For me it seems stranger to recommend an implementation which flies in the face of what Google recommends themselves.

Could you please link me to this recommendation? 

I mean, their client supports setting the `apiEndpoint` and `port` in the constructor, so saying using it is "flying in their face" is a bit much. Even regardless if this is their recommendation, then their recommendation 100% does not work for use cases where there is more than one Spanner container, which is a use case Testcontainers does support. I guess it is why it is just a recommendation.

JohanObrink: This is the documentation I followed when first using the emulator:

https://cloud.google.com/spanner/docs/emulator#client-libraries

Step 1 is setting the environment variable.

The whole pattern of running stuff in GCP (as far as I have been able to figure out at least ;)) is that you run everything inside your project and, as long as you do, any managed service just works without mucking about with credentials. You set up your service permissions on an infra level and the code just works.

This means that running locally requires your environment to work as expected in GCP. That is why I built my first example with the container being started with an explicit projectId (it will always be test-project if you run any emulator without specific configuration) and setting the environment variable. The helper does the setup for you but at least now you've seen it.

There will never be a situation where you need more than one emulator running. For most regular setups, more than one instance would be unthinkable (if nothing else then for cost reasons). Connecting cross projects seems just insane.

When it comes to tests, just starting the db is costly enough that you will always want to place it in a beforeAll together with your instance and database creation (in prod these will not be defined in app code but in infra code - usually Terraform). Your individual tests (including beforeEach:es) will create/drop tables and clear rows but that's about it.

Actually, the way we run it, we have our own homegrown migration framework so the tests a structured as:

beforeAll: start container, create instance, create databases, run migrations up

beforeEach: truncate tables

afterAll: run migrations down, remove databases, remove instance

So I really don't see the problem of the tests assuming one emulator with one global setting. But maybe I am missing something?

cristianrgreco: > For most regular setups, more than one instance would be unthinkable (if nothing else then for cost reasons). Connecting cross projects seems just insane.

We're not talking about a production deployment (where I would agree with you), we're talking about how users use Testcontainers in their tests. We have a lot of users who _expect_ to be able to run their tests in parallel. This would be the first module where we knowingly don't allow for it.

> When it comes to tests, just starting the db is costly enough that you will always want to place it in a beforeAll together with your instance and database creation 

This is not necessarily true. I have an M1 mac. If I run the container test it completes in 5 seconds. If I run the container test 5 times in parallel, it also completes in 5 seconds (with 5 containers). For this reason many users skip the global setup and overhead of trying to manage state.

> Your individual tests (including beforeEach:es) will create/drop tables and clear rows but that's about it.

Creating a single container in a `beforeAll` and trying to manage its state by creating/dropping tables/clearing rows in between tests is another way to inhibit running tests in parallel. As well as a container for each test, you could imagine a scenario where we have a container per test suite, where the test suites run in parallel but the tests within run sequentially, which is another common use case as it's much easier to deal with the container state, if it's even necessary.

---

How about a middle ground. 

Let's make the default behaviour as I suggest. So by default everything just works out of the box. Users don't even have to read up about Google's recommendations, they can just follow the same simple patterns as all other modules, and can also run their tests in parallel.

If a user wants to set the host/port via their environment, let's also support that.

So something like:

```ts
const client = process.env.SPANNER_EMULATOR_HOST
  ? new Spanner({ projectId: container.getProjectId() });
  : new Spanner({
      projectId: container.getProjectId(),
      apiEndpoint: container.getHost(),
      port: container.getGrpcPort(),
      sslCreds: container.getSslCredentials(),
    });
```

WDYT?

JohanObrink: I see what you mean 👍🏻

I have absolutely no problem with adding two examples - one with environment variable set and one with explicit credentials for parallell execution 😊

But then I also want to make the helper use the second method - otherwise it will be unusable for parallell execution _and_ behave unpredictably.

I'll update the code. Thank you for your patience!

packages/modules/gcloud/src/spanner-emulator-container.test.ts
```diff
@@ -28,6 +28,36 @@ describe("SpannerEmulatorContainer", { timeout: 240_000 }, () => {
...
+
+  // startupWithExplicitClient {
+  it("should start, expose endpoints and accept real client connections using explicitly configured client", async () => {
+    const container = await new SpannerEmulatorContainer(IMAGE).withProjectId("test-project").start();
+
+    // 1) get the endpoint
+    const grpcEndpoint = container.getEmulatorGrpcEndpoint();
+
+    // 2) configure the client to talk to our emulator
+    process.env.SPANNER_EMULATOR_HOST = grpcEndpoint;
```
cristianrgreco: Can be removed

packages/modules/gcloud/src/spanner-emulator-container.test.ts
```diff
@@ -6,8 +6,8 @@ import { SpannerEmulatorContainer } from "./spanner-emulator-container";
 const IMAGE = getImage(__dirname, 3);
 
 describe("SpannerEmulatorContainer", { timeout: 240_000 }, () => {
```
cristianrgreco, line 8: This comment is assuming https://github.com/testcontainers/testcontainers-node/pull/1076/commits/e0b030dfe6e75c71abd484ef63b9ec56ea6b67e0#r2213512361 is complete.

---

It looks like even though for the explicit client test we set the api endpoint and port, _if_ `process.env.SPANNER_EMULATOR_HOST` is set, it _still_ takes precedence. 

E.g. I can set `process.env.SPANNER_EMULATOR_HOST = "lol"` and the test fails.

For this reason we cannot run the test suite concurrently. I suggest we create a nested `describe.sequential` block where we can put the tests which rely on shared state. That way all the tests can run concurrently, except for those.

docs/modules/gcloud.md
```diff
@@ -68,7 +68,11 @@ The BigQuery emulator is by [Masaaki Goshima](https://github.com/goccy) and uses
 The Cloud Spanner emulator container wraps Google's official emulator image. It exposes gRPC and HTTP ports, and provides a `helper` for instance/database operations.
 
 <!--codeinclude-->
-[Starting a Spanner Emulator container and exposing endpoints](../../packages/modules/gcloud/src/spanner-emulator-container.test.ts) inside_block:startup
+[Starting a Spanner Emulator container and exposing endpoints using projectId and SPANNER_EMULATOR_HOST](../../packages/modules/gcloud/src/spanner-emulator-container.test.ts) inside_block:startupWithEnvironmentVariable
```
cristianrgreco: Please swap the order of the examples around, so parallelism works for the first example users see.

packages/modules/gcloud/src/spanner-emulator-container.test.ts
JohanObrink: Yup... That was just sloppy of me 😞 Fixed it now

cristianrgreco (approved):

cristianrgreco:
Thanks for the contribution @JohanObrink! Is there anything you'd like to change or are we good to merge?

JohanObrink:
I think we're good 👍🏻 Thanks for all feedback and suggestions 😊

PR #1079, opened 2025-07-16
Implement HashiCorp Vault module

inf1nite-lo0p:


packages/modules/vault/src/vault-container.test.ts
```diff
@@ -0,0 +1,54 @@
...
+import { StartedVaultContainer, VaultContainer } from "./vault-container";
+
+const VAULT_TOKEN = "my-root-token";
+
+describe("vault", { timeout: 180_000 }, () => {
+  let container: StartedVaultContainer;
+
+  afterEach(async () => {
+    await container?.stop();
+  });
```
cristianrgreco: The tests are passing but there is a race condition. The tests run concurrently by default, so it's possible one test will stop the other's container (and likely other tests' containers aren't being stopped). You can reproduce this by adding more tests, or wrapping a test in a for loop, you'll see it starts to fail.

In the future we'll be able to use TS 5.2's `using` to get rid of this boilerplate, but for now we need it.

docs/modules/vault.md
```diff
@@ -0,0 +1,33 @@
...
+## Install
+
+```bash
+npm install @testcontainers/vault --save-dev
+```
+
+## Examples
+
+<!--codeinclude-->
+
```
cristianrgreco: I wonder if these empty lines are the reason there is whitespace between each line of code: https://deploy-preview-1079--testcontainers-node.netlify.app/modules/vault/. 

Could we try removing them?

packages/modules/vault/src/vault-container.test.ts
```diff
@@ -0,0 +1,54 @@
+import vault from "node-vault";
+import { setTimeout } from "node:timers/promises";
+import { StartedVaultContainer, VaultContainer } from "./vault-container";
+
+const VAULT_TOKEN = "my-root-token";
+
+describe("vault", { timeout: 180_000 }, () => {
```
cristianrgreco:
```suggestion
describe("VaultContainer", { timeout: 180_000 }, () => {
```

packages/modules/vault/src/vault-container.test.ts
```diff
@@ -0,0 +1,54 @@
...
+    });
+
+    await client.write("secret/data/hello", {
+      data: {
+        message: "world",
+        other: "vault",
+      },
+    });
+
+    await setTimeout(1000);
```
cristianrgreco: Can be removed

packages/modules/vault/src/vault-container.ts
```diff
@@ -0,0 +1,129 @@
+import { AbstractStartedContainer, GenericContainer, Wait } from "testcontainers";
+
+const VAULT_PORT = 8200;
+
+/**
+ * Testcontainers module for HashiCorp Vault.
+ *
+ * Based on the upstream image `hashicorp/vault:1.13.0`, this container exposes Vault
```
cristianrgreco: Let's not hardcode the image as it gets bumped automatically by Dependabot

packages/modules/vault/src/vault-container.ts
```diff
@@ -0,0 +1,129 @@
...
+   * Constructs a VaultContainer with a default image and healthcheck strategy.
+   *
+   * - Sets VAULT_ADDR to internal container address
+   * - Adds IPC_LOCK capability (required by Vault)
+   * - Exposes Vault on port 8200
+   * - Waits for HTTP 200 response from /v1/sys/health
+   *
+   * @param image Optional Docker image to use (default: hashicorp/vault:1.13.0)
+   */
+  constructor(image: string = "hashicorp/vault:1.13.0") {
```
cristianrgreco: Remove the default value, otherwise if we want to change it, it would require a major release (as it could break everyone's tests who are relying on the defaults).

packages/modules/vault/src/vault-container.ts
```diff
@@ -0,0 +1,129 @@
...
+   * This is typically used to pre-configure secret engines or seed test data.
+   *
+   * @param commands Array of CLI commands (without `vault` prefix)
+   */
+  public async execVaultCommands(commands: string[]): Promise<void> {
+    const cmd = commands.map((c) => `vault ${c}`).join(" && ");
+    const result = await this.exec(["/bin/sh", "-c", cmd]);
+
+    if (result.exitCode !== 0) {
+      throw new Error(`Vault init commands failed: ${result.output}`);
```
cristianrgreco: The method is called `execVaultCommands` but the log says `Vault init commands`, could we make them consistent?

inf1nite-lo0p:
Thanks for putting the time into reviewing this.

I'll address the remarks, and ask for another review.

docs/modules/vault.md
inf1nite-lo0p: Yes, that was the cause. Great catch!

https://github.com/testcontainers/testcontainers-node/pull/1079/commits/85c8091f8e68ffbab5459a376f357b3a1d92987b

packages/modules/vault/src/vault-container.test.ts
inf1nite-lo0p: https://github.com/testcontainers/testcontainers-node/pull/1079/commits/6a04ce8540fea4a36df3df42ace31660a8e7da04

packages/modules/vault/src/vault-container.test.ts
inf1nite-lo0p: https://github.com/testcontainers/testcontainers-node/pull/1079/commits/94037565a65b84f5c45ee03fab16aac2ab4f2217

packages/modules/vault/src/vault-container.ts
inf1nite-lo0p: https://github.com/testcontainers/testcontainers-node/pull/1079/commits/7a1579aaf549725053e83a56a65cc4fbb878232f

packages/modules/vault/src/vault-container.ts
inf1nite-lo0p: https://github.com/testcontainers/testcontainers-node/pull/1079/commits/7a1579aaf549725053e83a56a65cc4fbb878232f

packages/modules/vault/src/vault-container.ts
inf1nite-lo0p: https://github.com/testcontainers/testcontainers-node/pull/1079/commits/2f40855af51b8c733a622209bee8d236b039f683

packages/modules/vault/src/vault-container.test.ts
inf1nite-lo0p: Thanks for the heads up!

https://github.com/testcontainers/testcontainers-node/pull/1079/commits/05211d132ea2d40fb8493dca0ea06517ca257c97

cristianrgreco:
Thanks, @inf1nite-lo0p!

PR #1080, opened 2025-07-17
Implement `AsyncDisposable` to support TS 5.2's `using`

cristianrgreco:
Feature info here: https://www.typescriptlang.org/docs/handbook/release-notes/typescript-5-2.html#using-declarations-and-explicit-resource-management

Before:

```ts
const container = await new GenericContainer(/* ... */).start();
/* ... */
await container.stop();
```

After:

```ts
await using container = await new GenericContainer(/* ... */).start();
/* ... */
```

Same for `Network` and `DockerComposeEnvironment`.

### DIFF
--- a/packages/modules/gcloud/src/spanner-emulator-container.test.ts
+++ b/packages/modules/gcloud/src/spanner-emulator-container.test.ts
@@ -8,7 +8,7 @@ const IMAGE = getImage(__dirname, 3);
 describe("SpannerEmulatorContainer", { timeout: 240_000 }, () => {
   // startupWithExplicitClient {
   it("should start, expose endpoints and accept real client connections using explicitly configured client", async () => {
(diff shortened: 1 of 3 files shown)


packages/testcontainers/src/generic-container/abstract-started-container.ts
```diff
@@ -98,4 +98,8 @@ export class AbstractStartedContainer implements StartedTestContainer {
   public logs(opts?: { since?: number; tail?: number }): Promise<Readable> {
     return this.startedTestContainer.logs(opts);
   }
+
+  async [Symbol.asyncDispose]() {
+    await this.startedTestContainer[Symbol.asyncDispose]();
+  }
```
joebowbeer, line 104: @cristianrgreco Shouldn't this call `this.stop`?

```suggestion
  async [Symbol.asyncDispose]() {
    await this.stop();
  }
```

PR #1096, opened 2025-08-03
Use configured health checks as the default wait strategy

### LINKED ISSUE #687: Default to HealthCheckWaitStrategy for containers with health checks defined
**Expected Behaviour**

For containers or compose services with health checks defined, the default wait strategy should be the `HealthCheckWaitStrategy`.

This would be particularly useful for compose projects where many services can be defined and health checks can be defined directly in the compose config.

**Actual Behaviour**

The default wait strategy is the `HostPortWaitStrategy` which is not always applicable. While we can override the health check manually, it would be great if `testcontainers` could automatically use the best strategy.

**Testcontainer Logs**
...

**Environment Information**
- Operating System: Ubuntu 22.04
- Docker Version: 24.0.7
- Node version: 18.18.2
- Testcontainers version: 10.3.2

digital88:
Closes #687 

Had to augment ImageInspectInfo & ContainerInspectInfo because dockerode types do not expose HealthCheck nested property in Config property. I opened PR [here](https://github.com/DefinitelyTyped/DefinitelyTyped/pull/73357) but not sure when it will be accepted.


packages/testcontainers/src/docker-compose-environment/docker-compose-environment.ts
```diff
@@ -23,7 +24,7 @@ export class DockerComposeEnvironment {
   private profiles: string[] = [];
   private environment: Environment = {};
   private pullPolicy: ImagePullPolicy = PullPolicy.defaultPolicy();
-  private defaultWaitStrategy: WaitStrategy = Wait.forListeningPorts();
+  private defaultWaitStrategy: WaitStrategy = new NullWaitStrategy();
```
cristianrgreco: Looks like this `NullWaitStrategy` can be removed by setting the type of the `defaultWaitStrategy` to `WaitStrategy | undefined`, and changing L214 to `if (containerWaitStrategy return containerWaitStrategy;`

packages/testcontainers/src/docker-compose-environment/docker-compose-environment.ts
```diff
@@ -207,4 +206,22 @@ export class DockerComposeEnvironment {
...
+  private selectWaitStrategy(containerName: string, inspectResult: ContainerInspectInfo) {
+    const containerWaitStrategy = this.waitStrategy[containerName]
+      ? this.waitStrategy[containerName]
+      : this.defaultWaitStrategy;
+    if (!(containerWaitStrategy instanceof NullWaitStrategy)) return containerWaitStrategy;
+    const healthcheck = (
+      inspectResult as ContainerInspectInfo & {
+        Config: ContainerInspectInfo["Config"] & {
+          Healthcheck: HealthConfig | undefined;
+        };
```
cristianrgreco: Would you be able to raise a PR for `@types/dockerode` to get this added? I've raised a few in the past it only takes a few mins. I'm uncomfortable depending on a custom type for something as critical as the wait strategy. It'd also simplify this code

packages/testcontainers/src/generic-container/abstract-started-container.ts
```diff
@@ -76,6 +76,10 @@ export class AbstractStartedContainer implements StartedTestContainer {
     return this.startedTestContainer.getIpAddress(networkName);
   }
 
+  public getWaitStrategy() {
```
cristianrgreco: This is added only for test purposes, is it possible to verify the new behaviour without exposing the underlying wait strategy?

packages/testcontainers/src/docker-compose-environment/docker-compose-environment.ts
cristianrgreco: Looking at the build, it seems possible that Podman doesn't expose this same information

digital88: > Would you be able to raise a PR for `@types/dockerode` to get this added? I've raised a few in the past it only takes a few mins. I'm uncomfortable depending on a custom type for something as critical as the wait strategy. It'd also simplify this code

I did update my PR in DefinitelyTyped repo to include `ContainerInspectInfo.Config.Healthcheck` too like in `ImageInspectInfo`. It's been two weeks without review, people are on vacation probably 😄

digital88: > Looking at the build, it seems possible that Podman doesn't expose this same information

What should we do here? Run tests only for docker? The code will fallback to default `HostPortWaitStrategy` in case of podman running testcontainers.
Edit: I run tests for docker only.

packages/testcontainers/src/generic-container/abstract-started-container.ts
digital88: Yes, fixed

packages/testcontainers/src/docker-compose-environment/docker-compose-environment.ts
digital88: Fixed

packages/testcontainers/src/docker-compose-environment/docker-compose-environment.ts
cristianrgreco: We need to find a way to support this for podman as well. Possibly the health check info is somewhere else in the inspect result. Would need to try with podman and see

digital88: > We need to find a way to support this for podman as well. Possibly the health check info is somewhere else in the inspect result. Would need to try with podman and see

I did not deep dive into it yet, but Podman [exposes](https://docs.podman.io/en/latest/_static/api.html#tag/images-(compat)/operation/ImageInspect) the same API as Docker.

kiview: Looking at the [container inspect API](https://docs.podman.io/en/latest/_static/api.html#tag/containers/operation/ContainerInitLibpod), the `Healthcheck` config response seems to differ from [Docker](https://docs.docker.com/reference/api/engine/version/v1.45/#tag/Container/operation/ContainerInspect).

digital88: > Looking at the [container inspect API](https://docs.podman.io/en/latest/_static/api.html#tag/containers/operation/ContainerInitLibpod), the `Healthcheck` config response seems to differ from [Docker](https://docs.docker.com/reference/api/engine/version/v1.45/#tag/Container/operation/ContainerInspect).

There are [another set of (compat)](https://docs.podman.io/en/latest/_static/api.html#tag/containers-(compat)/operation/ContainerInspect) API which is mirroring Docker API.

kiview: It still looks different in the JSON response example though. But maybe that is just a docs issue.

digital88: So at least my PR in DT repo got merged. I'm going to remove this hardcoded custom types a bit later.

digital88: I can't make podman inspect to return Healthcheck..It's probably podman issue. I'm using v5.6.0.
Here is Dockerfile, this one (except small modification in HEALTHCHECK command): `packages/testcontainers/fixtures/docker/docker-with-health-check/Dockerfile`
I build it like this: `podman build -t 1:latest .`
image
image
image

digital88: More specifically this is probably podman API problem..
`podman inspect localhost/1` reports Healthcheck in resulting json.

cristianrgreco: Looks like we can get the healthcheck config from the container instead of the image. We can get it before the container is started. For example right after here https://github.com/testcontainers/testcontainers-node/blob/40829ec740a9aa298a65038cdc8309d3535b807d/packages/testcontainers/src/generic-container/generic-container.ts#L170

digital88: Yes, this seems to work. Thanks for the hint 😄

digital88:
I fixed some invalid WaitStrategy references in kafka and red panda. Tests pass locally

digital88:
@cristianrgreco 
Hi, is there something else that should be done in this PR? Or we are just waiting for next major release before merging this?

cristianrgreco:
> @cristianrgreco Hi, is there something else that should be done in this PR? Or we are just waiting for next major release before merging this?

Hey @digital88, apologies for the delay, been super busy with work lately! I still need to review this PR, but yes you're right it'll eventually get queued for a major release.

cristianrgreco:
I've had another look at this PR (need to release a major version for dropping node 20, and wanted to add this as part of the release) and found the following issues:

- Reused containers bypass the new selector. In [generic-container.ts:164-170](https://github.com/testcontainers/testcontainers-node/blob/27f20fb3cb08721a312af31cfa5115b8eb5284d7/packages/testcontainers/src/generic-container/generic-container.ts#L164-L170), the reuse path falls back directly to Wait.forListeningPorts() and only applies withStartupTimeout() to this.waitStrategy. Existing reusable containers with image/custom health checks will still wait on ports, and fallback strategies lose the configured startup timeout. Resolve the wait strategy from the reused container inspect result, then apply the timeout to that resolved strategy.

- Disabled health checks are treated as enabled. [generic-container.ts:126](https://github.com/testcontainers/testcontainers-node/blob/27f20fb3cb08721a312af31cfa5115b8eb5284d7/packages/testcontainers/src/generic-container/generic-container.ts#L126) and [docker-compose-environment.ts:214-216](https://github.com/testcontainers/testcontainers-node/blob/27f20fb3cb08721a312af31cfa5115b8eb5284d7/packages/testcontainers/src/docker-compose-environment/docker-compose-environment.ts#L214-L216) check only whether Healthcheck.Test exists. HEALTHCHECK NONE / Compose test: ["NONE"] disables health checks, so this would select Wait.forHealthCheck() for a container that never gets State.Health, causing a timeout instead of the port wait fallback. Docker and Compose both document this disable form: [Dockerfile reference](https://docs.docker.com/reference/dockerfile/#healthcheck), [Compose spec](https://github.com/compose-spec/compose-spec/blob/main/spec.md#healthcheck).

- Docs still describe the old default. [docs/features/wait-strategies.md:15](https://github.com/testcontainers/testcontainers-node/blob/27f20fb3cb08721a312af31cfa5115b8eb5284d7/docs/features/wait-strategies.md#L15) and [docs/features/compose.md:46](https://github.com/testcontainers/testcontainers-node/blob/27f20fb3cb08721a312af31cfa5115b8eb5284d7/docs/features/compose.md#L46) still say listening ports is always the default. This PR changes public behavior, so the docs should explain health-check defaulting and explicit override behavior.

@digital88 WDYT? I notice this PR hasn't gained much traction so not sure how important it is to include. I appreciate that this PR is old (my fault) but if we can fix it up we can probably include it. I'll give it a go.

cristianrgreco:
@digital88 LMK if you want to review the changes, if not I reckon we're good to merge.

PR #1109, opened 2025-08-15
Feat/container manager destory all containers within single command

zYasser:
Hi,
This first open source contribution, let's talk about this feature:
This feature lets users shut down all test containers with a single command, making it convenient to clean up after reusing containers across tests, especially once all test suites have finished running.

### DIFF
--- a/packages/testcontainers/src/enum.ts
+++ b/packages/testcontainers/src/enum.ts
@@ -1,5 +1,0 @@
-export enum Signal {
-  KILL = "KILL",
-  ADD = "ADD",
-  REMOVE = "REMOVE",
-}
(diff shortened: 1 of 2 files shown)


kiview:
Hi @zYasser, thanks for your first OSS contribution 🙌 

Given you are saying:
>  making it convenient to clean up after reusing containers across tests, especially once all test suites have finished running.

That's what Ryuk is already for, since it will remove all resources once the Testcontainers process ends. Do you really need to have a manual mechanism to clean this up?
And in case we do, also in that case Ryuk can be used instead of creating new concepts within the code base.

zYasser:
> That's what Ryuk is already for, since it will remove all resources once the Testcontainers process ends. Do you really need to have a manual mechanism to clean this up?

sorry, i didn't know about this for some reason when i create a new container with .withReuse() after all the test is complete for some reason the containers is still alive

kiview:
Yes, that is the purpose of `withReuse`, to have containers survive the process lifetime.
In Java, we use the [container singleton pattern](https://java.testcontainers.org/test_framework_integration/manual_lifecycle_control/#singleton-containers) to share containers across test classes (which will still respect Ryuk for overall cleanup), not sure how this pattern can be best implemented in Node.js.

I am sure @cristianrgreco can help (we should probably also add this to the docs, if we have something similar in Node.js).

PR #1117, opened 2025-08-22
Add username option to Valkey container

rqbazan:
## Description

Hey there! ✋ 

I made a small change to the Valkey test container to allow passing a `username` option. Right now, my workaround is to create a generic container like this:

```ts
const container = new GenericContainer(image)
  .withName(name)
  .withStartupTimeout(120_000)
  .withWaitStrategy(Wait.forLogMessage('Ready to accept connections'))
  .withExposedPorts({
    container: 6379,
    host: port,
  })
  .withCommand([
    'valkey-server',
    ...(password ? [`--requirepass "${password}"`] : []),
    ...(username ? [`--user "${username}" on >${password} ~* +@all`] : [])
  ]);
```

But would be nice if `@testcontainers/valkey` support this natively

## Changes

- Add `withUsername` method to ValkeyContainer
- Conditional apply `--user` if username was provided


packages/modules/valkey/src/valkey-container.test.ts
```diff
@@ -33,6 +33,18 @@ describe("ValkeyContainer", { timeout: 240_000 }, () => {
     client.destroy();
   });
 
+  it("should connect with username & password and execute set-get", async () => {
```
cristianrgreco: Do we need this test if we have `should start with username and password`?

rqbazan: I don't think so, will be deleted in my next commit

rqbazan:
Hey @cristianrgreco ✋ I was wondering if you have any plans for the next release, as I’d love to use the changes from my recent contribution.

cristianrgreco:
Hey @rqbazan, apologies for the delay I've been away for the last week. I'm going to cut a release today

PR #1130, opened 2025-09-09
Add InfluxDB module

mdodsworth:
Tried to follow the java implementation on features and version support.


cristianrgreco:
Thanks @mdodsworth, could you please also add some docs? Please follow the template from other modules
- https://github.com/testcontainers/testcontainers-node/blob/main/mkdocs.yml#L55
- https://github.com/testcontainers/testcontainers-node/tree/main/docs/modules

packages/modules/influxdb/package.json
```diff
@@ -0,0 +1,40 @@
...
+  "publishConfig": {
+    "access": "public"
+  },
+  "scripts": {
+    "prepack": "shx cp ../../../README.md . && shx cp ../../../LICENSE .",
+    "build": "tsc --project tsconfig.build.json"
+  },
+  "devDependencies": {
+    "@influxdata/influxdb-client": "^1.35.0",
+    "axios": "^1.7.9"
```
cristianrgreco, line 35: We now use node's built-in HTTP client in the core and other modules, let's do the same here so we don't add a lot of unnecessary dependencies

packages/modules/influxdb/src/influxdb-container.test.ts
```diff
@@ -0,0 +1,202 @@
...
+
+describe("InfluxDBContainer", { timeout: 240_000 }, () => {
+  describe("InfluxDB 2.x", { timeout: 240_000 }, () => {
+    let container: StartedInfluxDBContainer;
+
+    afterEach(async () => {
+      if (container) {
+        await container.stop();
+      }
+    });
```
cristianrgreco, line 15: Let's use Typescript's `await using` syntax so the container is automatically stopped after each test

packages/modules/influxdb/src/influxdb-container.ts
```diff
@@ -0,0 +1,242 @@
...
+  private admin = "admin";
+  private adminPassword = "password";
+  private database?: string;
+
+  private isVersion2: boolean = true;
+
+  constructor(image: string) {
+    super(image);
+    // Determine version from image tag (default to v2 if unknown)
+    this.isVersion2 = this.isInfluxDB2(this.imageName.tag ?? "latest");
```
cristianrgreco, line 25: The `?? "latest"` is not needed, the `ImageName` constructor will automatically set the tag to `latest` if one is not provided

packages/modules/influxdb/src/influxdb-container.ts
```diff
@@ -0,0 +1,242 @@
...
+  }
+
+  public withDatabase(database: string): this {
+    this.database = database;
+    return this;
+  }
+
+  public override async start(): Promise<StartedInfluxDBContainer> {
+    // Re-evaluate version based on the final image tag
+    this.isVersion2 = this.isInfluxDB2(this.imageName.tag ?? "latest");
```
cristianrgreco, line 92: Why does it need to be re-evaluated?

packages/modules/influxdb/src/influxdb-container.ts
```diff
@@ -0,0 +1,242 @@
...
+    if (this.database) {
+      env.INFLUXDB_DB = this.database;
+    }
+
+    this.withEnvironment(env);
+  }
+
+  private isInfluxDB2(tag: string): boolean {
+    if (tag === "latest") {
+      return true; // Assume latest is v2
```
cristianrgreco, line 149: Does this module work for InfluxDB v3? If so I'd change the naming of this logic to be either `v1` or `> v1`, instead of `v1` or `v2` or `...`

packages/modules/influxdb/src/influxdb-container.ts
```diff
@@ -0,0 +1,242 @@
...
+    }
+
+    // Parse version number
+    const versionMatch = tag.match(/^(\d+)(?:\.(\d+))?/);
+    if (versionMatch) {
+      const majorVersion = parseInt(versionMatch[1], 10);
+      return majorVersion >= 2;
+    }
```
cristianrgreco, line 157: Let's use the `compare-versions` library which we use in other modules for this, e.g. https://github.com/testcontainers/testcontainers-node/blob/a7d4ca4c78b5f5c05693285eb1dba18cf5d9a669/packages/modules/kafka/src/kafka-container.ts#L180

packages/modules/influxdb/src/influxdb-container.ts
```diff
@@ -0,0 +1,242 @@
...
+        bucket: this.bucket,
+      });
+
+      if (this.adminToken) {
+        params.append("token", this.adminToken);
+      }
+
+      return `${this.getUrl()}?${params.toString()}`;
+    } else {
+      // InfluxDB 1.x connection string format
```
cristianrgreco, line 232: I wonder whether it would be better to have an `InfluxDBContainer` interface, and have v1 and vX implementations instead of almost every method having an if/else. Perhaps for the future, maybe if v3 also has incompatibilities

packages/modules/influxdb/src/influxdb-container.ts
mdodsworth: I ran a quick check and it looks like compare-versions doesn't handle non-semver tags like `1.8-alpine` (it throws).

```

describe("compare-versions non-semver handling", () => {
  it("should throw when provided a non-semver InfluxDB 1.x tag", () => {
    expect(() => satisfies("1.8-alpine", ">=2.0.0")).toThrow(/Invalid argument/);
  });
});
```

packages/modules/influxdb/src/influxdb-container.ts
mdodsworth: thanks. Will remove

packages/modules/influxdb/src/influxdb-container.ts
mdodsworth: @cristianrgreco giving this a bump.

cristianrgreco: Sorry @mdodsworth I didn't realise this was blocking the other changes. 

Good catch! Then yes please the implementation as it is

Issue #1150, opened 2025-09-29
RedisContainer does not accept arbitrary setup commands


danilofuchs:
**Expected Behaviour**

Running tests with Redis, we want to run the DB in-memory to save time and resources.

Using `withCommand` does not apply the commands correctly

```
const redis = await new RedisContainer("redis:7")
    .withCommand(["redis-server", "--save", "0", "--appendonly", "no"])
    .start()
```
**Actual Behaviour**

The container always runs with the default arguments:

https://github.com/testcontainers/testcontainers-node/blob/e5f28f88a5c190503d1615b84e7f894a0254618e/packages/modules/redis/src/redis-container.ts#L51

**Testcontainer Logs**
...

**Steps to Reproduce**
Create a test container with arguments:
```
const redis = await new RedisContainer("redis:7")
    .withCommand(["redis-server", "--save", "0", "--appendonly", "no"])
    .start()
```

Log the actual arguments in the redis-container file

**Environment Information**
- Operating System: Pop_OS 22.04
- Docker Version: 28.4.0
- Node version: 22.18.0
- Testcontainers version: 11.5.1

cristianrgreco:
Agreed. Looks like `redisArgs` should include existing command arguments. PR welcome

cristianrgreco:
Released in https://github.com/testcontainers/testcontainers-node/releases/tag/v11.8.1

PR #1155, opened 2025-10-03
Fix Redis container to accept command override

ziggornif:
Hi,

Here's a small fix to resolve the Redis container startup issue with custom commands (#1150).

I noticed a similar issue with the redis-stack image that also overrides `this.createOpts.Env`. 
Would you prefer I include this fix in the same PR, or should I create a separate issue to track it first?

### DIFF
--- a/packages/modules/redis/src/redis-container.test.ts
+++ b/packages/modules/redis/src/redis-container.test.ts
@@ -122,7 +122,7 @@ describe("RedisContainer", { timeout: 240_000 }, () => {
 
   it("should start redis with custom command", async () => {
     const container = new RedisContainer(IMAGE).withCommand(["redis-server", "--loglevel", "verbose"]);
-    const startedContainer = await container.start();
+    await using startedContainer = await container.start();
 
     // @ts-expect-error - accessing private property for testing
     expect(container.createOpts.Cmd).toEqual(["redis-server", "--loglevel", "verbose"]);
@@ -138,7 +138,7 @@ describe("RedisContainer", { timeout: 240_000 }, () => {
 
   it("should start redis-stack with custom env", async () => {
     const container = new RedisContainer(REDISSTACK_IMAGE).withEnvironment({ REDIS_ARGS: "--loglevel verbose" });
-    const startedContainer = await container.start();
+    await using startedContainer = await container.start();
 
     // @ts-expect-error - accessing private property for testing
     expect(container.createOpts.Env).toEqual(["REDIS_ARGS=--loglevel verbose"]);
@@ -158,7 +158,7 @@ describe("RedisContainer", { timeout: 240_000 }, () => {
     const container = new RedisContainer(IMAGE)
       .withCommand(["redis-server", "--loglevel", "verbose"])
       .withPersistence(sourcePath);
-    const startedContainer = await container.start();
+    await using startedContainer = await container.start();
 
     // @ts-expect-error - accessing private property for testing
     expect(container.createOpts.Cmd).toEqual([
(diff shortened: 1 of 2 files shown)


cristianrgreco:
Hi @ziggornif, yes please include the redis-stack command change in this PR

ziggornif:
Hi @cristianrgreco, I just made the modification for redis-stack as well. The REDIS_ARGS variable now contains the merge of the user's values with the default values.

I kept the current solution of creating the variable via `this.withEnvironment` which I recreate with the merge of both values. 
We could have also edited the existing variable in `this.createOpts.Env`. If the current solution works for you, I'll leave it as is.

packages/modules/redis/src/redis-container.ts
```diff
@@ -44,11 +44,22 @@ export class RedisContainer extends GenericContainer {
       ...(this.persistenceVolume ? ["--save 1 1 ", "--appendonly yes"] : []),
     ];
     if (this.imageName.image.includes("redis-stack")) {
+      const existingRedisArgs =
+        (this.createOpts.Env || []).find((e) => e.startsWith("REDIS_ARGS="))?.split("=", 2)[1] || "";
```
cristianrgreco: Potential bug here `?.split("=", 2)[1]`, if the value is `undefined`, then calling `[1]` will throw `TypeError`, would need to be `?.[1]`

Also please use `??` instead of `||` for specifying default values. 

I would use `this.environment` instead of `this.createOps.Env` to simplify this (don't need to split strings).

packages/modules/redis/src/redis-container.test.ts
```diff
@@ -119,4 +119,84 @@ describe("RedisContainer", { timeout: 240_000 }, () => {
     client.destroy();
     // }
   });
+
+  it("should start redis with custom command", async () => {
```
cristianrgreco, line 123: The redis containers aren't stopped at the end of the tests. Let's see if we can follow the existing `await using` syntax to have this handled automatically for us

packages/modules/redis/src/redis-container.ts
```diff
@@ -44,11 +44,22 @@ export class RedisContainer extends GenericContainer {
...
+
+      // merge with filter to remove empty items
+      const mergedRedisArgs = [existingRedisArgs, ...redisArgs].filter(Boolean).join(" ");
+
+      // remove existing REDIS_ARGS to avoid duplicates
+      this.createOpts.Env = (this.createOpts.Env || []).filter((e) => !e.startsWith("REDIS_ARGS="));
```
cristianrgreco, line 53: Instead of directly mutating `this.createOpts.Env`, it would be better to invoke `withEnvironment`

ziggornif: I've already tried to use withEnvironment like this :
```ts
const existingRedisArgs = this.environment["REDIS_ARGS"] ?? "";

// merge with filter to remove empty items
const mergedRedisArgs = [existingRedisArgs, ...redisArgs].filter(Boolean).join(" ");

this.withEnvironment({
  REDIS_ARGS: mergedRedisArgs,
}).withEntrypoint(["/entrypoint.sh"]);
```

But with this solution `container.createOpts.Env` has REDIS_ARGS env var duplicated when user passes a custom REDIS_ARGS value:

container.createOpts.Env in test `should start redis-stack with custom env and keeping default args` with this solution :

```
0 = 'REDIS_ARGS=--loglevel verbose'
1 = 'REDIS_ARGS=--loglevel verbose --save 1 1  --appendonly yes'
```

packages/modules/redis/src/redis-container.test.ts
ziggornif: Here I couldn’t use the using because I need to separate the start from the container to access the createOpts only available in the GenericContainer class, once started we are on a StartedContainer instance.

If you have a solution to access the createOpts on a StartedContainer I am interested

cristianrgreco:
```ts
await using startedContainer = await container.start();
```

ziggornif:
@cristianrgreco i remove the split maniplation, use `??` instead of `||` i kept the solution with `withEnvironment` as you said. 
It's works now by removing previous `REDIS_ARGS` value before.

I also update tests with `await using`.

Tell me if these changes correspond to what you wanted

cristianrgreco:
Thanks @ziggornif

PR #1161, opened 2025-10-14
perf: replace archiver and tar-fs with modern-tar

### LINKED ISSUE #1135: Replace `archiver` with lighter alternative
`archiver` has 64 dependencies and weighs over 7mb:
https://npmgraph.js.org/?q=archiver
https://pkg-size.dev/archiver

Some of those dependencies were just compromised: https://socket.dev/blog/npm-author-qix-compromised-in-major-supply-chain-attack. It's also responsible for about half the dependencies in this project

Perhaps it could be replaced with a lighter alternative like https://github.com/ayuhito/modern-tar? Or alternatively, `tar` or `tar-stream` (which is what `archiver` uses internally)?

ayuhito:
## Overview

Resolves #1135 using [`modern-tar`](https://github.com/ayuhito/modern-tar). Around a 48MB --> 38MB reduction for package size which adds up a lot when considering download count. We're probably also a bit faster.

The next most impactful thing could be to reduce dependencies in `dockerode`, but they seem to be targeting `node@8` so upping their limit to at least `node@18` could help us start removing dependencies there.

image

Ref: https://npmgraph.js.org/?q=testcontainers

Please let me know if there are any concerns! I'm happy to look into it all!

### DIFF
--- a/packages/testcontainers/src/generic-container/generic-container.test.ts
+++ b/packages/testcontainers/src/generic-container/generic-container.test.ts
@@ -411,7 +411,11 @@ describe("GenericContainer", { timeout: 180_000 }, () => {
       .withExposedPorts(8080)
       .start();
 
-    expect((await container.exec(`stat -c "%a %n" /tmp/newdir/test.txt`)).output).toContain("777");
+    // Check directory permissions
+    expect((await container.exec(`stat -c "%a %n" /tmp/newdir`)).output).toContain("777");
+
+    // Verify files retain their original permissions
+    expect((await container.exec(`stat -c "%a %n" /tmp/newdir/test.txt`)).output).toContain("644");
   });
 
   it("should copy directory to started container", async () => {
--- a/packages/testcontainers/src/generic-container/generic-container.ts
+++ b/packages/testcontainers/src/generic-container/generic-container.ts
@@ -1,7 +1,7 @@
-import archiver from "archiver";
 import AsyncLock from "async-lock";
 import { Container, ContainerCreateOptions, HostConfig } from "dockerode";
 import { Readable } from "stream";
+import { buffer } from "stream/consumers";
 import { containerLog, hash, log, toNanos } from "../common";
 import { ContainerRuntimeClient, getContainerRuntimeClient, ImageName } from "../container-runtime";
 import { CONTAINER_STATUSES } from "../container-runtime/clients/container/types";
@@ -178,9 +178,30 @@ export class GenericContainer implements TestContainer {
       await client.container.connectToNetwork(container, network, this.networkAliases);
     }
 
-    if (this.filesToCopy.length > 0 || this.directoriesToCopy.length > 0 || this.contentsToCopy.length > 0) {
-      const archive = this.createArchiveToCopyToContainer();
-      archive.finalize();
+    const sources = [
+      ...this.filesToCopy.map((file) => ({
+        type: "file" as const,
+        source: file.source,
+        target: file.target,
+        mode: file.mode,
+      })),
+      ...this.directoriesToCopy.map((dir) => ({
+        type: "directory" as const,
+        source: dir.source,
+        target: dir.target,
+        mode: dir.mode,
+      })),
+      ...(await Promise.all(
+        this.contentsToCopy.map(async ({ content, target, mode }) => {
+          const data = content instanceof Readable ? await buffer(content) : content;
+          return { type: "content" as const, content: data, target, mode };
+        })
+      )),
+    ];
+
+    if (sources.length > 0) {
+      const { packTar } = await import("modern-tar/fs");
+      const archive = packTar(sources);
       await client.container.putArchive(container, archive, "/");
     }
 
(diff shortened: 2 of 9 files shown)


packages/testcontainers/src/generic-container/generic-container.test.ts
```diff
@@ -411,7 +411,11 @@ describe("GenericContainer", { timeout: 180_000 }, () => {
       .withExposedPorts(8080)
       .start();
 
-    expect((await container.exec(`stat -c "%a %n" /tmp/newdir/test.txt`)).output).toContain("777");
+    // Check directory permissions
+    expect((await container.exec(`stat -c "%a %n" /tmp/newdir`)).output).toContain("777");
+
+    // Verify files retain their original permissions
+    expect((await container.exec(`stat -c "%a %n" /tmp/newdir/test.txt`)).output).toContain("644");
```
ayuhito, line 418: I think this is a bug/edge case from `archiver`, because `node-tar` and `modern-tar` do not mix directory modes and file modes together. That's not very standard behaviour.

packages/testcontainers/src/generic-container/generic-container.ts
```diff
@@ -178,9 +178,30 @@ export class GenericContainer implements TestContainer {
...
+      ...this.directoriesToCopy.map((dir) => ({
+        type: "directory" as const,
+        source: dir.source,
+        target: dir.target,
+        mode: dir.mode,
+      })),
+      ...(await Promise.all(
+        this.contentsToCopy.map(async ({ content, target, mode }) => {
+          const data = content instanceof Readable ? await buffer(content) : content;
+          return { type: "content" as const, content: data, target, mode };
```
ayuhito, line 197: `archiver` implicitly buffers `Readable`. We're explicit here (and also maybe more efficient since we use Node native buffering).

tsconfig.base.json
```diff
@@ -6,11 +6,11 @@
     "types": [
       "vitest/globals"
     ],
-    "module": "commonjs",
+    "module": "Node16",
     "declaration": true,
     "sourceMap": true,
     "strict": true,
-    "moduleResolution": "node",
+    "moduleResolution": "node16",
```
ayuhito: You may notice that I dynamically import `modern-tar` in this codebase. That's because `modern-tar` is ESM only, which also forced this `tsconfig.json` change.

I'm not sure what thoughts the maintainers have towards modules, so I've left this as my interim solution. I'd love a bit more clarification on what `testcontainers` wants to do though.

ayuhito:
I tried to validate as many tests as possible locally, but obviously this suite is very heavy so CI might fail if I didn't cover everything once these workflows are approved 🙈

packages/testcontainers/src/global.d.ts
```diff
@@ -0,0 +1,4 @@
+// Can be removed if testcontainers switches to ESM.
+declare module "modern-tar/fs" {
+  export * from "modern-tar/dist/fs/index";
+}
```
ayuhito, line 4: `module: "commonjs"` doesn't really support ESM only packages which use an [export map](https://github.com/ayuhito/modern-tar/blob/0fd121487616db202a9d2be815241e957d591c2f/package.json#L12). It can just read the old main import instead which is our browser specific export.

Alternatively, we could modernize the whole tsconfig stack. Then you could still export to CJS via something like [`tsdown`](https://tsdown.dev/). But this is the least-intrusive approach I can think of for now.

cristianrgreco: We already have a few dependencies which are ESM only, such as `get-port`, and it is enough to do an `await import("get-port")`. If modern-tar produces an export map then things should just work with TS without needing to add declarations

cristianrgreco:
Hi @ayuhito, thanks for raising an excellent PR! Although I accept that the current dependency `archiver` is too heavyweight for our needs, I am hesitant in moving to a dependency which is just a few weeks old, as opposed to using `archiver` which is tried and tested. I'd happily accept a PR to use a dependency which is tried and tested (e.g. we have seen prompt releases for security vulnerabilities, has a large user base, proven to be responsive to issues, etc).

ayuhito:
Hi @cristianrgreco, that is very reasonable and I can understand the tentativeness behind this change. 

The reason I created this new package in the first place was that there was no reasonable lightweight alternative, but it definitely needs time to mature in the ecosystem. Although I would say I've done extensive work hardening the project and matching `node-tar`'s extensive test suite. Maybe something to consider for the future then?

cristianrgreco:
> Maybe something to consider for the future then?

Definitely! Thanks again for raising the PR, hopefully will collaborate again soon 🙂

cristianrgreco:
@ayuhito Hello again 🙂 Looks like modern-tar is doing very well, I am looking at completing this migration considering our existing archiver deps have gone ESM only and are not fixing previous version vulnerabilities.

Is this something you can help with? No problem if not, I can find sometime to look into it

ayuhito:
@cristianrgreco, sounds good to me! Let me look into this soon 👍

PR #1162, opened 2025-10-15
Add Bun support

cristianrgreco:
Checking if things have changed after https://github.com/oven-sh/bun/pull/15579

---

Looks like everything works except for process lifecycle with ryuk. Because the ryuk container is not explicitly stopped, the bun process doesn't exit and so the process hangs. Issue opened https://github.com/oven-sh/bun/issues/23776.

Workaround: set `TESTCONTAINERS_RYUK_DISABLED=true`


cristianrgreco:
Closing this PR as blocked on upstream issue.

As of today it's mostly working with the workaround in this PR description.

PR #1164, opened 2025-10-20
Fix file-lock to handle existing lock files

rdziarnowski-clickup:
Changed createEmptyTmpFile to use the 'wx' flag when writing the lock file, which fails if the file already exists. Added error handling to catch and ignore EEXIST errors, allowing the lock mechanism to work correctly when the lock file already exists from a previous operation.

This prevents race conditions and improves robustness of the file locking mechanism by properly handling the case where multiple processes or sequential calls attempt to create the same lock file.

### DIFF
--- a/packages/testcontainers/src/common/file-lock.test.ts
+++ b/packages/testcontainers/src/common/file-lock.test.ts
@@ -121,11 +121,10 @@ describe("withFileLock", () => {
     const tmp = await import("tmp");
     const file = path.resolve(tmp.tmpdir, testFileName);
 
-    // Pre-create the lock file
+    // Pre-create the lock file (this simulates the file already existing)
     await writeFile(file, "", { flag: "w" });
-    expect(existsSync(file)).toBe(true);
 
-    // Should still work with existing file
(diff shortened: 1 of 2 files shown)


cristianrgreco:
Hi @rdziarnowski-clickup, I don't understand what specific scenario this PR fixes. How the temporary file is created shouldn't matter. If it already exists it means we should wait for the lock to be released, not to throw an error. If I revert all the changes made to `file-lock.ts` and run the new test suite everything passes.

packages/testcontainers/src/common/file-lock.test.ts
```diff
@@ -0,0 +1,164 @@
+import { existsSync, unlinkSync } from "fs";
+import { writeFile } from "fs/promises";
+import path from "path";
+import { withFileLock } from "./file-lock";
+
+describe("withFileLock", () => {
```
cristianrgreco, line 6: Running this test suite fails intermittently on my machine. It's because the tests by default run in parallel, and these tests which share a file lock inherently cannot run in parallel. Needs to be `describe.sequential`.

PR #1170, opened 2025-11-02
Add s3mock module

### LINKED ISSUE #1169: S3Mock Module
S3Mock is a popular [mock for S3 by Adobe](https://github.com/adobe/S3Mock).

The testcontainers for Java already have a module for it: https://testcontainers.com/modules/s3mock/

I created a testcontainer for some other Node.js project. You can find it here:
https://github.com/nikeee/lean-s3/blob/main/src/test/S3MockContainer.ts

Does it make sense to add it to testcontainers-node? If so, I'd flesh out the implementation, align it with the other modules in testcontainers-node and contribute it.

nikeee:
See #1169

I used the implementation of the localstack module as a guideline.

Resolves #1169


nikeee:
My bad! Forgot to update the lockfile in the project root

cristianrgreco:
Looks great, thanks @nikeee

PR #1193, opened 2025-12-03
Add support for disabling TLS in OpenSearch module

rafaeltab:
The default wait strategy is unable to talk to http only servers, and was not updated when security was explicitly disabled. This caused the container not to be recognized as started, when security was disabled.

The URL returned by the started container also always contained `https` for the schema, even if security was disabled.

### DIFF
--- a/packages/modules/opensearch/src/opensearch-container.test.ts
+++ b/packages/modules/opensearch/src/opensearch-container.test.ts
@@ -109,10 +109,7 @@ describe("OpenSearchContainer", { timeout: 180_000 }, () => {
 
     const client = new Client({
       node: container.getHttpUrl(),
-      auth: {
-        username: container.getUsername(),
-        password: container.getPassword(),
-      },
+      // no auth, or ssl required
     });
     // }
 


packages/modules/opensearch/src/opensearch-container.test.ts
```diff
@@ -102,4 +102,23 @@ describe("OpenSearchContainer", { timeout: 180_000 }, () => {
...
+      auth: {
+        username: container.getUsername(),
+        password: container.getPassword(),
+      },
+    });
+    // }
+
+    await client.indices.create({ index: "people" });
+
+    const { body } = await client.indices.exists({ index: "people" });
```
cristianrgreco, line 121: It's worth asserting the `container.getHttpUrl()` as well, because even without any implementation changes (other than adding an empty `withSecurityEnabled()` method) this test would pass. What we want to see here is that connecting via a `http` URL also works

packages/modules/opensearch/src/opensearch-container.test.ts
```diff
@@ -102,4 +102,23 @@ describe("OpenSearchContainer", { timeout: 180_000 }, () => {
     const { body } = await client.indices.exists({ index: "people" });
     expect(body).toBe(true);
   });
+
+  it("should be reachable with security disabled", async () => {
+    // opensearchDisableSecurity {
+    await using container = await new OpenSearchContainer(IMAGE).withSecurityEnabled(false).start();
```
cristianrgreco, line 108: I'm not sure about calling the method `withSecurityEnabled`, because it's a bit unclear what it does (the service can be insecure in many different ways), or what is the default behaviour. I wonder if something like `disableTls` would be better. WDYT?

rafaeltab: I agree, the name was confusing to me as well. I just didn't want to change it as it would be a breaking change.

packages/modules/opensearch/src/opensearch-container.test.ts
rafaeltab: It actually wouldn't pass. 

| Security | schema | result |
|--------|--------|--------|
| enabled | https | ❌ TLS is not trusted because `rejectUnauthorized` is not set to `false`  |
| enabled | http | ❌ attempt to use http when it requires https |
| disabled | https | ❌ attempt to use https when it requires http | 
| disabled | http | ✅connection successful (expected) | 

It is still good to add an assertion though, so I will.

packages/modules/opensearch/src/opensearch-container.test.ts
cristianrgreco: Oh geez I just saw that `withSecurityEnabled` was already there 🤦 . Ignore this comment

rafaeltab: Turns out, it doesn't just disable TLS, it also disables auth. A complete list of the effects can be found [here](https://docs.opensearch.org/1.1/security-plugin/index/). Taking that in consideration, the name `withSecurityEnabled` makes more sense I think, as you can fully enable/disable all security features. I made sure the test doesn't include any auth anymore, as it doesn't do anything.

PR #1242, opened 2026-02-20
Add Oracle Free module

collinmurd:
Adds a module for Oracle DB support using oracle free. Follows similar implementation to the existing Java and Python modules, which utilize [these images](https://github.com/gvenzl/oci-oracle-free)

### DIFF
--- a/packages/modules/oraclefree/src/oraclefree-container.test.ts
+++ b/packages/modules/oraclefree/src/oraclefree-container.test.ts
@@ -4,7 +4,7 @@ import { OracleDbContainer, StartedOracleDbContainer } from "./oraclefree-contai
 
 const IMAGE = getImage(__dirname);
 
-describe.sequential("OracleFreeContainer", { timeout: 240_000 }, () => {
+describe("OracleFreeContainer", { timeout: 240_000 }, () => {
   describe("default configuration", () => {
     let container: StartedOracleDbContainer;
 
@@ -35,21 +35,6 @@ describe.sequential("OracleFreeContainer", { timeout: 240_000 }, () => {
         user: container.getUsername(),
         password: container.getPassword(),
         connectString: container.getConnectionDescriptor(),
-      });
-
-      const result = await connection.execute("SELECT 1 FROM DUAL");
-      expect(result.rows![0]).toEqual([1]);
-
-      await connection.close();
-    });
-
-    it("should work with restarted container", async () => {
-      await container.restart();
-
-      const connection = await oracledb.getConnection({
-        user: container.getUsername(),
-        password: container.getPassword(),
-        connectString: container.getUrl(),
       });
 
       const result = await connection.execute("SELECT 1 FROM DUAL");
@@ -104,5 +89,21 @@ describe.sequential("OracleFreeContainer", { timeout: 240_000 }, () => {
 
     await connection.close();
     // }
+  });
+
+  it("should work with restarted container", async () => {
+    const container = await new OracleDbContainer(IMAGE).start();
+    await container.restart();
+
+    const connection = await oracledb.getConnection({
+      user: container.getUsername(),
+      password: container.getPassword(),
+      connectString: container.getUrl(),
+    });
+
+    const result = await connection.execute("SELECT 1 FROM DUAL");
+    expect(result.rows![0]).toEqual([1]);
+
+    await connection.close();
   });
 });


packages/modules/oraclefree/src/oraclefree-container.ts
```diff
@@ -0,0 +1,114 @@
...
+    this.password = password;
+    return this;
+  }
+
+  /**
+   * Sets a custom application database/service name.
+   *
+   * @throws if attempting to use the image default database name.
+   */
+  public withDatabase(database: string): this {
```
cristianrgreco, line 38: `withDatabase` should reject empty/whitespace names. Right now `withDatabase("")` is accepted, which later causes a mismatch between runtime DB and reported DB metadata.

packages/modules/oraclefree/src/oraclefree-container.ts
```diff
@@ -0,0 +1,114 @@
...
+    if (database === DEFAULT_DATABASE)
```
cristianrgreco: This check is case sensitive whereas Oracle is not, e.g. `withDatabase("freepdb1")` is accepted but Oracle will reject it later - let's add a test for this

cristianrgreco:
Thanks for raising @collinmurd!

Looks like testcontainers-java exports a single `OracleContainer` module, where you can then use free or XE based on the provided image name: 

- https://java.testcontainers.org/modules/databases/oraclefree/
- https://java.testcontainers.org/modules/databases/oraclexe/

Let's do the same here. Also update the module name from oraclefree to oracle (`npm i @testcontainers/oracle`).

collinmurd:
> Thanks for raising @collinmurd!
> 
> Looks like testcontainers-java exports a single `OracleContainer` module, where you can then use free or XE based on the provided image name:
> 
> * https://java.testcontainers.org/modules/databases/oraclefree/
> * https://java.testcontainers.org/modules/databases/oraclexe/
> 
> Let's do the same here. Also update the module name from oraclefree to oracle (`npm i @testcontainers/oracle`).

[Oracle Free is meant to be the successor to Oracle XE](https://blogs.oracle.com/database/database-23c-free-ga). I haven't found anything that indicates a retirement path from Oracle for XE, but the maintainer for these Oracle images [isn't supporting XE anymore](https://github.com/gvenzl/oci-oracle-xe?tab=readme-ov-file#new-version-of-oracle-database-and-these-images).

testcontainers-python only exports an `testcontainers[oracle-free]` package. I'm thinking we should only export with a Oracle Free module to discourage XE use.

cristianrgreco:
Ah I didn't know, makes sense then, please disregard my last comment

packages/modules/oraclefree/src/oraclefree-container.ts
collinmurd: @cristianrgreco I also figured it would be cleaner not to throw if they supply the default PDB name and treat it as a no-op instead.

packages/modules/oraclefree/src/oraclefree-container.test.ts
```diff
@@ -0,0 +1,107 @@
+import oracledb from "oracledb";
+import { getImage } from "../../../testcontainers/src/utils/test-helper";
+import { OracleDbContainer, StartedOracleDbContainer } from "./oraclefree-container";
+
+const IMAGE = getImage(__dirname);
+
+describe.sequential("OracleFreeContainer", { timeout: 240_000 }, () => {
```
cristianrgreco: These can run in parallel

packages/modules/oraclefree/Dockerfile
```diff
@@ -0,0 +1 @@
+FROM gvenzl/oracle-free:slim-faststart
```
cristianrgreco: Let's pin the image version, `gvenzl/oracle-free:23.26.1-slim-faststart` is equivalent to current latest. Dependabot will keep it up to date and will catch future breaking changes

cristianrgreco:
@collinmurd build failed

collinmurd:
> @collinmurd build failed

Sorry, totally missed it

Still researching. Tests work locally with Docker so I suspect is a resource thing

cristianrgreco:
I would make the change myself but I don't have write permission. The other modules spin up a container per test to enable running in parallel. The issue here is there's a single container for all the tests, so for example the test which checks the restart kills the container while the other tests are running.

Would you be able to make this last change? Apologies for the back and forth.

collinmurd:
> Would you be able to make this last change? Apologies for the back and forth.

No worries! Thanks for catching the restart issue. Oracle is a pretty heavy process, running more than 4 or more containers would exceed my 8GB WSL limit. So I was trying to minimize the number of containers in parallel. This leaves with exactly 3 which should hopefully work on most people's machines.

cristianrgreco:
Thanks @collinmurd for the PR and your patience with the review!

PR #1307, opened 2026-04-24
security(deps): upgrade to dockerode@5.0.0

AviVahl:
- new release drops usage of vulnerable uuid, reducing the vulnerability count.
- also fixed up package-lock.json installing older testcontainers versions for internal packages. they should all be linked now. this deduped several depedencies.

ref: https://github.com/apocas/dockerode/issues/829

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -2019,72 +2019,6 @@
         "chromadb": "^3.1.2"
       }
     },
-    "node_modules/@chroma-core/ollama/node_modules/@types/dockerode": {
-      "version": "3.3.47",
-      "resolved": "https://registry.npmjs.org/@types/dockerode/-/dockerode-3.3.47.tgz",
-      "integrity": "sha512-ShM1mz7rCjdssXt7Xz0u1/R2BJC7piWa3SJpUBiVjCf2A3XNn4cP6pUVaD8bLanpPVVn4IKzJuw3dOvkJ8IbYw==",
-      "dev": true,
-      "license": "MIT",
-      "dependencies": {
-        "@types/docker-modem": "*",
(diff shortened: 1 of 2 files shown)


AviVahl:
Heya @cristianrgreco 
Any chance you've got time to review/merge this?
We're using testcontainers and currently forcibly override dockerode to fix alerts.

cristianrgreco:
Hi @AviVahl, I typically defer to Dependabot for updating deps. See https://github.com/testcontainers/testcontainers-node/pull/1322 which updates Dockerode + several others. 

The reason it's taking a while is because many deps have now dropped support for node 20 (EOL), meaning that I have to too, meaning I have to plan a major release (as I have to bump the minimum node engine requirement). That's why this is taking a while (as well as fitting in this work around my job 😄).

Current issues:
- Undici version bump is breaking the arangojs client, issue raised https://github.com/arangodb/arangojs/issues/855. I will likely just defer the undici upgrade if everything else is compatible
- Kafka/redpanda/qdrant are failing with the new v26 LTS due to node-gyp, so the build is currently failing. I can add v26 to the test matrix later though so that's not blocking
- Fix up https://github.com/testcontainers/testcontainers-node/pull/1096 to include as part of the major release

I'm hoping to get this all resolved within the next few days.

AviVahl:
Thank you for the response and explanation.
I'll close this one in favor of the dependabot PR, and will wait for the next major.
Much appreciated.

cristianrgreco:
FYI @AviVahl, v12.0.0 which includes these dep updates has just been released

AviVahl:
Awesome. Thank you. We'll upgrade on our end.

PR #1383, opened 2026-06-22
Throw when FixedPortGenerator runs out of ports

cristianrgreco:
## Summary
`FixedPortGenerator.generatePort()` returned `this.ports[this.portIndex++]` with no bounds check, yielding `undefined` (typed as `number`) once the pool was exhausted. It now throws a clear error so misuse fails fast.

## Verification
- Added a unit test asserting the first call returns the seeded port and the next throws (no Docker required).
- `npm ci` && `npx vitest run packages/testcontainers/src/utils/port-generator.test.ts` → **passed**
- `npm run check-compiles` → clean · `npm run lint` → clean

## Test results
New unit test passing.

## Not breaking
Adds a thrown error only on the previously-undefined exhaustion path (which returned a non-number). No change to valid usage or to `RandomPortGenerator`.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/packages/testcontainers/src/utils/port-generator.test.ts
+++ b/packages/testcontainers/src/utils/port-generator.test.ts
@@ -9,11 +9,13 @@ describe("PortGenerator", () => {
       await expect(fixedPortGenerator.generatePort()).resolves.toBe(1001);
     });
 
-    it("should throw when no more ports are available", async () => {
+    it("should reject when no more ports are available", async () => {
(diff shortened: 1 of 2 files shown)


cristianrgreco:
Closing: FixedPortGenerator is an internal, test-only helper — it is not exported from the package and is not referenced by any production code (only RandomPortGenerator is). The exhaustion path this guarded can't be reached today, so the change isn't worth carrying.

PR #1390, opened 2026-06-28
Add Mosquitto module

ChxGuillaume:
## Add Mosquitto module

This PR introduces a new `@testcontainers/mosquitto` module for the [Eclipse Mosquitto](https://mosquitto.org/) MQTT broker, following the same structure as the existing HiveMQ module.

## What's included

- `MosquittoContainer` — starts an `eclipse-mosquitto:2` container with a custom config injected at startup, since Mosquitto 2.x disallows anonymous connections by default
- Anonymous mode (`allow_anonymous true`) out of the box
- Authenticated mode via `withUsername()` / `withPassword()` — creates the password file using `mosquitto_passwd -b` before launching the broker, with a `chown` to fix ownership since Mosquitto drops privileges to the `mosquitto` user
- `StartedMosquittoContainer.getPort()` and `getConnectionString()` — the latter includes credentials in the URL when auth is configured
- Two test files (`mosquitto-container.test.ts` and `mosquitto-container-auth.test.ts`) kept separate to avoid a codeinclude block-name collision between `mosquittoConnect` and `mosquittoConnectWithCredentials`
- Documentation at `docs/modules/mosquitto.md` with codeinclude examples for both usage patterns, added to the mkdocs navigation


mkdocs.yml
```diff
@@ -73,6 +73,7 @@ nav:
       - Localstack: modules/localstack.md
       - MariaDB: modules/mariadb.md
       - MinIO: modules/minio.md
+      - Mosquitto: modules/mosquitto.md
```
cristianrgreco: Nit: this nav list is alphabetical, so `Mosquitto` should sort after `MongoDB` (line 78), not between MinIO and Mockserver (`Mo`**`c`** ` < Mo`**`n`** ` < Mo`**`s`**`). Could you move this entry below the MongoDB line?

packages/modules/mosquitto/src/mosquitto-container.ts
```diff
@@ -0,0 +1,69 @@
...
+const PASSWORD_FILE_PATH = "/mosquitto/config/passwords";
+
+export class MosquittoContainer extends GenericContainer {
+  private username?: string;
+  private password?: string;
+
+  constructor(image: string) {
+    super(image);
+    this.withExposedPorts(MQTT_PORT)
+      .withWaitStrategy(Wait.forLogMessage(/running mosquitto as user/i))
```
cristianrgreco: This keys on the log line `running mosquitto as user`, which fires when the process starts rather than when the listener is accepting connections, and the wording can drift across `eclipse-mosquitto:2` minor tags. Since the tests connect immediately after `start()`, could you confirm this is stable across tags — or switch to a port/listener-based wait for robustness?

packages/modules/mosquitto/src/mosquitto-container.ts
```diff
@@ -0,0 +1,69 @@
...
+
+  public withUsername(username: string): this {
+    this.username = username;
+    return this;
+  }
+
+  public withPassword(password: string): this {
+    this.password = password;
+    return this;
+  }
```
cristianrgreco, line 26: For consistency with the other credentialed modules (e.g. MongoDB's `withUsername`/`withPassword`), consider rejecting empty values:

```ts
public withUsername(username: string): this {
  if (!username) throw new Error("Username should not be empty.");
  this.username = username;
  return this;
}
```

packages/modules/mosquitto/src/mosquitto-container.ts
```diff
@@ -0,0 +1,69 @@
...
+
+  public override async start(): Promise<StartedMosquittoContainer> {
+    if (this.username !== undefined && this.password !== undefined) {
```
cristianrgreco, line 33: Auth is only enabled when **both** username and password are set, so calling just `withUsername(...)` silently starts an anonymous broker and returns a credential-less connection string. Consider throwing when exactly one of the two is provided so misconfiguration fails loudly rather than silently degrading.

packages/modules/mosquitto/src/mosquitto-container-auth.test.ts
```diff
@@ -0,0 +1,36 @@
+import mqtt from "mqtt";
```
cristianrgreco: Could this be merged into `mosquitto-container.test.ts` as a second `it()`? The PR description cites a codeinclude block-name collision, but codeinclude blocks are referenced by their distinct names — `mosquittoConnect` and `mosquittoConnectWithCredentials` don't collide in one file. Every other module keeps multiple blocks in a single `*-container.test.ts` (e.g. redis has `redisStartContainer` + `redisStartWithCredentials` together; azurite has 8 blocks in one file), so a single file would match the convention.

cristianrgreco (commented):
Thanks for the module — nice and clean, closely following the HiveMQ structure. A few notes inline. Two points that don't anchor well to a single line:

- **`package-lock.json`**: the diff also removes `libc` arrays from several unrelated packages (lightningcss and one MPL-2.0 dep). That looks like npm-version drift rather than part of this module. Could you regenerate the lockfile with the project's npm version so the only change here is the new `@testcontainers/mosquitto` entry?

- **Prior art**: for what it's worth, Mosquitto modules already exist in .NET (`Testcontainers.Mosquitto`), Python (`testcontainers[mqtt]`) and Rust (`testcontainers-rs-modules-community`), so this is a welcome addition. The .NET module is the closest reference if we ever want cross-language API consistency.

packages/modules/mosquitto/Dockerfile
```diff
@@ -0,0 +1 @@
+FROM eclipse-mosquitto:2
```
cristianrgreco: Could we pin a concrete version here rather than the floating major tag `eclipse-mosquitto:2`? Every other module pins a specific tag (e.g. `redis:8.8`, `mongo:8.2.11`, `valkey/valkey:9.1`, `hivemq/hivemq-ce:2026.5`) so that a new upstream release can't break CI without a deliberate bump in this repo. With `:2`, any future Mosquitto 2.x publish is picked up automatically. Suggest something like `eclipse-mosquitto:2.0.20` (or whatever the current 2.x patch is).

ChxGuillaume: Seems logic, they have quite weird tag naming so used the latest "non specific tag" so non -alpine
Can be found there: https://github.com/testcontainers/testcontainers-node/pull/1390/commits/09e46770f764ebdce58b8735bf1c2e81a087a426

packages/modules/mosquitto/src/mosquitto-container-auth.test.ts
ChxGuillaume: Yea i went the easy round and did not search too much into it, i think it was a problem of overlaping block names
Changed that in https://github.com/testcontainers/testcontainers-node/pull/1390/commits/6574f9f44d560104970985251db7f44f05574c75

packages/modules/mosquitto/src/mosquitto-container.ts
ChxGuillaume: Oh yea sure, can find the changes in there: https://github.com/testcontainers/testcontainers-node/pull/1390/commits/e5b5965a4bff4346b4980852e20920b76dad93d0

packages/modules/mosquitto/src/mosquitto-container.ts
ChxGuillaume: That's a good point, can find changes here https://github.com/testcontainers/testcontainers-node/pull/1390/commits/1723a5af30b00a87ef84340c372f81def0782a2e

packages/modules/mosquitto/src/mosquitto-container.ts
ChxGuillaume: Okay, switched to a port based wait strategy, tested and seem to work properly
Changes in https://github.com/testcontainers/testcontainers-node/pull/1390/commits/2d6560723422d9bdcf4b9b3fa47f41897880492c

cristianrgreco (commented):
Thanks for the quick turnaround — this addresses almost all of the earlier feedback:

- ✅ Image pinned to `eclipse-mosquitto:2.0.22`
- ✅ Nav entry now alphabetically placed (after MongoDB)
- ✅ Wait strategy switched to `Wait.forListeningPorts()` — more robust than the log regex
- ✅ `withUsername`/`withPassword` reject empty values
- ✅ Partial auth now fails loudly via the XOR check (`Both username and password must be set together.`)
- ✅ Tests merged into a single `mosquitto-container.test.ts` with two `it()` blocks, docs updated

The XOR guard and the `forListeningPorts()` switch both look correct. One item from the previous round is still outstanding:

**`package-lock.json`** — the diff still strips `libc` arrays from several unrelated packages (lightningcss and one MPL-2.0 dep); the hunk is unchanged from before. This is npm-version drift rather than part of the module. Could you regenerate the lockfile with the project's npm version (or revert those specific hunks) so the only change is the new `@testcontainers/mosquitto` entry? Once that's cleaned up I think this is good to go.

ChxGuillaume:
> Thanks for the quick turnaround — this addresses almost all of the earlier feedback:
> 
> * ✅ Image pinned to `eclipse-mosquitto:2.0.22`
> * ✅ Nav entry now alphabetically placed (after MongoDB)
> * ✅ Wait strategy switched to `Wait.forListeningPorts()` — more robust than the log regex
> * ✅ `withUsername`/`withPassword` reject empty values
> * ✅ Partial auth now fails loudly via the XOR check (`Both username and password must be set together.`)
> * ✅ Tests merged into a single `mosquitto-container.test.ts` with two `it()` blocks, docs updated
> 
> The XOR guard and the `forListeningPorts()` switch both look correct. One item from the previous round is still outstanding:
> 
> **`package-lock.json`** — the diff still strips `libc` arrays from several unrelated packages (lightningcss and one MPL-2.0 dep); the hunk is unchanged from before. This is npm-version drift rather than part of the module. Could you regenerate the lockfile with the project's npm version (or revert those specific hunks) so the only change is the new `@testcontainers/mosquitto` entry? Once that's cleaned up I think this is good to go.

Hey i'm quite not used to handle this kind of package-lock, i've tried to generate it but it was always removing them, is it because i'm working on a Mac or something would like know if you can help me learn more on that thing 😅

Updated it kind "manually" to have them back but i'm not a huge fan of this move...

cristianrgreco:
@ChxGuillaume Which version of Node/NPM are you using?

ChxGuillaume:
> @ChxGuillaume Which version of Node/NPM are you using?

npm --version && node -v
11.16.0
v24.18.0

Just installed the last Node version with NVM to try and generate it

packages/modules/mosquitto/src/mosquitto-container.ts
```diff
@@ -0,0 +1,73 @@
...
+export class StartedMosquittoContainer extends AbstractStartedContainer {
+  private readonly port: number;
+
+  constructor(
+    startedTestContainer: StartedTestContainer,
+    private readonly username?: string,
+    private readonly password?: string
+  ) {
+    super(startedTestContainer);
+    this.port = startedTestContainer.getMappedPort(MQTT_PORT);
```
cristianrgreco: The mapped port is snapshotted here at construction time, so `getPort()` (and therefore `getConnectionString()`) returns a stale value after `container.restart()` — Docker assigns a new random host port on restart, and `StartedGenericContainer.restart()` refreshes its bound ports accordingly, but this cached `this.port` won't pick that up.

Resolving it on demand keeps it correct across restarts (this is what MongoDB/RabbitMQ do — e.g. `getConnectionString()` calls `this.getMappedPort(...)` inline):

```ts
export class StartedMosquittoContainer extends AbstractStartedContainer {
  constructor(
    startedTestContainer: StartedTestContainer,
    private readonly username?: string,
    private readonly password?: string
  ) {
    super(startedTestContainer);
  }

  public getPort(): number {
    return this.getMappedPort(MQTT_PORT);
  }
  // ...
}
```

Minor in practice (these rarely get restarted mid-test, which is why the same pattern in HiveMQ has gone unnoticed), but the fix is trivial and matches the convention in the other modules.

ChxGuillaume: Ah yea, applied it here: https://github.com/testcontainers/testcontainers-node/pull/1390/commits/f25806a9ae21329af43268dcc3792bc2bb1471bb

cristianrgreco:
@ChxGuillaume Two things to get CI green:

**1. Lint failure (`Lint (mosquitto)`)**

It's just a Prettier formatting nit — an extra space:

```
packages/modules/mosquitto/src/mosquitto-container.ts
  60:3  error  Delete `·`  prettier/prettier
```

CI runs ESLint without `--fix`, so it fails on it. From the repo root just run:

```bash
npm run lint
```

That auto-fixes it (it's `eslint --fix` with the Prettier plugin). `npm run format` works too. Then commit the result.

**2. The `package-lock.json`**

Now that you're on Node 24.18 / npm 11.16 you're on the right toolchain — that version doesn't prune the cross-platform optional deps, so no need to edit anything by hand. The key is to start from a clean lockfile rather than regenerating on top of the manually-edited one:

```bash
git checkout main -- package-lock.json   # discard the hand-edited version
npm install --package-lock-only           # npm 11.16 re-resolves cleanly
git diff package-lock.json                 # should show ONLY the @testcontainers/mosquitto entry
```

If that diff shows just the new `@testcontainers/mosquitto` block (plus your module's workspace entry) and nothing about sharp/couchbase/lightningcss/`libc`, it's correct — commit it.

For background on why it was happening before: the disappearing packages are platform-specific native binaries (Linux/Windows/musl), and older npm versions rewrite the lockfile from what's installed locally, dropping everything that doesn't match your Mac. npm 11.x fixes that, so you're good now. 👍

cristianrgreco (approved):

cristianrgreco:
Thanks for your contribution and for the fast feedback @ChxGuillaume

PR #1396, opened 2026-07-01
Add MongoDB Atlas Local support

jbl428:
- Add `MongoDBAtlasLocalContainer` to the MongoDB module
- Support Atlas Local connection strings and optional credentials
- Add Atlas Search example coverage and module exports
- Document Atlas Local usage alongside the existing MongoDB docs

### DIFF
--- a/docs/modules/mongodb.md
+++ b/docs/modules/mongodb.md
@@ -6,7 +6,7 @@
 npm install @testcontainers/mongodb --save-dev
 ```
 
-## Examples
+## MongoDBContainer
 
 These examples use the following libraries:
 
@@ -28,7 +28,7 @@ Choose an image from the [container registry](https://hub.docker.com/_/mongo) an
 [](../../packages/modules/mongodb/src/mongodb-container.test.ts) inside_block:connectWithCredentials
 <!--/codeinclude-->
 
-## MongoDB Atlas Local
+## MongoDBAtlasLocalContainer
 
 The MongoDB Atlas Local image combines the MongoDB database engine with MongoT for Atlas Search.
 
--- a/docs/modules/mongodb.md
+++ b/docs/modules/mongodb.md
@@ -27,3 +27,34 @@ Choose an image from the [container registry](https://hub.docker.com/_/mongo) an
 <!--codeinclude-->
 [](../../packages/modules/mongodb/src/mongodb-container.test.ts) inside_block:connectWithCredentials
 <!--/codeinclude-->
+
+## MongoDB Atlas Local
+
+The MongoDB Atlas Local image combines the MongoDB database engine with MongoT for Atlas Search.
+
+Choose an image from the [container registry](https://hub.docker.com/r/mongodb/mongodb-atlas-local) and substitute `IMAGE`.
+The connection string returned by `getConnectionString()` does not include `directConnection=true`; pass `directConnection: true` to your MongoDB client options instead.
+
+### Get a connection string
+
+<!--codeinclude-->
+[](../../packages/modules/mongodb/src/mongodb-atlas-local-container.test.ts) inside_block:connectAtlasLocal
+<!--/codeinclude-->
+
+### Get a database connection string
+
+<!--codeinclude-->
+[](../../packages/modules/mongodb/src/mongodb-atlas-local-container.test.ts) inside_block:connectAtlasLocalDatabase
+<!--/codeinclude-->
+
+### Connect with credentials
+
+<!--codeinclude-->
+[](../../packages/modules/mongodb/src/mongodb-atlas-local-container.test.ts) inside_block:connectAtlasLocalWithCredentials
+<!--/codeinclude-->
+
+### Create an Atlas Search index and search it
+
+<!--codeinclude-->
+[](../../packages/modules/mongodb/src/mongodb-atlas-local-container.test.ts) inside_block:createAtlasIndexAndSearchIt
+<!--/codeinclude-->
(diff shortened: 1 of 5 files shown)


packages/modules/mongodb/src/mongodb-atlas-local-container.ts
```diff
@@ -0,0 +1,71 @@
...
+    url.hostname = this.getHost();
+    url.port = this.getMappedPort(MONGODB_PORT).toString();
+    url.pathname = databaseName ? `/${databaseName}` : "/";
+
+    if (this.username && this.password) {
+      url.username = this.username;
+      url.password = this.password;
+    }
+
+    return url.toString();
```
cristianrgreco, line 72: With credentials, `getDatabaseConnectionString()` returns `.../test` with no `authSource`, so the driver auths against `test` — but the root user lives in `admin`, and auth fails. `MongoDBContainer` and testcontainers-go both set `authSource=admin`. Add it when creds are present:

```ts
url.searchParams.set("authSource", "admin");
```

packages/modules/mongodb/src/mongodb-atlas-local-container.test.ts
```diff
@@ -0,0 +1,137 @@
...
+
+    expect(container.getConnectionString()).toBe(
+      `mongodb://customUsername:customPassword@${container.getHost()}:${container.getMappedPort(27017)}/`
+    );
+    expect(container.getDatabaseConnectionString()).toBe(
+      `mongodb://customUsername:customPassword@${container.getHost()}:${container.getMappedPort(27017)}/test`
+    );
+  });
+
+  it("should connect to mongodb atlas local", async () => {
```
cristianrgreco, line 61: This only asserts the connection string — it never opens an authenticated connection, so it won't catch the missing `authSource` above. Add a test that connects with credentials and does an insert.

packages/modules/mongodb/src/mongodb-atlas-local-container.test.ts
```diff
@@ -0,0 +1,137 @@
+import mongoose from "mongoose";
+import { IntervalRetry } from "../../../testcontainers/src/common";
+import { MongoDBAtlasLocalContainer } from "./mongodb-atlas-local-container";
+
+const IMAGE = "mongodb/mongodb-atlas-local:8.2.2";
```
cristianrgreco: Hardcoding the tag means Renovate won't update it. Add `FROM mongodb/mongodb-atlas-local:8.2.2` as a second line in `Dockerfile` and use `getImage(__dirname, 1)` (see `packages/modules/chromadb`).

cristianrgreco (commented):
Nice work — mirrors testcontainers-java's `MongoDBAtlasLocalContainer` and correctly relies on the managed image. Main asks: credentialed connection strings are missing `authSource=admin` (and no test actually authenticates), and the test image should be pinned in the Dockerfile. Rest is inline.

Keeping this as its own container instead of a flag on `MongoDBContainer` is the right call — different image, different internals, and matches java/go.

docs/modules/mongodb.md
```diff
@@ -27,3 +27,34 @@ Choose an image from the [container registry](https://hub.docker.com/_/mongo) an
 <!--codeinclude-->
 [](../../packages/modules/mongodb/src/mongodb-container.test.ts) inside_block:connectWithCredentials
 <!--/codeinclude-->
+
+## MongoDB Atlas Local
```
cristianrgreco: `## MongoDB Atlas Local` is an H2 sibling of `## Examples`, so the base container's examples end up at the same level as the whole Atlas container. Rename `## Examples` to `## MongoDBContainer` so both containers are peer H2 sections.

packages/modules/mongodb/src/mongodb-atlas-local-container.test.ts
```diff
@@ -0,0 +1,137 @@
...
+    const collection = db.collection("test");
+    await collection.insertOne(obj);
+
+    const result = await collection.findOne({ value: 1 });
+    expect(result).toEqual(obj);
+
+    await db.close();
+  });
+
+  // createAtlasIndexAndSearchIt {
```
cristianrgreco: This marker is outside `it(...)`, so the doc snippet includes the test wrapper and `try/finally`. The other three blocks in this file put the marker inside the `it` body — move it inside here too for consistency and a cleaner example.

packages/modules/mongodb/src/mongodb-atlas-local-container.ts
```diff
@@ -0,0 +1,71 @@
...
+  public withPassword(password: string): this {
+    if (!password) throw new Error("Password should not be empty.");
+    this.password = password;
+    return this;
+  }
+
+  public override async start(): Promise<StartedMongoDBAtlasLocalContainer> {
+    if (this.username && this.password) {
+      this.withEnvironment({
+        MONGODB_INITDB_ROOT_USERNAME: this.username,
```
cristianrgreco, line 34: Worth a comment that this is `MONGODB_INITDB_*` (vs `MONGO_INITDB_*` in `mongodb-container.ts`) — different by design, easy to "fix" incorrectly to match the sibling file.

docs/modules/mongodb.md
jbl428: Fixed in 91296cf7
Also renamed the Atlas section to `## MongoDBAtlasLocalContainer` to match the class-name convention on both H2 headings.
Just flagging that this is the only module doc using class names as headings (others like gcloud.md use a single `## Examples` with `###` per container) — fine to keep it as the exception, just wanted to confirm.

cristianrgreco (approved):
All feedback addressed — `authSource=admin` is now set for credentialed connection strings and covered by a test that actually authenticates against the `/test` database, the Atlas image is pinned in the Dockerfile, and the docs/codeinclude structure is consistent. Nice work. LGTM.

PR #1445, opened 2026-08-27
Add InfluxDB module

JonnasFigueiredo:
Adds `@testcontainers/influxdb` for testing against InfluxDB OSS 1.x and 2.x, with separate classes for each major version:

- `InfluxDB1Container` / `StartedInfluxDB1Container`: creates an authenticated application user with access to the default `test` database and exposes administrator credentials for administrative queries.
- `InfluxDB2Container` / `StartedInfluxDB2Container`: initializes an organization, bucket, user, and usable default admin token. Passwords shorter than eight characters fail before container startup.

The selected class determines the configuration; image tags are not parsed. Examples and integration tests use `influxdb:1.13` and `influxdb:2.9`. InfluxDB 3 is outside this module's current scope.

Documentation includes complete examples with imports, explicit image versions, authentication, and assertions that the written points are returned. Custom configuration tests exercise the actual credentials, database, organization, bucket, token, and retention policy.

### Verification

Validated locally on Windows with Node.js 24.16.0 and Docker 29.5.3:

- `npm ci --no-audit --no-fund` using `NODE_USE_SYSTEM_CA=1` for the local certificate trust store; `package-lock.json` unchanged.
- `npm run format` and `npm run lint`: passed.
- `npm exec -- eslint 'packages/modules/influxdb/src/**/*.ts' --max-warnings=0`: passed.
- `npm run build --ignore-scripts --workspace packages/testcontainers -- --project tsconfig.json`: passed.
- `npm run build --ignore-scripts --workspace packages/modules/influxdb -- --project tsconfig.json --noEmit`: passed, including test types.
- `npm run build --ignore-scripts --workspace packages/modules/influxdb`: passed.
- `npm test -- packages/modules/influxdb`: all 9 tests passed.
- Executed both examples extracted from the documentation against real containers: both passed.
- `git diff --check`: passed.

Regression evidence: the default-configuration write/query tests were first run against the original combined class. InfluxDB 1.x failed with HTTP 404 because the default database did not exist; InfluxDB 2.x failed with unauthorized access because no token was exposed. Both tests passed after switching to the corrected classes.

### Compatibility

This is an additive module (`enhancement` / `minor`). It does not change existing released packages or the core Testcontainers API. The class split replaces the unreleased API introduced by this PR.

This picks up the earlier effort in #1130, which was closed for missing docs. Thanks @mdodsworth for the original attempt.

### DIFF
--- a/docs/modules/influxdb.md
+++ b/docs/modules/influxdb.md
@@ -1,0 +1,29 @@
+# InfluxDB
+
+## Install
+
+```bash
+npm install @testcontainers/influxdb --save-dev
+```
+
+## Examples
+
+The InfluxDB 2.x examples use the following libraries:
+
+- [@influxdata/influxdb-client](https://www.npmjs.com/package/@influxdata/influxdb-client)
+
+        npm install @influxdata/influxdb-client
+
+`InfluxDBContainer` supports both InfluxDB 2.x and the legacy 1.x line. The major version is derived from the image tag, so substitute `IMAGE` with a `2.x` tag (the default flavour) or a `1.x` tag from the [container registry](https://hub.docker.com/_/influxdb).
+
+### Write and query points (InfluxDB 2.x)
+
+<!--codeinclude-->
+[](../../packages/modules/influxdb/src/influxdb-container.test.ts) inside_block:influxdb2WriteAndQuery
+<!--/codeinclude-->
+
+### Write and query points (InfluxDB 1.x)
+
+<!--codeinclude-->
+[](../../packages/modules/influxdb/src/influxdb-container.test.ts) inside_block:influxdb1WriteAndQuery
+<!--/codeinclude-->
--- a/mkdocs.yml
+++ b/mkdocs.yml
@@ -67,6 +67,7 @@ nav:
       - Etcd: modules/etcd.md
       - GCloud: modules/gcloud.md
       - HiveMQ: modules/hivemq.md
+      - InfluxDB: modules/influxdb.md
       - K3s: modules/k3s.md
       - Kafka: modules/kafka.md
       - KurrentDB: modules/kurrentdb.md
--- a/package-lock.json
+++ b/package-lock.json
@@ -3627,6 +3627,13 @@
         "url": "https://opencollective.com/libvips"
       }
     },
+    "node_modules/@influxdata/influxdb-client": {
+      "version": "1.35.0",
+      "resolved": "https://registry.npmjs.org/@influxdata/influxdb-client/-/influxdb-client-1.35.0.tgz",
+      "integrity": "sha512-woWMi8PDpPQpvTsRaUw4Ig+nOGS/CWwAwS66Fa1Vr/EkW+NEwxI8YfPBsdBMn33jK2Y86/qMiiuX/ROHIkJLTw==",
+      "dev": true,
+      "license": "MIT"
+    },
     "node_modules/@inquirer/ansi": {
       "version": "2.0.5",
       "resolved": "https://registry.npmjs.org/@inquirer/ansi/-/ansi-2.0.5.tgz",
@@ -5076,6 +5083,10 @@
     },
     "node_modules/@testcontainers/hivemq": {
       "resolved": "packages/modules/hivemq",
+      "link": true
+    },
+    "node_modules/@testcontainers/influxdb": {
+      "resolved": "packages/modules/influxdb",
       "link": true
     },
     "node_modules/@testcontainers/k3s": {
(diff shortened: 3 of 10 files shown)


cristianrgreco:
Hi @JonnasFigueiredo, thanks for raising the PR. Apologies in the delay in reviewing it. A few issues:

### 1. Split into one class per major version

Right now the version is guessed from the image tag, and anything that doesn't parse as `1.x` is treated as 2.x. That breaks on several official tags that exist today:

- `3-core`, `3.11-core`, `core` and `enterprise` are InfluxDB 3. They get the 2.x setup, and `influxdb:3-core` hangs until the startup timeout ("URL /ping not accessible").
- `data`, `meta` and `data-alpine` are 1.x Enterprise, and they're also configured as 2.x.

Instead of making the tag parsing stricter, I'd prefer separate classes in this same package:

```ts
export { InfluxDB1Container, StartedInfluxDB1Container } from "./influxdb1-container";
export { InfluxDB2Container, StartedInfluxDB2Container } from "./influxdb2-container";
```

This also:
- gets rid of getters that mean nothing for one of the versions (`getOrganization()`/`getBucket()` on 1.x, `getDatabase()` on 2.x) and the `isInfluxDB2()` check;
- leaves room for an `InfluxDB3Container` later. The open 3.x PRs in Java and Python both add a separate class, because 3.x uses a different port (8181), startup command and token model.

### 2. InfluxDB 2.x: set a default admin token

If `withAdminToken()` isn't called, `getAdminToken()` returns `undefined`. InfluxDB does generate a token, but there's no easy way to read it back, so the client can't be used out of the box. Please set a default token, so the getter can return `string` and the zero-config path just works.

### 3. InfluxDB 1.x: the defaults don't give a usable user

With the defaults (auth on, no database), the created user can't do anything: `CREATE DATABASE` returns 403 ("requires admin privilege"), and the admin credentials aren't exposed. The docs example avoids this with `withAuthEnabled(false)`, so the default path isn't tested. I'd suggest defaulting the database (e.g. `test`) so the user gets privileges on it, and adding `getAdminUsername()` / `getAdminPassword()`.

### 4. Tests

- The 1.x test can't fail. InfluxDB returns 200 even for errors (for example `{"results":[{"statement_id":0,"error":"database not found: nope"}]}`), so `expect(body.results).toBeDefined()` passes even if the write went nowhere. Please assert that the written point comes back.
- "should apply custom configuration" only checks that the getters return what was set. It would still pass if the settings never reached the container. Writing and querying with the custom token/org/bucket would test it for real.
- Please also add a test for the default 1.x path with auth enabled.

### 5. Nits

- The class comment says tags "below `2`" are treated as 2.x; the code does the opposite. (This goes away with the split.)
- Please bump the images to the current lines: `influxdb:2.9` and `influxdb:1.13`.
- The docs say to substitute `IMAGE`, but the examples use `INFLUXDB2_IMAGE` / `INFLUXDB1_IMAGE`.
- Optional: 2.x setup rejects passwords shorter than 8 characters, but that only shows up in the container logs, and the caller waits out the full 120s timeout. A length check in `start()` would fail fast.

Happy to help with any of this.

JonnasFigueiredo:
Thanks for the detailed review, @cristianrgreco! I've addressed the points in 68bd0ee: split the classes by major version, fixed the default token and database setup, exposed the 1.x admin credentials, strengthened the integration tests, and updated the images and documentation. I also added the short-password validation.

All 9 module tests and both documentation examples pass locally against real containers, along with lint and compilation. Ready for another look whenever you have time. Thanks!

cristianrgreco:
Thanks @JonnasFigueiredo, this covers everything from the last round.

One thing before merging:

- **Docs:** please switch back to `<!--codeinclude-->` with `inside_block` markers in the tests instead of hand-written snippets. Every other module doc includes its examples from the tests, so they're compiled and run in CI. Inline copies drift. Keep the "Choose an image from the [container registry](https://hub.docker.com/_/influxdb) and substitute `IMAGE`." line like the other docs. The paragraphs describing the defaults are useful, so keep those.

Nits:

- The tests import from `./index`. Other modules import from the container file directly (`./influxdb1-container`, `./influxdb2-container`).
- `influxdb1-container.test.ts` repeats the Basic auth header and the expected `cpu_load` result in every test. A couple of small helpers in the test file would cut that down.

JonnasFigueiredo:
Done in bce6d9d. Docs are back on codeinclude with `inside_block` markers, the tests import from `./influxdb1-container` and `./influxdb2-container`, and the repeated Basic auth header and `cpu_load` result are helpers now. I left the header and the expected payload inline in the first 1.x test on purpose, since that is the block the docs pull in.

9 tests pass locally and `mkdocs build --strict` renders both blocks. The deploy preview is red, but the docs build fine here and that site also errored on a few dependabot PRs, so it looks unrelated. Could you retry it?

JonnasFigueiredo:
The red jobs on this run are the 4 minio ones, not influxdb. They fail before starting a container, because `quay.io/minio/minio` is private now: `unauthorized: access to the requested resource is not authorized`. It hits every PR and main, so I opened #1466 with a fix. All influxdb jobs here are green.

PR #1447, opened 2026-08-27
Apply `withStartupTimeout()` to the port-binding pre-wait

### LINKED ISSUE #1446: withStartupTimeout() is not applied to the port-binding pre-wait (hardcoded 10 s) — measured under CPU contention
### Problem

`GenericContainer.start()` waits for the host port bindings to appear (`inspectContainerUntilPortsExposed`) **before** the wait strategy runs. That pre-wait has a hardcoded 10 s budget: the util has `timeout = 10_000` as a default parameter, and neither call site in `generic-container.ts` passes `this.startupTimeoutMs`. So `withStartupTimeout(...)` only extends the wait strategy; a container whose port binding takes longer than 10 s fails with

```
Timed out after 10000ms while waiting for container ports to be bound to the host
```

no matter what startup timeout the caller configured.

This looks like #1281, but I'd like to add a measured data point that is not Kafka/flaky-environment related — the pre-wait itself is the thing that trips, and only under CPU contention.

### Measurements (testcontainers 12.1.0, Node 24.19, Docker 29.x, `postgres:18-alpine`)

- Local (Apple M3 Pro, one container): host port bound in **1.1–2.6 s**.
- Local, six workspace packages each starting one `postgres:18-alpine` at the same time (vitest workers): successful starts bound their ports in **2.2–7.2 s**; a cold image pull pushed some past 10 s.
- GitHub Actions `ubuntu-latest` (2 vCPU, `nproc` = 2), same six concurrent containers: one start measured **16.7 s** until the port was bound — the container was healthy afterwards, it was just slow under load. It failed at 10 s five times in one workflow run although every container had `withStartupTimeout(120_000)`.

So the 10 s default is not "a few milliseconds of Docker Desktop race" in practice: on a loaded host the port binding legitimately takes longer, and the configured startup timeout is the natural budget for it.

### Root cause

`packages/testcontainers/src/generic-container/generic-container.ts` — both calls omit the third argument:

```ts
inspectResult = await inspectContainerUntilPortsExposed(() => client.container.inspect(container), container.id);   // reuseContainer
const inspectResult = await inspectContainerUntilPortsExposed(
  () => client.container.inspect(container),
  container.id
);                                                                                                                    // startContainer
```

`inspect-container-util-ports-exposed.ts`: `timeout = 10_000`.

### Proposed fix

Pass `this.startupTimeoutMs` at both call sites. When `withStartupTimeout` was never called the value is `undefined`, so the default parameter still yields 10 s — behaviour only changes for callers who explicitly asked for a longer budget. We run exactly this as a `pnpm patch` since yesterday; the same suite went from 5 pre-wait timeouts per run to 0. PR follows.

### Environment

- testcontainers: 12.1.0 (`latest` at the time of writing)
- Node.js: v24.19.0
- Docker: 29.6.1 (Docker Desktop, arm64) locally; GitHub-hosted `ubuntu-latest` in CI

kenzox:
Fixes #1446.

`GenericContainer.start()` waits for host port bindings (`inspectContainerUntilPortsExposed`) before the wait strategy runs, but neither call site passes `this.startupTimeoutMs`, so that pre-wait is always capped at the util's 10 s default and `withStartupTimeout()` does not apply to it.

This passes `this.startupTimeoutMs` at both call sites (`reuseContainer` and `startContainer`). When no startup timeout was configured the argument is `undefined` and the existing 10 s default still applies, so behaviour only changes for callers who explicitly asked for a longer budget.

Measured (12.1.0, six `postgres:18-alpine` containers starting concurrently on a 2-vCPU GitHub runner): port binding took up to 16.7 s and failed at 10 s despite `withStartupTimeout(120_000)`; with this change the same suite passes.

Verified locally on the branch: `npm ci` → 0, `tsc -p packages/testcontainers/tsconfig.json --noEmit` → 0, eslint → 0, prettier → 0. The Docker-backed integration suite was not run here.

### DIFF
--- a/packages/testcontainers/src/generic-container/generic-container.ts
+++ b/packages/testcontainers/src/generic-container/generic-container.ts
@@ -160,7 +160,11 @@ export class GenericContainer implements TestContainer {
     if (!inspectResult.State.Running) {
       log.debug("Reused container is not running, attempting to start it");
       await client.container.start(container);
-      inspectResult = await inspectContainerUntilPortsExposed(() => client.container.inspect(container), container.id);
+      inspectResult = await inspectContainerUntilPortsExposed(
+        () => client.container.inspect(container),
+        container.id,
+        this.startupTimeoutMs
+      );
     }
 
     const mappedInspectResult = mapInspectResult(inspectResult);
@@ -217,7 +221,8 @@ export class GenericContainer implements TestContainer {
 
     const inspectResult = await inspectContainerUntilPortsExposed(
       () => client.container.inspect(container),
-      container.id
+      container.id,
+      this.startupTimeoutMs
     );
     const mappedInspectResult = mapInspectResult(inspectResult);
     const boundPorts = BoundPorts.fromInspectResult(client.info.containerRuntime.hostIps, mappedInspectResult).filter(


cristianrgreco:
Thanks for raising the PR, apologies for the delay in reviewing it. Looks good but a few gaps:

1. **`restart()` has the same bug.** `StartedGenericContainer.restart()` (`started-generic-container.ts:95`) also calls `inspectContainerUntilPortsExposed` without a timeout, so #1446 still reproduces on restart. The wait strategy is available there, so you can pass `this.waitStrategy.isStartupTimeoutSet() ? this.waitStrategy.getStartupTimeout() : undefined`.

2. **Timeouts set on the wait strategy itself are still ignored.** `withWaitStrategy(Wait.forLogMessage(...).withStartupTimeout(120_000))` is a documented pattern and doesn't touch `this.startupTimeoutMs`, so the pre-wait would still be capped at 10 s. Deriving the value from the effective strategy rather than the container field would cover both.

3. **This also shortens the pre-wait for anyone with a timeout under 10 s.** The description says only longer budgets change behaviour, but `withStartupTimeout(3_000)` now caps port binding at 3 s where it previously had 10 s (#1038 raised it from 5 s because 5 s was flaky). Either `Math.max(timeout, 10_000)` or call it out explicitly.

4. **Consider making the util's `timeout` parameter required.** The silent default is what let the three call sites drift in the first place, and it's how `restart()` got missed here.

5. A small unit test asserting the third argument is passed through (mocking `./inspect-container-util-ports-exposed`) would lock this in.

PR #1453, opened 2026-09-06
Only reuse reaper containers created by testcontainers-node

### LINKED ISSUE #1442: Reaper discovery adopts another language binding's Ryuk, then leaks every container (findReaperContainers matches only org.testcontainers.ryuk)
### Expected Behaviour

Reaper discovery should only adopt a Ryuk container that belongs to `testcontainers-node`, or — if it adopts one belonging to another language binding — it should register its session with that reaper so the containers it creates are still reaped.

### Actual Behaviour

`findReaperContainers` matches **any** running Ryuk on the host, regardless of which language binding started it. On a CI host shared between a `testcontainers-node` project and a `testcontainers-python` project, the Node workers adopt the Python binding's Ryuk and then leak every container they create.

`packages/testcontainers/src/reaper/reaper.ts` (verified in the published `build/reaper/reaper.js` of both **11.14.0** and **12.1.0** — the code is identical in the two):

```js
async function findReaperContainers(client) {
    const containers = await client.container.list();
    return containers
        .filter((container) => container.State === "running" &&
        container.Labels[LABEL_TESTCONTAINERS_RYUK] === "true" &&
        container.Labels["TESTCONTAINERS_RYUK_TEST_LABEL"] !== "true")
        .sort((a, b) => b.Created - a.Created);
}
```

The only positive predicate is the image label `org.testcontainers.ryuk=true`. That label is on the `testcontainers/ryuk` image itself, so **every** binding's reaper carries it — confirmed on two versions in use side by side:

```
$ docker image inspect testcontainers/ryuk:0.14.0 --format '{{json .Config.Labels}}'
{"org.testcontainers.ryuk":"true"}
$ docker image inspect testcontainers/ryuk:0.8.1  --format '{{json .Config.Labels}}'
{"org.testcontainers.ryuk":"true"}
```

There is no check on `org.testcontainers.lang`, on `org.testcontainers.version`, or on the image tag — even though this library writes `org.testcontainers.lang: "node"` on everything it creates (`utils/labels.ts`, `createLabels()`).

**Why adoption then loses the session.** A reaper started by another binding carries no `org.testcontainers.session-id` label, so this line mints a fresh id per worker:

```js
const existingSessionId = reaperContainer.Labels[LABEL_TESTCONTAINERS_SESSION_ID] ?? new RandomUuid().nextUuid();
```

The adopted reaper is then asked to watch a session it was never told about in a form it durably owns, and the containers created under that id are never reaped.

### Testcontainer Logs

From the **foreign** (Python-started) reaper's own log while Node workers were running against it. Its own session id is a full UUID; the 12-hex ids are the ones this library minted per worker:

```
Adding {"label":{"org.testcontainers.session-id=cf5e8a51-7464-4627-9ede-6c092b479713":true}}
Adding {"label":{"org.testcontainers.session-id=3297709e6c31":true}}
Adding {"label":{"org.testcontainers.session-id=57a593d04394":true}}
```

The observable signature is one leaked container per distinct session id, with no reaper alive that ever owned them. We saw six containers survive with six distinct session ids and no test runner process alive, then reproduced the id-per-worker pattern at 13/13 and 23/23.

Ruled out by measurement, so these are not the cause: `TESTCONTAINERS_RYUK_DISABLED` is unset everywhere (the reaper is running and doing its job for live sessions), and there was no OOM kill.

### Steps to Reproduce

1. On one Docker host, start a `testcontainers-python` suite so that its `testcontainers/ryuk` container is running.
2. While it is up, run any `testcontainers-node` suite against the same host.
3. Observe the Node run log `Reusing existing Reaper for session "<id>"` and that the adopted reaper is the Python one.
4. Let the Node suite end. Its containers are not reaped.

### Environment

- `testcontainers` **11.14.0**; the same code is present in **12.1.0** (latest at time of writing), so this is not fixed by upgrading.
- Node 22.18.0, Linux, Docker with the containerd image store.
- `testcontainers/ryuk` `0.14.0` (this library) and `0.8.1` (the other binding) both present.

### Suggested fix

Narrow the discovery predicate so a reaper is only adopted when this binding can actually own the session. Any of:

- require `org.testcontainers.lang === "node"` on the reaper container, matching what `createLabels()` already writes; or
- require the reaper container to carry an `org.testcontainers.session-id` label at all, rather than falling back to a freshly minted id — adopting a reaper whose session cannot be identified is the step that silently breaks reaping; or
- match on the expected `REAPER_IMAGE` tag.

Happy to open a PR if you would like one, and to test a candidate against the mixed-binding host where this reproduces.

kilisamemarisaaa:
Fixes #1442

## Problem

`findReaperContainers` matched **any** running Ryuk on the host. On a CI host shared between a `testcontainers-node` project and e.g. a `testcontainers-python` project, Node workers adopted the Python binding's reaper. A foreign reaper carries no `org.testcontainers.session-id` label, so every worker fell into

```ts
const existingSessionId = reaperContainer.Labels[LABEL_TESTCONTAINERS_SESSION_ID] ?? new RandomUuid().nextUuid();
```

and minted a fresh per-worker session id — asking the adopted reaper to watch a session it was never durably told about. Every container created under those ids leaks once the run ends (measured 13/13 and 23/23 reproductions in the issue).

## Fix

Extract the adoption predicate into `reaper-discovery.ts` and narrow it so this binding only adopts a reaper it can actually own:

- require `org.testcontainers.lang === "node"` — this library already writes that label on everything it creates via `createLabels()`;
- require a durable `org.testcontainers.session-id` label on the reaper container — adopting a reaper whose session cannot be identified is precisely the step that silently broke reaping.

A reaper that fails the predicate is left to the binding that owns it, and the Node run starts its own reaper instead.

## Verification (red-green, per AGENTS.md)

New `reaper-discovery.test.ts` exercises the extracted predicate with stub clients (no Docker needed):

- **Red** (pre-fix logic, only `export` added): 2 failed / 1 passed — the foreign-reaper fixture was adopted and the anonymous-reaper fixture triggered the random-id mint.
- **Green** (after the predicate change): 3 passed / 0 failed, covering foreign-reaper rejection, unidentifiable-session rejection, and the preserved non-running/test-label exclusions.

Additionally `reaper-discovery.docker.test.ts` adds a Docker-gated integration case (starts an `alpine` container labeled as a foreign binding's reaper and asserts it is not adopted); it runs in CI and skips automatically when no Docker daemon is reachable.

- `tsc --noEmit` clean, prettier/eslint clean on all touched files.
- `reaper.test.ts` unchanged and still Docker-gated as before (fails locally only due to no Docker daemon, same as on `main`).

## Notes

- Local install required `--engine-strict=false` because the dev dependency `npm-check-updates@23.0.0` demands node `^24.15.0` while the repo itself requires `>= 22.22` — no repository files were changed for this.
- Cross-binding behavior follows the de-facto reference implementations: bindings only share a reaper they can identify as their own.

### DIFF
--- a/packages/testcontainers/src/reaper/reaper-discovery.docker.test.ts
+++ b/packages/testcontainers/src/reaper/reaper-discovery.docker.test.ts
@@ -1,0 +1,39 @@
+import { beforeAll, describe, expect, it } from "vitest";
+
+import { ContainerRuntimeClient, getContainerRuntimeClient } from "../container-runtime";
+import { GenericContainer } from "../generic-container/generic-container";
+import { LABEL_TESTCONTAINERS_LANG, LABEL_TESTCONTAINERS_RYUK } from "../utils/labels";
+import { findReaperContainers } from "./reaper";
+
+describe("Reaper discovery against a shared host", () => {
+  let client: ContainerRuntimeClient;
+  let dockerAvailable = true;
+
+  beforeAll(async () => {
+    try {
+      client = await getContainerRuntimeClient();
+      await client.container.list();
+    } catch {
+      dockerAvailable = false;
+    }
+  });
+
+  it("does not adopt a reaper container started by another language binding (#1442)", async () => {
+    if (!dockerAvailable) return;
+
+    const foreignReaper = await new GenericContainer("alpine:3.20")
+      .withCommand(["sleep", "60"])
+      .withLabels({
+        [LABEL_TESTCONTAINERS_RYUK]: "true",
+        [LABEL_TESTCONTAINERS_LANG]: "python",
+      })
+      .start();
+
+    try {
+      const result = await findReaperContainers(client);
+      expect(result.map((container) => container.Id)).not.toContain(foreignReaper.getId());
+    } finally {
+      await foreignReaper.stop({ timeout: 1000 });
+    }
+  });
+});
--- a/packages/testcontainers/src/reaper/reaper-discovery.test.ts
+++ b/packages/testcontainers/src/reaper/reaper-discovery.test.ts
@@ -1,0 +1,64 @@
+import { describe, expect, it } from "vitest";
+
+import { ContainerInfo } from "dockerode";
+import { ContainerRuntimeClient } from "../container-runtime";
+import { LABEL_TESTCONTAINERS_LANG, LABEL_TESTCONTAINERS_RYUK, LABEL_TESTCONTAINERS_SESSION_ID } from "../utils/labels";
+import { findReaperContainers } from "./reaper";
+
+function stubClient(containers: Partial<ContainerInfo>[]): ContainerRuntimeClient {
+  return { container: { list: async () => containers as ContainerInfo[] } } as unknown as ContainerRuntimeClient;
+}
+
+function reaperFixture(id: string, labels: Record<string, string>, state = "running"): Partial<ContainerInfo> {
+  return {
+    Id: id,
+    State: state,
+    Labels: {
+      [LABEL_TESTCONTAINERS_RYUK]: "true",
+      ...labels,
+    },
+    Created: 123,
+  };
+}
+
+describe("findReaperContainers", () => {
+  it("does not adopt reapers started by other language bindings (#1442)", async () => {
+    const foreignReaper = reaperFixture("foreign", { [LABEL_TESTCONTAINERS_LANG]: "python" });
+    const nodeReaper = reaperFixture("node", {
+      [LABEL_TESTCONTAINERS_LANG]: "node",
+      [LABEL_TESTCONTAINERS_SESSION_ID]: "0123456789ab",
+    });
+
+    const result = await findReaperContainers(stubClient([foreignReaper, nodeReaper]));
+
+    expect(result.map((container) => container.Id)).toEqual(["node"]);
+  });
+
+  it("does not adopt a reaper without an identifiable session id", async () => {
+    const anonymousReaper = reaperFixture("anonymous", { [LABEL_TESTCONTAINERS_LANG]: "node" });
+
+    const result = await findReaperContainers(stubClient([anonymousReaper]));
+
+    expect(result).toEqual([]);
+  });
+
+  it("does not adopt non-running containers or test reapers", async () => {
+    const stoppedReaper = reaperFixture(
+      "stopped",
+      {
+        [LABEL_TESTCONTAINERS_LANG]: "node",
+        [LABEL_TESTCONTAINERS_SESSION_ID]: "0123456789ab",
+      },
+      "exited"
+    );
+    const testReaper = reaperFixture("test", {
+      [LABEL_TESTCONTAINERS_LANG]: "node",
+      [LABEL_TESTCONTAINERS_SESSION_ID]: "0123456789ab",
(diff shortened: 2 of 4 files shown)


kilisamemarisaaa:
Self-review correction on my own PR, following the repo's cross-language guidance (I cited a de-facto-reference sentence without evidence — that was sloppy):

I verified the Java binding now: `RyukResourceReaper` (`core/src/main/java/org/testcontainers/utility/RyukResourceReaper.java`) **never discovers or adopts an existing Ryuk container** — it always starts its own per session (`maybeStart()`) and registers only its own containers via label filters sent over the Ryuk socket (`DEATH_NOTE` / `register(filters)`). There is no adoption/discovery path at all.

So the de-facto reference behavior is actually **stricter** than this PR's predicate: mature bindings don't adopt foreign reapers *period*, which further supports narrowing the node-side adoption predicate (here we keep the useful cross-process reuse of node-owned reapers, but refuse anything the binding can't own). I could not complete a verified check of the Python binding from this environment — leaving that part to reviewers rather than asserting it.

cristianrgreco:
Hi @kilisamemarisaaa, thanks for raising.

> I verified the Java binding now: RyukResourceReaper (core/src/main/java/org/testcontainers/utility/RyukResourceReaper.java) never discovers or adopts an existing Ryuk container — it always starts its own per session (maybeStart()) and registers only its own containers via label filters sent over the Ryuk socket (DEATH_NOTE / register(filters)). There is no adoption/discovery path at all.

Indeed, but in testcontainers-go and other langs where test frameworks spawn separate processes per test suite (e.g. jest/vitest), we decided that testcontainers-node should reuse reaper containers should they exist. Otherwise for example if you have 20 cores and are running 20 tests, it doesn't make much sense to have 20 ryuk containers.

I agree though it's not the intention that a testcontainers-node process picks up a ryuk container from a testcontainers-python process, and the solution to narrow down on language makes sense.

---

A few changes before merging:

1. **Remove the fallback in `getReaper`.** With the new check, a reaper without a session-id label is never returned, so `?? new RandomUuid().nextUuid()` can't run anymore. Please change it to read the label directly.

2. **Keep the check in `reaper.ts`.** A separate `reaper-discovery.ts` isn't needed for a five-line filter. It also defines its own copy of the `TESTCONTAINERS_RYUK_TEST_LABEL` constant, while `reaper.ts` still uses the string literal.

3. **Replace the two new test files with one case in the existing `reaper.test.ts`.** Follow the pattern of `should reuse existing reaper container if one is already running`:
   - take the real reaper's `ContainerInfo`
   - set `org.testcontainers.lang` to `python`
   - mock `client.container.list`
   - assert that `getReaper` creates a new reaper rather than reusing that one

   That tests what users actually see, and `findReaperContainers` wouldn't need to be exported. Two problems with the current Docker test:
   - It pulls alpine:3.20 rather than the image the rest of the suite uses (cristianrgreco/testcontainer:1.1.14). The cold pull made it time out locally, because the test has no timeout and vitest's default is 5s.
   - The `if (!dockerAvailable) return;` guard lets it pass without testing anything. The whole suite assumes Docker is available, so please drop the guard.

4. Minor: the per-worker random session id wasn't really what leaked. Each worker registers its own id with the reaper, so its containers get cleaned up as long as that reaper stays alive. The actual risk is that the other binding decides when its reaper stops and how it's configured. That still supports this change, but the PR description doesn't reflect it.

PR #1457, opened 2026-09-13
Add a global never-pull image policy

### LINKED ISSUE #1456: Support a global never-pull image policy
A pull policy returning `false` currently disables forced refreshes but still pulls missing images. CI systems that preload images cannot enforce local-only execution, including for helpers such as Ryuk.

Could we add `TESTCONTAINERS_PULL_POLICY=never` to use local images and fail before registry authentication or pulls if an image is unavailable? It should cover helper containers and leave defaults unchanged. This is an image-pull restriction, not general network isolation or an offline Docker build mode.

Related: #1122 addressed unnecessary registry access for existing images; this would also prohibit pulling missing images.

longlho:
Add `TESTCONTAINERS_PULL_POLICY=never` to require local images before registry auth or pulls, including helper containers. Document that builds, Compose, and container networking are outside its scope. Opt-in only; existing policies and defaults are unchanged.

Validation: 8 expected regression failures before the fix; all 13 focused/existing policy tests pass after it. Format, lint, cross-module TypeScript checks, and core build pass.

Closes #1456.


longlho:
@cristianrgreco would u mind taking a look?

cristianrgreco:
Will get to this soon @longlho. Apologies, quite busy personally at the moment

cristianrgreco:
Hi @longlho, I'm going to close this PR in favour of #1460, which better follows my suggestion here https://github.com/testcontainers/testcontainers-node/discussions/1455#discussioncomment-18416142. Having a `PullPolicy.neverPull()` better fits the existing API design. 

Please have a look and let me know if anything's missing or not ideal there and we can iterate.

longlho:
Yup thanks! I'll take a look

PR #1460, opened 2026-09-14
Add never-pull policy

gologames:
Implements the per-container pull policy discussed in #1455.

Adds `PullPolicy.neverPull()` for `GenericContainer` and Docker Compose. GenericContainer fails when the image is missing locally. Existing boolean pull policies retain their behavior.

Compose uses `--pull never` and skips the explicit pull step without disabling builds. Dockerfile builds reject this policy because the build API cannot guarantee that base images will not be pulled (https://github.com/docker/buildx/issues/1889). Helper images such as Ryuk are unaffected.

### Validation

- Formatting, lint, and cross-module TypeScript checks passed.
- All CI checks passed.

CI: https://github.com/gologames/testcontainers-node/pull/1/checks


cristianrgreco:
Thanks for this @gologames, the feature makes sense. A couple of things before it goes in:

**1. Pull policy shape**

Having both `shouldPull()` and `neverPull()` allows an invalid state (both true). That's why we need `resolvePullPolicy()`, the "conflicting settings" error, and conflict tests in several places. I'd rather `shouldPull` returned a third value so that state can't exist:

```ts
export interface ImagePullPolicy {
  shouldPull(): boolean | "never";
}
```

This isn't a breaking change: existing custom policies that return a boolean still type-check and keep their meaning. Call sites then need an explicit check, because `"never"` is truthy:

```ts
// GenericContainer
const shouldPull = this.pullPolicy.shouldPull();
if (shouldPull === "never") {
  if (!(await client.image.exists(this.imageName))) {
    throw new Error(`Image "${this.imageName.string}" does not exist locally and pull policy is "never"`);
  }
} else {
  await client.image.pull(this.imageName, { force: shouldPull, platform: this.createOpts.platform });
}
```

Using `exists()` gives a clearer error than the raw 404 from `inspect()`, and it reuses the existing images cache.

For Compose, `--pull never` on `up` plus skipping `compose.pull` is enough. I'd drop `--no-build` and the `withBuild()` rejection. They go beyond the ask, and users can pass `--no-build` through `commandOptions` if they want it.

Rejecting `"never"` in the Dockerfile builder is fine. It only needs one test, though: the check runs before BuildKit is involved, so looping over `buildkit=true/false` doesn't add anything.

**2. Size**

Only about 60 of the ~430 added lines are production code. I'd like to cut the tests down a lot:

- The conflict tests and "evaluated once" tests go away with the change above.
- Please drop the image filtering in `waitForDockerEvent`, `createTempImageTag`, and their tests. `vi.spyOn(client.image, "pull")` / `vi.spyOn(client.compose, "pull")` already proves no pull happened.
- The Compose `commandOptions` `it.each`, the selected-services test, and the `--build and --no-build are incompatible` test are testing Docker Compose's own flag parsing, not our code.
- The "propagate inspection failures" test isn't needed.

Roughly: one unit test for the policy. For GenericContainer, one test that a local image starts without a pull and one that a missing image fails without a pull. For Compose, one test that a missing image fails without a pull.

gologames:
Thanks @cristianrgreco, updated as requested: simplified the policy API, switched to `exists()`, removed the Compose build restrictions, and trimmed the tests and helpers. All CI checks pass in the fork: https://github.com/gologames/testcontainers-node/pull/1/checks

packages/testcontainers/src/generic-container/generic-container-dockerfile.test.ts
```diff
@@ -93,6 +94,15 @@ describe("GenericContainer Dockerfile", { timeout: 180_000 }, () => {
     });
   }
 
+  it("should reject never-pull before contacting the runtime", { concurrent: false }, async () => {
+    const clientSpy = vi.spyOn(containerRuntime, "getContainerRuntimeClient");
```
cristianrgreco: The `getContainerRuntimeClient` spy and the `* as containerRuntime` import aren't needed. The rejected build already shows the behaviour, and without the spy the test doesn't need `concurrent: false`

gologames: @cristianrgreco I removed the spy, unused import, and `concurrent: false`. All CI checks pass in the fork

cristianrgreco (approved):

